PR #13, opened 2021-05-18
Add availability type of SQL instance

mbayangdiop95:


terraform/variables.tf
```diff
@@ -48,6 +49,12 @@ variable "gcp_sql_user_name" {
   description = "The username for the app component database"
   default     = "hexa-app"
 }
+variable "availability_type" {
```
pvanliefland, line 52: @mbayangdiop95 , a few comments:

- Please use the same naming convention as for the other variables (`gcp_sql_instance_...`)
- Let's move this variable a bit higher so that it is grouped with other `gcp_sql_instance_...` variables
- It also needs to be defined (and used ) in the `terraform_import` directory

pvanliefland (changes requested):
@mbayangdiop95 a few todos:

- [x] Rename the variable (see remark)
- [x] Adapt `terraform_import`
- [ ] Create a secret (`GCP_SQL_INSTANCE_...`) in the test environment
- [ ] Adapt the 3 terraform workflows as well - they should provide a value for this variable

pvanliefland:
Great, merging

PR #21, opened 2021-07-20
HEXA-267 GraphQL for S3

nikita-marchant:


hexa/plugins/connector_s3/schema.py
```diff
@@ -23,7 +26,48 @@
...
+        # S3Object
+        bucket: S3Bucket
+        parent: S3Object
+        "Path of the object within the bucket"
+        s3Key: String
+        s3Size: Int
+        s3StorageClass: String
+        s3Type: String
+        s3Name: String
+        s3LastModified: DateTime
```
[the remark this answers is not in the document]
pvanliefland: Let's add `mimeType` as well

config/settings.py
```diff
@@ -176,3 +176,5 @@
 
 # Notebooks component
 NOTEBOOKS_URL = os.environ.get("NOTEBOOKS_URL", "http://localhost:8001")
+
+GRAPHQL_PAGE_SIZE = 10
```
pvanliefland, line 180: To discuss together

hexa/plugins/connector_s3/schema.py
```diff
@@ -23,7 +27,54 @@
         lastSyncedAt: DateTime
         tags: [CatalogTag!]
         icon: String!
+        Objects(
```
pvanliefland, line 30: Should be lowercased `objects`

hexa/plugins/connector_s3/schema.py
```diff
@@ -23,7 +27,54 @@
...
+
+        # S3Object
+        bucket: S3Bucket
+        parent: S3Object
+        "Path of the object within the bucket"
+        s3Key: String
+        s3Size: Int
+        s3StorageClass: String
+        "enum: 'file' or 'directory'"
+        s3Type: String
```
pvanliefland, line 61: If it's an enum, let's use an enum :) https://graphql.org/learn/schema/#enumeration-types

hexa/plugins/connector_s3/schema.py
```diff
@@ -23,7 +27,54 @@
...
+        s3Name: String
+        s3LastModified: DateTime
+        s3MimeType: String
+
+        Objects(
```
pvanliefland, line 66: Should be lowercased `objects`

hexa/plugins/connector_s3/tests/test_connector.py
```diff
@@ -0,0 +1,224 @@
...
+                    "variables": {},
+                    "query": payload,
+                }
+            ),
+            content_type="application/json",
+        ).content
+    )
+
+
+class ConnectorS3Test(test.TestCase):
```
pvanliefland, line 32: I would like to discuss this part with you too.

In a nutshell:

- I think we need to split the tests between something like `test_views`, `test_models` and `test_graph`
- The query method is going to be used a lot so we might as well have a `GraphQLTestCase` class with the `query` method
- It's not super important but I like using f-strings instead of interpolation

pvanliefland (changes requested):

pvanliefland (changes requested):

PR #28, opened 2021-08-04
HEXA-194 Add a PostgreSQL datasource

nikita-marchant:
Adds a new datasource plugin for PostgreSQL. The only level of detail is a `PostgresqlDatabase` that represents the tuple `(hostname, port,dbname, username, password)`. Tables and columns are not synced at the moment.
Environment variables are injected for each database in JupyterHub and work has started to also generate a `.pgpass` file.


config/settings.py
```diff
@@ -54,6 +54,7 @@
     "hexa.plugins.connector_dhis2.apps.Dhis2ConnectorConfig",
     "hexa.plugins.connector_s3.apps.S3ConnectorConfig",
     "hexa.plugins.connector_airflow.apps.ConnectorAirflowConfig",
+    "hexa.plugins.connector_postgresql.apps.SQLConnectorConfig",
```
pvanliefland, line 57: Should be `PostgreSQLConnectorConfig`

hexa/plugins/connector_postgresql/apps.py
```diff
@@ -0,0 +1,16 @@
+from hexa.plugins.app import ConnectorAppConfig
+
+
+class SQLConnectorConfig(ConnectorAppConfig):
```
pvanliefland, line 4: `PostgreSQLConnectorConfig`

hexa/plugins/connector_postgresql/models.py
```diff
@@ -0,0 +1,100 @@
...
+            bucketpermission__team__in=[t.pk for t in user.team_set.all()]
+        )
+
+
+class PostgresqlDatabase(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    hostname = models.CharField(max_length=200)
+    username = models.CharField(max_length=200)
+    password = EncryptedTextField(max_length=200)
```
pvanliefland, line 33: I would prefer that we keep the `Credentials` model pattern - let's discuss it...

pvanliefland (changes requested):
All in all quite good, but I want to discuss the module name and the (absence of) a `Credentials` model

PR #42, opened 2021-08-23
Index S3 files + use tg_trgm in search to accomodate `/` in the names

nikita-marchant:


hexa/catalog/models.py
```diff
@@ -2,7 +2,13 @@
 
 from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.contenttypes.models import ContentType
-from django.contrib.postgres.search import SearchVector, SearchQuery, SearchRank
+from django.contrib.postgres.search import (
+    SearchVector,
+    SearchQuery,
+    SearchRank,
+    TrigramSimilarity,
+    TrigramDistance,
```
pvanliefland, line 10: Are we using the distance?

hexa/catalog/models.py
```diff
@@ -44,20 +51,26 @@ def search(self, query, *, limit=10):
...
             "short_name",
             "external_short_name",
             "description",
             "external_description",
             "countries",
-        )
+        ]
+        search_vector = SearchVector(*fields)
         search_query = SearchQuery(query, config=models.F("text_search_config"))
         search_rank = SearchRank(vector=search_vector, query=search_query)
```
pvanliefland, line 65: Maybe just add a comment to summarize the approach?

pvanliefland (approved):
Good to merge as-is, but 2 little suggestions :)

PR #43, opened 2021-08-24
Upgrade Tailwind and use pip-compile

nikita-marchant:


docker-compose.yaml
```diff
@@ -1,23 +1,9 @@
...
-  db:
-    image: postgres:12.5
-    environment:
-      - POSTGRES_DB=hexa-app
-      - POSTGRES_USER=hexa-app
-      - POSTGRES_PASSWORD=hexa-app
-    ports:
-      - 5432:5432
-  app:
+x-app: &default-app
```
pvanliefland, line 3: Could you comment?

pvanliefland (changes requested):
If you test it and it works properly, feel free to merge.

Suggestions:
- Add a few explanations
- Update `README`
- Check the deletion of static files (css / js)

PR #57, opened 2021-09-07
[enh] sync of postgresql datasource to queue/async

ekatsah:
This PR enables openhexa to delay postgresql ```sync()``` and do the work in a separate process (called ```postgresql_sync_worker```). This change of behavior is activated with a new settings, ```DATASOURCE_ASYNC_REFRESH```, which is False by default (no change of behavior, no change of deploy settings needed). It use the django-postgres-queue because it doesnt need a lot of new infrastructure (like celery or something else) and is suffisant for the immediate future.

Notes:
- add a new test, which set the ASYNC_REFRESH to both values to see if queued sync or immediate sync works. use package "mock" to check when datasource.sync() is called
- test pass localy, but not on GH because of dockerhub credentials, sorry :/
- it needs a new worker to run the postgresql task queue, but not by default

If the proposed approch is OK, i can do s3/dhis2 next :)

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -42,6 +42,7 @@ INSTALLED_APPS = [
     "corsheaders",
     "django_countries",
     "django_ltree",
+    "dpq",
     "ariadne.contrib.django",
     "tailwind",
     "hexa.user_management",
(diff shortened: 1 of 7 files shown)


pvanliefland (changes requested):
Looking good, thanks a lot :)

I think we should consider having a single queue, and a single view, for sync operations. What do you think?

A mention / explanation about workers and queues in the README would be nice as well.

ekatsah:
Closed because rewrited on catalog module

PR #67, opened 2021-09-14
Enable async datasource refresh for all catalogs

ekatsah:
This PR enables openhexa to delay ```sync()``` for all catalogs, and do the work in a separate process (called ```sync_datasources_worker```). This change of behavior is activated with a new settings, ```DATASOURCE_ASYNC_REFRESH```, which is False by default (no change of behavior, no change of deploy settings needed). It use the django-postgres-queue because it doesnt need a lot of new infrastructure (like celery or something else) and is suffisant for the immediate future.

Notes:
- add a new test, which set the ASYNC_REFRESH to both values to see if queued sync or immediate sync works. use package "mock" to check when datasource.sync() is called
- test pass localy, but not on GH because of dockerhub credentials, sorry :/
- it needs a new worker to run the postgresql task queue, but not by default
- since we check perm on manager of each subtype (```filter_for_user()```), i created a DatasourceQuerySet to use for all datasources's querysets. 

README update coming

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ refresh with the settings ```DATASOURCE_ASYNC_REFRESH```.If you have slow dataso
 to decrease the latency of the web server. In that case, a queue will be used to dispatch the refresh request to a
 background worker. This worker is called ```sync_datasources_worker```, callable like this:
 ```bash
-./manage.py sync_datasources_worker
+python manage.py sync_datasources_worker
 ```
 
 Building the Docker image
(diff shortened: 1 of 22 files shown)


pvanliefland:
@ekatsah it looks good, but I think you forgot to include `queue.py`.

pvanliefland:
And maybe your tests

ekatsah:
Oh my.. the dev version of forgetting to attach the files in an important mail :)

pvanliefland (approved):
LGTM! Thanks @ekatsah , nice work 👍 

@nikita-marchant you can merge it after your review if no blocking issue.

nikita-marchant (approved):

PR #69, opened 2021-09-16
Cache busting for js

pvanliefland:
This PR enables Whitenoise caching / cache busting/

See https://whitenoise.evans.io/en/stable/django.html#add-compression-and-caching-support and https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#manifeststaticfilesstorage.

In a nutshell:

- `collectstatic` will generate static files with a unique hash
- using the `static` template tag will generate a link to the file names with hashes

### DIFF
--- a/config/settings.py
+++ b/config/settings.py
@@ -32,6 +32,8 @@ ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "").split(",")
 
 # Application definition
 INSTALLED_APPS = [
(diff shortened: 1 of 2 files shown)


pvanliefland:
@nikita-marchant let's discuss this

nikita-marchant:
Nice catch the `TEST_RUNNER = "hexa.core.test.runner.DiscoverRunner"` !
This is why my PR did not work https://github.com/BLSQ/openhexa-app/pull/44

nikita-marchant (approved):

PR #74, opened 2021-09-20
Add a parameter to force the sync to be synchronous

nikita-marchant:


pvanliefland (changes requested):
Let's add a test for this as well

PR #76, opened 2021-09-22
Better handling of S3 relocation detection with ETag

nikita-marchant:


pvanliefland (approved):
Let's try that until we find a better way

PR #81, opened 2021-10-11
[fix] HEXA-377 Make sync CRON more resilient

ekatsah:


hexa/catalog/management/commands/sync.py
```diff
@@ -22,4 +26,7 @@ def handle(self, *args, filter_app, **options):
                     pbar = tqdm.tqdm(instances)
                     pbar.set_description(f"   {model.__name__:15}")
                     for instance in pbar:
-                        instance.sync()
+                        try:
+                            instance.sync()
+                        except Exception:
+                            logger.exception("sync")
```
pvanliefland, line 32: Logging is good but as of now, we can't really see logs (there is a card for that).

With this change, we will lose Sentry notifications - we should trigger it manually.

PR #96, opened 2021-11-07
HEXA-343 Read-only S3 buckets

ekatsah:
Note: test should pass if we merge #97 before + merge master. It helps us remove one migration and remove the need for a migration merge.


hexa/plugins/connector_s3/api.py
```diff
@@ -152,23 +155,48 @@ def generate_sts_user_s3_credentials(
...
+    ro_bucket_names: typing.Optional[typing.Sequence[str]] = None,
+) -> typing.Dict:
+    statements = []
+
+    if ro_bucket_names:
+        statements.append(
+            {
+                "Sid": "S3ROActions",
+                "Effect": "Allow",
+                "Action": ["s3:ListBucket", "s3:GetObject*"],
```
pvanliefland, line 169: Are we sure that `s3:GetObject` (without the `*`) is not enough?

hexa/plugins/connector_s3/api.py
```diff
@@ -64,9 +64,10 @@ def generate_sts_app_s3_credentials(
 def generate_sts_user_s3_credentials(
     *,
     principal_credentials: hexa.plugins.connector_s3.models.Credentials,
-    buckets: typing.Sequence[hexa.plugins.connector_s3.models.Bucket],
     role_identifier: str,
     session_identifier: str,
+    ro_buckets: typing.Sequence[hexa.plugins.connector_s3.models.Bucket],
```
pvanliefland, line 69: Let's use `read_only_buckets` and `read_write_buckets`

hexa/plugins/connector_s3/api.py
```diff
@@ -152,23 +155,48 @@ def generate_sts_user_s3_credentials(
...
 
 
-def generate_s3_policy(bucket_names: typing.Sequence[str]) -> typing.Dict:
-    policy = {
-        "Version": "2012-10-17",
-        "Statement": [
+def generate_s3_policy(
+    rw_bucket_names: typing.Sequence[str] = None,
+    ro_bucket_names: typing.Optional[typing.Sequence[str]] = None,
+) -> typing.Dict:
```
pvanliefland, line 161: Same here, `read_only_buckets` and `read_write_buckets`

pvanliefland (changes requested):
Almost ready imho, except for:

- A few (minor) change requests
- Adapt `test_credentials_s3fs` in `test_credentials.py` to check the mode
- We also miss a test that really verifies that we generate the proper policy depending on user bucket permissions

ekatsah:
See #99

PR #97, opened 2021-11-07
HEXA-398 Fix WHO tracker demo sync

ekatsah:
Add auto_sync flag for all datasource + check that flag at sync time

Note: test should pass if we merge #96 before + merge master. It helps us remove one migration and remove the need for a migration merge.


pvanliefland (changes requested):
LGTM except for one thing: why don't we have a migration for the `Bucket` class? It extends `Datasource` as well...

ekatsah:
it's in the #96, in another migration. i propose to validate the "concept" of this PR first, then merge #96 & #97 together to validate CI then merge to master. The goal is to remove the need of 2 separate migrations + a merge.

ekatsah:
See #99

PR #99, opened 2021-11-08
HEXA-398 Fix WHO tracker demo sync, HEXA-343 Read-only S3 buckets

ekatsah:


hexa/plugins/connector_s3/models.py
```diff
@@ -46,13 +46,37 @@ def display_name(self):
...
+                return self
+
+        if mode is None:
+            # return all buckets
+            modes = [BucketPermissionMode.READ_ONLY, BucketPermissionMode.READ_WRITE]
+        else:
+            modes = [mode]
+
+        return self.filter(
+            bucketpermission__team__in=[t.pk for t in user.team_set.all()],
```
pvanliefland, line 70: Is this one still needed?

pvanliefland (approved):
LGTM! Still a question about `filter_by_mode` but it does not prevent merging.

hexa/plugins/connector_s3/models.py
ekatsah: It's used in the views for example

PR #101, opened 2021-11-16
HEXA-383 improve datasource/environment sync/index log output

ekatsah:


pvanliefland (approved):
LGTM! If you have the time, I would log both the start and the end time (with the duration). But can be merged as is.

ekatsah:
the times will be in the log messages, probably suffisant for now?

PR #102, opened 2021-11-16
Tabbed detail pages in catalog

pvanliefland:
Goal: replacing the long scrolling detail screen by a tabbed screen, with content displayed first.

Quite a lot of changes:

- New detail header component
- Slimmed down S3 datacards
- Moved content, details, comments and usage to their own tab
- Fixed download (only for objects, not directories)
- Adapted upload feature to new layout
- Cleanup / minor refactoring

https://user-images.githubusercontent.com/690667/143424830-5bb6959e-b5ee-4a2b-a55a-a01204bbfa72.mov


hexa/plugins/connector_s3/datacards.py
```diff
@@ -1,11 +1,11 @@
...
 from django.utils.translation import ugettext_lazy as _
 
 from hexa.catalog.datacards import OpenHexaMetaDataSection
 from hexa.plugins.connector_s3.models import Bucket, Object
-from hexa.ui.datacard import Action, CodeProperty, Datacard, Section, TextProperty
+from hexa.ui.datacard import CodeProperty, Datacard, Section, TextProperty
+from hexa.ui.utils import StaticText
 
 
 class BucketSection(Section):
```
ekatsah, line 11: Line 18: seems redundant with templates/.../usage.html

hexa/plugins/connector_s3/templates/connector_s3/object_detail.html
```diff
@@ -1,24 +1,38 @@
...
-            {% endslot %}
-            {% slot actions %}
-                {% include "connector_s3/components/upload.html" with bucket=datasource %}
-            {% endslot %}
-            {% slot content %}
-                {{ datagrid }}
-            {% endslot %}
-        {% endembed %}
-    {% endif%}
+    {% static "connector_s3/img/symbol.svg" as symbol %}
```
ekatsah, line 6: The tab bar for a file seems useless: it contains only one tab. Maybe it could be removed?

Futhermore, if object.type != directory => don't show tabs, maybe this template should be different for file or directory?
download button + {{ object_card }}
Directory: tabs + {{ object_grid }} + {{ object_card }}
Maybe too early generalisation?

ekatsah (approved):
Super nice!

ekatsah:
Consider merging #111 in this branch before merging this one in main

hexa/plugins/connector_s3/datacards.py
pvanliefland: Good catch, fixed

hexa/plugins/connector_s3/templates/connector_s3/object_detail.html
pvanliefland: I agree but this can be done afterward. I'm also hesitating because we should put back the metadata / comments as soon as we have a solution for preserving metadata when moving or renaming files.

PR #106, opened 2021-11-20
HEXA-457 cannot authenticate in notebooks if email address is too long

ekatsah:


hexa/plugins/connector_s3/api.py
```diff
@@ -141,9 +146,19 @@ def generate_sts_user_s3_credentials(
         aws_secret_access_key=principal_credentials.secret_access_key,
     )
 
+    role_session_name = f"sts-{principal_credentials.username}-{session_identifier}"
+
+    if len(role_session_name) >= 63:
```
pvanliefland, line 151: We need to be able to identify the user. Can we use the user id instead?

It think passing the user id as the `session_identifier` would do the trick.

pvanliefland: (the need is for auditing purpose)

hexa/plugins/connector_s3/tests/test_api.py
```diff
@@ -120,6 +120,32 @@ def test_generate_sts_app_team_credentials(self, _):
             self.assertIsInstance(credentials[key], str)
             self.assertGreater(len(credentials[key]), 0)
 
+    @mock_iam
+    @mock_sts
+    @patch("hexa.plugins.connector_s3.api.sleep", return_value=None)
+    def test_generate_sts_app_team_credentials_long_ident(self, _):
```
pvanliefland, line 126: Not sure we need the test if we use the user id...

pvanliefland (changes requested):

pvanliefland (approved):
Perfect, merging

PR #109, opened 2021-11-25
Rename dashboards

ekatsah:
- I didn't manage to the migrations right: it create some new tables, indexes etc. If merged, it will require some manual cleaning of production and demo DB.
- ~30min :)


hexa/visualizations/apps.py
```diff
@@ -2,5 +2,5 @@
 
 
 class DashboardsConfig(AppConfig):
```
pvanliefland, line 4: Should be VisualizationsConfig

hexa/visualizations/models.py
```diff
@@ -65,7 +65,7 @@ class ExternalDashboard(IndexableMixin, models.Model):
     picture = models.FileField(upload_to="external_dashboard")
     credentials = EncryptedTextField(null=True, blank=True)
 
-    indexes = GenericRelation("dashboards.Index")
+    indexes = GenericRelation("visualizations.Index")
```
pvanliefland, line 68: (Not related to this line but can't comment above)

Queryset should be `ExternalDashboardQuerySet`

pvanliefland: Also `ExternalDashboardPermission.__str__()`: should be `self.external_dashboard`

pvanliefland (changes requested):
Almost ok, a few things to rename / fix

PR #110, opened 2021-11-27
HEXA-447 Sync and index DHIS2 org units

ekatsah:


hexa/plugins/connector_dhis2/api.py
```diff
@@ -43,6 +43,8 @@ def get_relations(self) -> dict[Dhis2Relation, list]:
         for relation in self.RELATIONS:
             links = [
                 x[relation.dhis2_target_name]["id"]
+                if relation.dhis2_target_name
```
pvanliefland, line 46: This feels weird - let's discuss it

hexa/plugins/connector_dhis2/api.py
```diff
@@ -165,3 +183,15 @@ def fetch_indicators(self):
...
         ):
             yield [IndicatorResult(data) for data in page["indicators"]]
+
+    def fetch_organisation_units(self):
+        for page in self._api.get_paged(
+            "organisationUnits", params={"fields": ":all"}, page_size=100
+        ):
+            # rewrite path -> replace "/" by "." for correct ltree path
+            # warning: in place edit, can side effect on tests
+            for element in page["organisationUnits"]:
```
pvanliefland, line 193: This rewrite should happen in `sync.py`

pvanliefland: But let's discuss it, might be ok until we agree on a "fetch & sync" pattern

hexa/plugins/connector_dhis2/api.py
```diff
@@ -127,6 +129,22 @@ class IndicatorTypeResult(Dhis2Result):
...
+class OrganisationUnitResult(Dhis2Result):
+    FIELD_SPECS = {
+        "code": (str, ""),
+        "path": (str, "/"),
+        "leaf": (bool, None),
+    }
+    RELATIONS = [
+        Dhis2Relation(
+            dhis2_field_name="dataSets",
+            dhis2_target_name=None,
```
pvanliefland, line 141: Also weird

hexa/plugins/connector_dhis2/views.py
```diff
@@ -142,6 +200,68 @@ def data_element_detail(
     )
 
 
+def organisation_unit_detail(
```
pvanliefland, line 203: I think it would be nice to have some of the query logic in queryset methods (nice to have)

hexa/ui/templates/ui/icons/organisation.html
```diff
@@ -0,0 +1,11 @@
+<svg
```
pvanliefland, line 1: Where does this icon come from? If it's from heroicons, we should keep the "heroicons" name.

hexa/plugins/connector_dhis2/views.py
```diff
@@ -142,6 +200,68 @@ def data_element_detail(
...
+        page_parameter_name="ds_page",
+        request=request,
+    )
+
+    sub_organisation_unit_grid = OrganisationUnitGrid(
+        instance.organisationunit_set.filter(
+            path__match=str(organisation_unit.path) + ".*{1}"
+        ).prefetch_indexes(),
+        page=int(request.GET.get("ou_page", "1")),
+        page_parameter_name="ou_page",
```
pvanliefland, line 225: I would suggest to drop pagination in detail pages, let's discuss this too.

pvanliefland (changes requested):
Cool, great job! I think it's almost perfect, I would just like to discuss 2 or 3 items before merging.

hexa/plugins/connector_dhis2/api.py
ekatsah: I agree. What about this?

hexa/ui/templates/ui/icons/organisation.html
ekatsah: Random SVG found on internet without license + manipulation in inkscape.

hexa/plugins/connector_dhis2/api.py
ekatsah: I agree. I hesited a lot where to do that operation, and I propose the following "rule": When doing input validation & cleaning, do it as early as possible.

hexa/plugins/connector_dhis2/views.py
ekatsah: Good point

PR #111, opened 2021-11-28
support RO buckets and directories in catalog views

ekatsah:
- fix security flaw in views.object_uploads
- disable front/upload button if bucket RO

image


pvanliefland (approved):
LGTM! I have a suggestion for permissions management but this can be discussed later.

PR #112, opened 2021-11-29
Recursion issue when bucket contains a "/" directory

pvanliefland:
Fix for our annoying recursion issue when syncing s3 buckets with `/` directories.

There was a strange `!= "directory"` test in the sync that prevented the sync to skip "self-replicating syncs". Added a test to make sure.


ekatsah (approved):

pvanliefland:
Not good enough

PR #119, opened 2021-12-10
Dhis2 CSV exports

pvanliefland:
This PR adds a CSV export action to DHIS2 datagrids.

The metadata is not included in the CSV for now.

Screenshot 2021-12-10 at 12 18 56


ekatsah (commented):
I find it surprising that, when the datagrid shows a subset of (dataset, org unit etc), the download generates a file containing more items inside (all dataset, org unit etc from the dhis2 instance)
what our users tell us about that?

PR #123, opened 2021-12-18
HEXA-476 sync use boto3 instead of s3fs

ekatsah:


hexa/plugins/connector_s3/models.py
```diff
@@ -490,6 +486,8 @@ def update_metadata(self, object_data):
         self.parent_key = self.compute_parent_key(object_data["true_key"])
         if "size" in object_data:
```
pvanliefland, line 487: Probably not needed anymore

hexa/plugins/connector_s3/models.py
```diff
@@ -141,58 +140,55 @@ def clean(self):
...
-            secret=sts_credentials["SecretAccessKey"],
-            token=sts_credentials["SessionToken"],
-        )
-        fs.invalidate_cache()
+
+        # transform objects -> extract directories
+        directories, last_modified = set(), {}
+        for object in objects:
+            object["type"] = "file"
+            object["true_key"] = object["Key"]
```
pvanliefland, line 152: Check if `true_key` is still needed

hexa/plugins/connector_s3/models.py
```diff
@@ -141,58 +140,55 @@ def clean(self):
...
+                    dirname not in last_modified
+                    or last_modified[dirname] < object["LastModified"]
+                ):
+                    last_modified[dirname] = object["LastModified"]
+
+        # transform directories
+        directories = [
+            {
+                "Key": name,
+                "true_key": name,
```
pvanliefland, line 168: Check if we can remove `true_key`

pvanliefland (approved):
Perfect! Let's just check if we can remove the `s3fs` dependency and the old workarounds.

PR #124, opened 2021-12-19
HEXA-496 Dhis2 increase verbosity of sync operation

ekatsah:
Also, contains a missing migration for postgresql connector


pvanliefland (approved):
Maybe move `verbose` to client init?

PR #132, opened 2021-12-28
Duplicate dhis2 elements with the same id of different instance

ekatsah:
The PDSS and SNIS pyramids contain 24k org unit with the same id. when openhexa sync SNIS, it updates PDSS pyramid with SNIS data...


pvanliefland (approved):
LGTM!

PR #134, opened 2021-12-30
HEXA-534 missing length parameter from django4

ekatsah:


docker-entrypoint.sh
```diff
@@ -43,9 +43,12 @@ case "$command" in
   ;;
 "test")
   wait-for-it db:5432
+  python manage.py makemigrations --check
```
ekatsah, line 46: add more resilience: pushing code without migration to production should fail.

If we really need to make a deployment without proper migration (might happen due to kubernetes), the proposal is to manually comment this line for a release as a manual override

pvanliefland: Good idea!

pvanliefland (approved):
LGTM!

PR #136, opened 2021-12-31
HEXA-527 clean up the orphan concept for catalog/datasources

ekatsah:
- Remove orphan flag from s3 object
- simplify sync algo: fix bug if a key switch from a type directory to a file and vis versa
- refactor DatasourceSyncResult to contain a deleted count, improve consistency with postgres sync

Why DHIS2 doesnt delete objects in openhexa? is DHIS2 append only? (are we really sure of that?)


pvanliefland (approved):
This can be merged I think. If we have the time, let's add logging / sentry for files with size > 0 and a key that ends with /.

PR #142, opened 2022-01-04
Back to GraphQL

pvanliefland:
- A bit of fine-tuning to our existing GraphQL setup
- Removed parts of the Graph that are not stable / usable yet
- Add tests for auth


ekatsah (approved):
Look good to me, but i don't really know graphql :)

les mutations n'ont pas de namespace?

PR #154, opened 2022-02-03
Accessmod GraphQL fine-tuning

pvanliefland:
This PR:

- Adds queries for `FilesetRoles`
- Adds an admin for `FilesetRoles`
- Moves the type defs in separate `.graphql` files
- Fixes case issues (`accessMod` vs `accessmod`)
- Modifies the `prepareAccessmodFileUpload` so that it works with `filesetId`
- Adds a mandatory `projectId` field in the `accessmodFilesets` query field


hexa/plugins/connector_accessmod/schema.py
```diff
@@ -352,4 +216,21 @@ def resolve_delete_accessmod_file(_, info, **kwargs):
     return {"success": True}
 
 
+@accessmod_query.field("accessmodFilesetRole")
+def resolve_accessmod_fileset_role(_, info, **kwargs):
+    try:
+        return FilesetRole.objects.get(id=kwargs["id"])
+    except Fileset.DoesNotExist:
```
qgerome, line 223:
```suggestion
    except FilesetRole.DoesNotExist:
```

qgerome (approved):

qgerome (changes requested):
See comment

qgerome (approved):
I've applied the fix

PR #166, opened 2022-02-20
Fine tune pipelines

ekatsah:
- add a label to datasource injected into a DAG definition, to hint dag builder knows which datasources is what
- move description and sample_config from DAG to DAGTemplate, because it's the same for all dags of the same template
- when an user trigger a DAG instead of the scheduler, add its email to dag_run to send him an email when the dag finish


pvanliefland (approved):
LGTM! A small suggestion regarding the usage of `TextProperty` but can be merged as is

pvanliefland:
Didn't see the failing test - should be easy to fix

PR #168, opened 2022-02-22
HEXA-555 Fix DHIS2 demo

ekatsah:
If locale is none, fallback on english locale and not the first one available. If no english locale, use field value


pvanliefland (approved):
LGTM!

PR #176, opened 2022-02-28
File(set) name & file download

pvanliefland:
This PR:

- [x] Uses the original file name to generate a (slugified) S3 object name
- [x] Adds uniqueness constraints on project, filesets and files
- [ ] Handle non-unique file(set) names gracefully
- [x] Adds a `prepareAccessmodFileDownload` mutation


pvanliefland:
@qgerome for now, if you try to create a fileset (or a project) with a name that is already used, you will just get a `{"success": "false"}` result. I haven't started working on errors yet - let me know if this is blocking.

PR #183, opened 2022-03-06
Analysis / DAGRun signal

pvanliefland:
This PR adds a simple signal for syncing the status of analyses when the state of DAGRuns changes.

Not bulletproof but well-tested and not too complex.


ekatsah (approved):
Technically this PR will do the job.

That being said, signals are kind of an anti pattern in django: they are hard to reason/debug because there is no static link between caller and callee. Maybe there are alternatives?

PR #187, opened 2022-03-11
Inject DHIS2 Credentials into notebooks if permission allows it

ekatsah:
A few weeks ago, we decided to inject DHIS2 credentials into pipeline env. Unfortunately, it was partially solving the problem: the code in papermill to use these credentials differ between development (in notebook) and in automated production/manual run (pipeline). To solve that, I propose to inject DHIS2 credentials into notebooks for specific teams.

Let's discuss: terrible idea? maybe we should use a feature flag to be on a per-user basis?

image

image

```
{
	  "username": "jstilmant@bluesquarehub.com",
	  "env": {
		    "DHIS2_PLAY_URL": "https://play.dhis2.org/demo",
		    "DHIS2_PLAY_USERNAME": "admin",
		    "DHIS2_PLAY_PASSWORD": "district",
		    "DHIS2_WHO_AGGREGATE_URL": "https://who-demos.dhis2.org/demo",
		    "DHIS2_WHO_AGGREGATE_USERNAME": "admin",
		    "DHIS2_WHO_AGGREGATE_PASSWORD": "District#1",
		    "DHIS2_WHO_TRACKER_URL": "https://who-demos.dhis2.org/tracker_demo",
		    "DHIS2_WHO_TRACKER_USERNAME": "TB_Demo",
		    "DHIS2_WHO_TRACKER_PASSWORD": "TBDemo2020!",
		    "DHIS2_INSTANCES_SLUGS": "play,who-aggregate,who-tracker",
		    "POSTGRESQL_HEXA_APP_DEMO_HOSTNAME": "127.0.0.1",
		    "POSTGRESQL_HEXA_APP_DEMO_USERNAME": "test",
		    "POSTGRESQL_HEXA_APP_DEMO_PASSWORD": "test",
		    "POSTGRESQL_HEXA_APP_DEMO_DATABASE": "hexa-app-demo",
		    "POSTGRESQL_HEXA_APP_DEMO_PORT": "5432",
		    "POSTGRESQL_HEXA_APP_DEMO_URL": "postgresql://test:test@127.0.0.1:5432/hexa-app-demo",
		    "POSTGRESQL_DATABASE_NAMES": "hexa-app-demo"
	  },
	  "files": {
		  "~/.pgpass": "MTI3LjAuMC4xOjU0MzI6ZGJfbmFtZTp0ZXN0OnRlc3Q="
	  }
}
```


hexa/plugins/connector_dhis2/models.py
```diff
@@ -213,6 +226,9 @@ def get_absolute_url(self):
 
 class InstancePermission(Permission):
     instance = models.ForeignKey("Instance", on_delete=models.CASCADE)
+    show_credentials = models.BooleanField(
```
pvanliefland, line 229: Let's rename that to `enable_notebooks_credentials` or something better.

hexa/plugins/connector_dhis2/models.py
```diff
@@ -8,6 +8,7 @@
 from django.template.defaultfilters import pluralize
 from django.urls import reverse
 from django.utils import timezone
+from django.utils.text import slugify
```
pvanliefland, line 11: I would prefer if we use `python-slugify` instead (already installed within the project). Django's slugify does not cover some edge cases.

hexa/plugins/connector_dhis2/datacards.py
```diff
@@ -25,6 +26,29 @@ class InstanceSection(Section):
...
+class UsageSection(Section):
+    title = "Code samples"
+
+    usage_python = CodeProperty(
+        label="Usage in Python", code="get_python_usage", language="python"
+    )
+
+    def get_python_usage(self, item: Instance):
+        return """
+%pip install --upgrade git+https://@github.com/blsq/blsq_dqapp.git
```
pvanliefland, line 38: `blsq_dqapp` is too specific to our use cases at Bluesquare. 

Let's rewrite this section using `git@github.com:davidhuser/dhis2.py.git`

```bash
pip install dhis2.py
```

```python
from dhis2 import Api

api = Api('play.dhis2.org/demo', 'admin', 'district')
```

hexa/plugins/connector_dhis2/models.py
```diff
@@ -94,10 +97,20 @@ class Meta:
...
+        if not self.slug:
+            self.slug = slugify(self.name)
+        super().save(*args, **kwargs)
+
     @property
     def display_name(self):
         return self.name if self.name != "" else self.url
 
+    @property
+    def env_name(self):
```
pvanliefland, line 110: Let's rename it to `notebooks_credentials_prefix`

hexa/plugins/connector_dhis2/datacards.py
```diff
@@ -33,6 +57,8 @@ class InstanceCard(Datacard):
 
     external = InstanceSection()
     metadata = OpenHexaMetaDataSection(value="index")
+    # FIXME: this should be shown only if the user has permission to see the credentials
+    usage = UsageSection()
```
pvanliefland, line 61: Let's fix it! We already have a similar use case with datacard `Action`, using the `enabled_when` parameter (see https://github.com/BLSQ/openhexa-app/blob/main/hexa/ui/datacard/actions.py#L25). This could be easily replicated for this use case.

hexa/plugins/connector_dhis2/migrations/0024_instance_slug_instancepermission_show_credentials.py
```diff
@@ -0,0 +1,37 @@
...
+from django.db import migrations, models
+from django.utils.text import slugify
+
+
+def forwards_func(apps, schema_editor):
+
+    Instance = apps.get_model("connector_dhis2", "Instance")
+    db_alias = schema_editor.connection.alias
+    for instance in Instance.objects.all():
+        instance.slug = slugify(instance.name)
```
pvanliefland, line 12: This is function is not called - it should be removed

hexa/plugins/connector_dhis2/migrations/0025_instance_slug_init.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations
+from django.utils.text import slugify
+
+
+def forwards_func(apps, schema_editor):
+
+    Instance = apps.get_model("connector_dhis2", "Instance")
+    db_alias = schema_editor.connection.alias
+    for instance in Instance.objects.all():
+        instance.slug = slugify(instance.name)
```
pvanliefland, line 12: The `slugify` call is unnecessary unless I missed something

hexa/plugins/connector_dhis2/tests/test_credentials.py
```diff
@@ -0,0 +1,73 @@
...
+    def test_credentials_200_regular_user(self):
+        self.client.force_login(self.REGULAR_USER)
+        response = self.client.post(reverse("notebooks:credentials"))
+
+        self.assertEqual(response.status_code, 200)
+        response_data = response.json()
+        self.assertEqual(
+            "instance1",
+            response_data["env"]["DHIS2_INSTANCES_SLUGS"],
+        )
```
pvanliefland, line 73: We should add the negative case (when user has not the setting to include credentials)

pvanliefland (changes requested):
I think it's a good idea. And the feature is almost ready.

I have a few requests about naming and a few things to iron out here and there.

pvanliefland:
Another small thing we can improve: I think it would be good to re-generate a single migration when the PR is ready (not just this one, as a best practice going forward)

ekatsah:
About migration: usually i try to separate data migration and schema migration (best practise), but i can merge the first schema migration and the data mig. You can't merge all three, because of pending trigger after the data mig (doesn't fit in the migration transaction)

PR #188, opened 2022-03-14
HEXA-510 S3 credentials for pipelines vs user permissions

ekatsah:
PR to have the same environment variables when running papermill in a pipeline as when running it in a notebook:
- Each plugin exposes its credentials mostly like they do for the notebooks but uses the DAGAuthorizedDatasource instead of the permissions from the user’s team.
- If a DAGAuthorizedDatasource  has a label, this label overrides the name of the original datasource. This enables pipelines (not papermill) to recieve generic datasources (eg.  a database named hexa-cmr-msp  with an DAGAuthorizedDatasource  labeled input  will expose a variable POSTGRESQL_INPUT_PASSWORD to the pipeline).

It will close the current problem where a running pipeline has access to every S3 bucket. Now each DAG has an associated role in IAM and the roles receives permissions only to use the buckets specified with DAGAuthorizedDatasource. This role is then sent to airflow and when running the DAG, airflow calls “assume role” to limit the scope.

WIP: to merge only when airflow side is reviewed and OK


hexa/plugins/connector_postgresql/apps.py
```diff
@@ -11,6 +11,10 @@ class PostgresqlConnectorConfig(ConnectorAppConfig):
         "hexa.plugins.connector_postgresql.credentials.notebooks_credentials"
     ]
 
+    PIPELINES_CONFIGURATION = [
```
pvanliefland, line 14: `PIPELINES_CREDENTIALS`

hexa/plugins/connector_s3/apps.py
```diff
@@ -11,6 +11,10 @@ class S3ConnectorConfig(ConnectorAppConfig):
         "hexa.plugins.connector_s3.credentials.notebooks_credentials"
     ]
 
+    PIPELINES_CONFIGURATION = [
```
pvanliefland, line 14: `PIPELINES_CREDENTIALS`

pvanliefland (changes requested):
A few requests:

- Rename `pipelines_configuration`  / `PipelinesConfiguration` / `get_configuration` to `pipelines_credentials` etc everywhere
- Rename `AuthorizedDataSource.label` to `code` and prevent punctation / special chars or slugify
- In the same vein, rename instance `slug` to `code`
-  Rename `credentials` to `env` and `connector_config` to `configuration` or `base_configuration`, `extra_configuration`, whatever ;)
- Check test without test methods

PR #191, opened 2022-03-15
Fix Codespaces Configuration

qgerome:


pvanliefland:
@qgerome the config was still not good (`isort` was failing during linting). I just did a small change in the `devcontainer.json`, can you test?

pvanliefland (approved):
LGTM!

PR #197, opened 2022-03-24
Add trailing slash to all urls

qgerome:
This is needed to make the proxy made y the new frontend work


pvanliefland (changes requested):
@ekatsah @qgerome needs those changes to make his NextJS proxy setup work - fine by me, the only things that might break are webhook calls.

IIRC `requests` follows redirects but we should test it. What do you think?

PR #198, opened 2022-03-28
AC-42: User should be able to reset their password

qgerome:
I'm not sure yet how to manage the second step which is to use the token received in the mail to set a new password. Django handles that nicely by putting the token in a cookie instead of the url when it returns the page.


qgerome:
Second part to use token via Graphql is pushed.

PR #201, opened 2022-03-30
Teams GraphQL query

pvanliefland:
This PR simply adds the `teams` and `team($id)` query, along with some adaptations in models (in particular the `role` field on `Membership`).


hexa/user_management/graphql/schema.graphql
```diff
@@ -14,6 +9,33 @@ type Avatar {
...
+    ADMIN
+    REGULAR
+}
+type MembershipPage {
+    pageNumber: Int!
+    totalPages: Int!
+    totalItems: Int!
+    items: [Membership!]!
+}
+type Membership {
```
qgerome, line 33: I'm wondering if `TeamMembership` would not be better but it's just a thought.

Also I'd add `createdAt`, `updatedAt` on this relation, wdyt ?

pvanliefland: For `TeamMembership`, not now - the model is called `Membership` as well.

I've added the `createdAt` / `updatedAt` fields as suggested

PR #202, opened 2022-04-04
User, teams and permissions

pvanliefland:
This PR:

- [ ] Creates a new [Django permissions backend](https://docs.djangoproject.com/en/4.0/topics/auth/customizing/#handling-authorization-in-custom-backends) (more on this below)
- [ ] Adds a `permissions.py` module in the AccessMod connector that demonstrates the new backend
- [ ] Cleans up a lot of obsolete / misplaced code regarding user & permissions

## Rationale

TBC


qgerome:
I'll put here all my remarks/notes for your return:

- `createMembership`: I understood user will add email addresses that will be invited to join teams. Matching between users & emails addresses is a backend matter. In the current state, I'd need to pass a `userId` but I cannot look for users in the graphql. Which solution do we want ? - look for users to add them; - Invite email address & create users on the fly if needed

hexa/user_management/permissions.py
```diff
@@ -0,0 +1,49 @@
...
+    return principal.is_admin_of(team)
+
+
+def create_membership(
+    principal: User,
+    team: Team,
+):
+    """We allow the creation of a membership in two cases:
+    1. If the principal is the team admin
+    2. If the team has no admin yet (for newly created teams)
```
qgerome, line 28: Shouldn't we make the creator of the team the admin of it ?

PR #203, opened 2022-04-07
HEXA-612: Implement Data collections prototype

qgerome:
The collections' pages are served by the new frontend (openhexa-frontend)


pvanliefland:
@qgerome actually let's talk about this one when you have 10 minutes - I have a suggestion for the menu

PR #208, opened 2022-04-26
AM-173 add validation background worker

ekatsah:
Notes:
- contains duplicated code with #202, need to undup for merging
- @qgerome added a status for "in progress"


hexa/plugins/connector_accessmod/queue.py
```diff
@@ -0,0 +1,34 @@
...
+
+def validate_fileset_job(queue, job):
+    try:
+        fileset = Fileset.objects.get(id=job.args["fileset_id"])
+    except Fileset.DoesNotExist:
+        logger.error("fileset %s not found", job.args["fileset_id"])
+    else:
+        fileset.status = FilesetStatus.VALID
+        fileset.save()
+        sleep(3)
```
pvanliefland, line 19: Shouldn't the sleep() come before the save()?

hexa/plugins/connector_accessmod/schema.py
```diff
@@ -289,7 +291,16 @@ def resolve_create_accessmod_file(_, info, **kwargs):
             mime_type=create_input["mimeType"],
             fileset=fileset,
         )
+        fileset.status = FilesetStatus.VALIDATING
```
pvanliefland, line 294: I would move that to the validation worker - until it's picked up by the queue, it should stay in `PENDING` imho

pvanliefland (changes requested):
Nice! Just 2 suggestions on my side

PR #209, opened 2022-04-26
Add POC gcs-fuse support

ekatsah:
Don't merge, it's only for comments only

Notes:
- add prototype of GCS support
- remove feature flag S3FS (we can't go back anyway)


hexa/plugins/connector_gcs/models.py
```diff
@@ -0,0 +1,174 @@
...
+from hexa.core.models.cryptography import EncryptedTextField
+from hexa.user_management import models as user_management_models
+from hexa.user_management.models import Permission, PermissionMode
+
+logger = getLogger(__name__)
+
+
+class Credentials(Base):
+    """We actually only need one set of credentials. These "principal" credentials will be then used to generate
+    short-lived credentials with a tailored policy giving access only to the buckets that the user team can
```
pvanliefland, line 24: I don't see the "short-lived" part - don't we use the full credentials every time at the moment?

hexa/plugins/connector_gcs/models.py
```diff
@@ -0,0 +1,174 @@
...
+    def filter_for_user(
+        self,
+        user: typing.Union[AnonymousUser, user_management_models.User],
+        mode: PermissionMode = None,
+        mode__in: typing.Sequence[PermissionMode] = None,
+    ):
+        if mode is not None and mode__in is not None:
+            raise ValueError('Please provide either "mode" or "mode_in" - not both')
+
+        if user.is_superuser or user.has_feature_flag("gcs"):
```
pvanliefland, line 65: If we don't filter the buckets at all, I would prefer if we dropped the `has_feature_flag` part and restrict the GCS plugin to super admins.

hexa/plugins/connector_gcs/models.py
```diff
@@ -0,0 +1,174 @@
...
+        index.datasource_id = self.id
+
+    @property
+    def display_name(self):
+        return self.name
+
+    def __str__(self):
+        return self.display_name
+
+    def writable_by(self, user):
```
pvanliefland, line 122: I know that this is what we have for S3 as well, but could you try using the new permissions for this part? And maybe adapt the S3 code as well?

pvanliefland (changes requested):
Great, this is really cool. I have a few minor question / remarks, let's check them together.

pvanliefland (approved):

PR #211, opened 2022-04-26
Add public bucket support

ekatsah:
Small PR to support public bucket workflow in openHexa:
- [x] all users can see public buckets in data catalog
- [x] all read views works on public buckets
- [x] download url is a public, external, sharable URL


pvanliefland (approved):
This is 👌

PR #212, opened 2022-04-27
Fix S3 bucket permissions mess

pvanliefland:


hexa/plugins/connector_s3/migrations/0030_permission_mode_data.py
```diff
@@ -0,0 +1,22 @@
...
+
+from django.db import migrations
+
+from hexa.user_management.models import PermissionMode
+
+
+def update_bucket_permissions(apps, schema_editor):
+    BucketPermission = apps.get_model("connector_s3", "BucketPermission")
+    BucketPermission.objects.filter(mode="RW").update(mode=PermissionMode.EDITOR)
+    BucketPermission.objects.filter(mode="RO").update(mode=PermissionMode.VIEWER)
```
ekatsah, line 11: Bad practice to depend on the value of a enum which can change in the future. A migration is frozen in time, maybe use a literal value?

pvanliefland: Good call, fixed

ekatsah (approved):

PR #214, opened 2022-05-03
Fix for permission conflicts in S3 bucket

pvanliefland:
This PR fixes a bad query that was causing conflicts within bucket permissions - especially when generating notebook credentials.

### DIFF
--- a/hexa/plugins/connector_s3/credentials.py
+++ b/hexa/plugins/connector_s3/credentials.py
@@ -23,15 +23,15 @@ from hexa.user_management.models import PermissionMode
 def notebooks_credentials(credentials: NotebooksCredentials):
     """Provides the notebooks credentials data that allows users to access S3 buckets in the notebooks component."""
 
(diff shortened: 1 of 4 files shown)


ekatsah (changes requested):
The approch seems good to me, but i would suggest that the behavior of ```mode``` should be the same as ```mode__in``` also, for example:

```
In [6]: Bucket.objects.filter_for_user(u, mode="VIEWER")
Out[6]: <BucketQuerySet [<Bucket: hexa-ner-msp>]>

In [7]: Bucket.objects.filter_for_user(u, mode__in=["VIEWER"])
Out[7]: <BucketQuerySet [<Bucket: hexa-metrics>, <Bucket: hexa-ner-msp>]>
```
(same kind of user/teams/buckets setup as this afternoon)

pvanliefland:
@ekatsah should be fixed in 6a9e4f8

ekatsah (approved):

PR #217, opened 2022-05-12
Validate Accessmod fileset

ekatsah:


hexa/plugins/connector_accessmod/queue.py
```diff
@@ -1,29 +1,220 @@
...
+    # if raster.bounds != fileset.project.extend.metadata: i guess
+    #     fileset.set_invalid("wrong extend")
+    #     return
+
+
+def validate_health_facilities(fileset: Fileset, filename: str):
+    facilities = gpd.read_file(filename)
+    validate_geopkg(facilities, fileset, ("Point",))
+
+    if fileset.status == FilesetStatus.VALIDATING:
```
qgerome, line 70: We need to extract gpkg columns and put them in the metadata.

PR #221, opened 2022-05-13
Fileset to acquire

qgerome:


hexa/plugins/connector_accessmod/schema.py
```diff
@@ -422,6 +422,7 @@ def resolve_create_accessmod_fileset(_, info, **kwargs):
                 id=create_input["projectId"]
             ),
             role=FilesetRole.objects.get(id=create_input["roleId"]),
+            status=create_input["status"] if "status" in create_input else None,
```
ekatsah, line 425: No input validation? Does it make sense for the front to be able to set a file as validating/valid/invalid for example?

ekatsah (approved):

hexa/plugins/connector_accessmod/schema.py
qgerome: I changed this to not use the status on the frontend.

PR #222, opened 2022-05-15
change of DAG run names  

wagzi:
changes in dag names and dates format


pvanliefland:
Hello @wagzi ! Great work on your first PR!

However, as you can see, the automated checks have failed.

Before I look at this PR, could you :

- Reformat your code to make sure it conforms to our code style (see "Code Style" in `README.md`)
- Run the tests locally and make sure that they succeed
- Rephrase your commit in English (I'll show you how) - we never use French for code, comments or commits
- Re-open a PR with an English title and description

pvanliefland (changes requested):
Could you please:

- [x] Rename the PR with an English title & description
- [ ] Fix the failing test
- [ ] Remove all the PyCharm-generated files (in `.idea`)

hexa/plugins/connector_airflow/datagrids.py
```diff
@@ -39,24 +40,30 @@ class DAGRunGrid(Datagrid):
...
 
     view = LinkColumn(text="View")
 
     @staticmethod
     def get_label(run: DAGRun) -> str:
         if hasattr(run, "favorite") and getattr(run, "favorite") is not None:
             return getattr(run, "favorite")
-
-        return run.run_id
+        elif hasattr(run, "run_id") and re.match(r"^manua\w", getattr(run, "run_id")):
```
pvanliefland, line 54: Please use `run.run_id.startswith()` instead of `re.match`. Simpler is always better than complex ;)

hexa/ui/datagrid.py
```diff
@@ -309,29 +307,11 @@ def context(self, model: DjangoModel, grid: Datagrid):
         return data
```
pvanliefland, line 309: What if I want to use `BooleanColumn` in another grid? In this case, you shouldn't delete it just because you don't use it ;). Sometimes, it's the way to go but not in this case.

hexa/ui/datagrid.py
```diff
@@ -309,29 +307,11 @@ def context(self, model: DjangoModel, grid: Datagrid):
...
-            "value": self.get_value(model, self.value, container=grid),
-        }
-
-
 class DateColumn(Column):
     """Date column, with one or two rows"""
 
     def __init__(
-        self, *, date=None, date_format="timesince", secondary_text=None, **kwargs
+        self, *, date=None, date_format="%Y-%m-%d %H:%M", secondary_text=None, **kwargs
```
pvanliefland, line 314: Don't change the default value of `date_format`. Other grids use this column as well and rely on the existing default.

In this case, your goal it to use a specific format, not change the default for all other use cases.

hexa/ui/datagrid.py
```diff
@@ -347,11 +327,6 @@ def format_date(self, date: datetime.datetime):
         if date is None:
             return date
```
pvanliefland, line 329: Please don't do this :)

In this case, you are probably breaking `DateColumn` for existing usages. Just change it for `DAGRunGrid`, not everywhere...

hexa/ui/tests/test_datagrid.py
```diff
@@ -24,10 +24,8 @@ class BikeDatagrid(Datagrid):
             )
             category = TextColumn(text="category")
             characteristics = TextColumn(text="suspension", secondary_text="group_set")
-            created_at = DateColumn(
-                date="created_at", date_format="%Y-%m-%d %H:%M:%S %Z"
-            )
-            launch_in = DateColumn(date="launch_in", date_format="timesince")
```
pvanliefland, line 30: The previous test was testing two different usages: one with a date format string, the other with `timesince`. Don't remove `timesince` and its test.

Existing tests should be changed very carefully, as you might get the wrong impression that "you are not breaking existing features because all tests pass". By changing the tests, you are taking a risk : breaking an existing feature without noticing it.

(I'm not saying that you can't change an existing test - just that you need to think about it carefully)

pvanliefland: However, in this case, I think this test should not be changed at all.

pvanliefland (changes requested):
Hey @wagzi , a couple of things to change - and to reflect on :)

pvanliefland (approved):
LGTM!

PR #226, opened 2022-05-17
Add default CRS on country in Graph

pvanliefland:
@yannforget @qgerome if we have to hardcode CRS, I'd prefer to hardcode it in the backend.

@yannforget what other "quick wins" can I implement until we have a table with country data?

Checklist:

- [x] Simple `defaultCRS`field with default value
- [ ] Check appropriate fallback value
- [ ] Add query for WHO regions

### DIFF
--- a/hexa/user_management/schema.py
+++ b/hexa/user_management/schema.py
@@ -256,15 +256,23 @@ country_object = ObjectType("Country")
 
 
 @country_object.field("alpha3")
-def resolve_alpha3(obj: Country, *_):
+def resolve_country_alpha3(obj: Country, *_):
     return obj.alpha3
 
 
 @country_object.field("flag")
-def resolve_flag(obj: Country, info):
+def resolve_country_flag(obj: Country, info):
     request: HttpRequest = info.context["request"]
 
     return request.build_absolute_uri(obj.flag)
+
+
+@country_object.field("defaultCRS")
+def resolve_country_default_crs(obj: Country, info):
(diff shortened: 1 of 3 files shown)


hexa/user_management/schema.py
```diff
@@ -256,17 +256,25 @@ def resolve_avatar(obj: User, *_):
...
 
     return request.build_absolute_uri(obj.flag)
 
 
+@country_object.field("defaultCRS")
+def resolve_country_default_crs(obj: Country, info):
+    if obj.alpha3 == "BFA":
+        return 6933
+
+    return 4326
```
qgerome, line 275: I'm not sure 4326 is a valid EPSG since it's degree based (https://epsg.io/4326). @yannforget Could you confirm ?

qgerome:
> @yannforget @qgerome if we have to hardcode CRS, I'd prefer to hardcode it in the backend.
> 
> @yannforget what other "quick wins" can I implement until we have a table with country data?

It would be awesome to also have the WHO region in this list 🙏🏻 . As of now I have to maintain this list in the frontend as well just because of this.

pvanliefland:
> > @yannforget @qgerome if we have to hardcode CRS, I'd prefer to hardcode it in the backend.
> > @yannforget what other "quick wins" can I implement until we have a table with country data?
> 
> It would be awesome to also have the WHO region in this list 🙏🏻 . As of now I have to maintain this list in the frontend as well just because of this.

Will do :)

PR #227, opened 2022-05-18
Country default CRS, WHO regions & fileset modes

pvanliefland:
This PR:

- [x] Adds the `defaultCRS` and `whoRegion` fields on the `Country` type in GraphQL
- [x] Adds a basic (and simplistic) logic to compute the value of those fields
- [x] Adds a `mode` property on the `Fileset` model (either `USER_INPUT` or `AUTOMATIC_ACQUISITION`)
- [x] Adds the mode field in the graph
- [x] Unskips a few test in the AccessMod plugin

Still to do:

- [x] Check appropriate fallback value for `defaultCRS`
- [x] Check FIXME in fileset test (validation queue)


ekatsah (approved):
looks very good to me

PR #244, opened 2022-05-25
Update fileset mutation

pvanliefland:
☢️ the PR also makes all the file-related mutation inputs mandatory, so this needs to be changed in the frontend as soon as we merge the PR.


ekatsah:
I am not against running the validation after an update.

If I understand correctly the PR, the API enable a update of the name, role, metadata but not the file content, right?
- Why the user would be able to change the role of a fileset? If the fileset is the result of an analysis or acquisition, can the user update the role, does it makes sense?
- For the metadata, i don't think it's a good idea to have a field "owned"/writed by several components, especialy if one of the writer is the user. In this case, it should at least be part of the field (only some keys in the metadata should be writable by the user, with some maxlen). If we don't do that, we will have the user, data worker and analysis pipeline overwriting each other and it will be hell to debug.

pvanliefland:
@ekatsah those are good remarks. I think that the use case @qgerome had in mind is to update an invalid fileset (because of  bad metadata but I can think of wrong role as well if uploaded at the project level) - in order to make it valid again.

Should we restrict that mutation to manually-uploaded, invalid filesets ?

pvanliefland:
(personally I would also consider deleting invalid filesets)

qgerome:
I'd skip the role in the update mutation. I think it's better to just delete the fileset and create a new one.
Regarding the metadata: I was more thinking of the `column_category` but that's my only example and since the values are already present, I don't even think it's needed to validate it again.

So maybe it's better to not trigger a new background job and just update the DB until we have a clear use case that would require it.

ekatsah:
> Regarding the metadata: I was more thinking of the `column_category`

Yeah that's perfect. Maybe the backend should check that it's the only field "touched" by the user. Should we could restrict the keys present from the user input and do an `metadata.update(user_input.metadata)` instead of a `setattr()`?

pvanliefland:
@qgerome @ekatsah I've simplified the update mutation :

- Only name and metadata can be updated
- Does not enqueue validation anymore

I haven't restricted the allowed metadata changes yet, let's finish this task on the frontend side first.

hexa/plugins/connector_accessmod/graphql/schema.graphql
```diff
@@ -253,6 +253,23 @@ enum CreateAccessmodFilesetError {
     NAME_DUPLICATE
     PERMISSION_DENIED
 }
+input UpdateAccessmodFilesetInput {
+    id: String!
+    name: String
+    roleId: String
```
qgerome, line 259: You can remove this if we can't change the role anymore

PR #251, opened 2022-05-31
Membership tests

pvanliefland:
@qgerome should be fine for the frontend, but could you check please?


pvanliefland:
(still need to fix a test before merging)

PR #267, opened 2022-06-22
Change default value for accessibility algorithm

yannforget:
Isotropic instead of anisotropic


pvanliefland:
@yannforget could you juste update the `test_accessmod_analysis_owner` test method to reflect your change?

PR #271, opened 2022-06-25
Plugin GCS: add download link in catalog

ekatsah:
Didn't find a simpler way...


pvanliefland (changes requested):
I think we should use [python-storage](https://github.com/googleapis/python-storage), Google's official python client for GCS.

There's a straightforward way of generating signed URLs documented here: https://googleapis.dev/python/storage/latest/storage/blobs.html#google.cloud.storage.blob.Blob.generate_signed_url

ekatsah:
rewrited in another PR

PR #275, opened 2022-07-03
Support GCS in AM and GCS download/upload links

ekatsah:


pvanliefland (changes requested):
This is really cool, just a couple of minor remarks for the GraphQL part

PR #285, opened 2022-07-08
After logout, redirect to JupyterHub's logout page 

ekatsah:
who will redirect us back to our login page ounce his job is done

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -43,6 +43,9 @@ urlpatterns = [
     path("metrics/", include("hexa.metrics.urls", namespace="metrics")),
     path("comments/", include("hexa.comments.urls")),
     # Order matters, we override the default logout view defined later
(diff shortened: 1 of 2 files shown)


config/urls.py
```diff
@@ -41,6 +42,12 @@
     path("pipelines/", include("hexa.pipelines.urls", namespace="pipelines")),
     path("metrics/", include("hexa.metrics.urls", namespace="metrics")),
     path("comments/", include("hexa.comments.urls")),
+    # Order matters, we override the default logout view defined later
```
pvanliefland, line 45: We should just add 1-2 sentences of explanation here : why do we do this? And Add a `  #TODO: use API when available https://github.com/jupyterhub/jupyterhub/issues/3688`)

pvanliefland (changes requested):

PR #293, opened 2022-07-20
Refactor s3 credentials, inject fuse config into pipeline env vars

ekatsah:
Preliminary work to support fuse mounts inside pipeline environment

Also: increase delay before IAM.assumeRole, should fix the "first dag run always crash"


pvanliefland (approved):
LGTM! Maybe consider aligning notebooks / credentials generation a bit more, for the `sleep` instruction for example.

PR #299, opened 2022-08-03
HEXA-809 Inject IASO user API token into notebooks

ekatsah:
- Add org units + sync org unit level 1/2 for IASO
   - left the IASO prefix because while individual model are namespaced, related manager are not, and we already have org unit elsewhere. i don't mind changing this later, let's discuss it at some point?
- Add API Token + injection in notebook credentials


hexa/plugins/connector_iaso/datacards.py
```diff
@@ -48,3 +65,47 @@ class FormCard(Datacard):
     @property
     def generic_description(self):
         return _("IASO Form")
+
+
+class IASO_OU_ID_Property(Property):
```
pvanliefland, line 70: `OrganisationUnitProperty` ;)

hexa/plugins/connector_iaso/models.py
```diff
@@ -55,20 +56,23 @@ def get_permission_set(self):
         return self.iasopermission_set.all()
 
     def sync(self):
-        result = get_forms_json(iaso_account=self)
+        iasofroms = get_forms_json(iaso_account=self)
```
pvanliefland, line 59: `forms` ?

hexa/plugins/connector_iaso/models.py
```diff
@@ -55,20 +56,23 @@ def get_permission_set(self):
...
+        iasoorgunits = get_orgunits_level12(iaso_account=self)
```
pvanliefland, line 60: `org_units` ?

hexa/plugins/connector_iaso/models.py
pvanliefland: Not only the iaso prefix, the typo too

pvanliefland (approved):
LGTM, except for some names ;)

Out of curiosity, in which case does the related manager names could conflict?

PR #301, opened 2022-08-04
Collections in list and detail

qgerome:


hexa/data_collections/models.py
```diff
@@ -65,6 +65,9 @@ def delete_if_has_perm(
 
         self.delete()
 
+    def get_absolute_url(self) -> str:
```
pvanliefland, line 68: This will work but we typically use `reverse()` from `django.urls` for this.

hexa/data_collections/properties.py
```diff
@@ -0,0 +1,19 @@
+from hexa.ui.datacard.properties import Property
```
pvanliefland, line 1: We could move this to `datacards.py` in the same module.

pvanliefland (changes requested):
Nice, just a few changes and this can be merged 👍

hexa/data_collections/models.py
qgerome: In this case we do not have urls for collections. Do we create stub urls for collections while we implement them in the new frontend ?

If not, what could I use as params in `reverse()` ?

hexa/data_collections/properties.py
qgerome: Yes

pvanliefland (approved):

PR #305, opened 2022-08-10
Improve datagrid width

qgerome:
I'm not really confortable with the default width of the `LinkColumn` but it's better to limit the size of this cell to leave more room for the other as all the cells without explicit width receives an equal width.


pvanliefland (approved):
Let's merge this and test it on the demo environment

PR #322, opened 2022-08-30
Fix fileset status overwritten

yannforget:
`process_raster()` function should end if fileset status is invalid


ekatsah (changes requested):
Can you add a test with an small but invalid raster in the repo?

ekatsah (approved):

PR #329, opened 2022-09-07
Update index permission for a datasource and all objects

ekatsah:
- When user add, delete or update a DatasourcePerm, the IndexPermission of the DS entry is updated, but not the IndexPermission of all its entries.
- Since some datasource have a huge number of entries, need to update async (test for 10k entries: > 1min)
- Rework the sync_worker to include another task: datasource_index, which will update permissions


pvanliefland (approved):
LGTM, let's deploy this in demo and test it!

PR #334, opened 2022-09-14
HEXA-873 Extend connector_airflow webhook

ekatsah:


hexa/plugins/connector_airflow/models.py
```diff
@@ -540,6 +552,36 @@ def is_in_favorites(self, user: User):
         except DAGRunFavorite.DoesNotExist:
             return False
 
+    def log_message(self, priority: str, message: str):
+        self.messages.append({"priority": priority, "message": message})
+        self.save()
+
+    def set_output(self, uri: str):
```
pvanliefland, line 559: I would suggest to add a message / title to the output as well

pvanliefland (approved):
LGTM !

I'm wondering whether we could have dedicated model for outputs and messages, but this can definitely wait for a bigger refactoring.

PR #346, opened 2022-10-06
feat(Features): Add features in graphql

qgerome:


hexa/user_management/graphql/schema.graphql
```diff
@@ -235,6 +235,16 @@ input OrganizationInput {
     url: String
     contactInfo: String
 }
+
+type FeatureFlag {
+    code: String!
+    config: JSON!
+}
+
+extend type Me {
```
pvanliefland, line 244: `Me` is defined just a few lines above - why extend it?

pvanliefland (approved):
LGTM - just a small remark

hexa/user_management/graphql/schema.graphql
qgerome: Because I didn't see it 🙈

PR #347, opened 2022-10-12
Fix sentry traces

qgerome:


pvanliefland (approved):
Let's test this on demo

PR #350, opened 2022-10-24
feat(Features) : Add dashboard visualization to graphql api

cheikhgwane:


qgerome:
@cheikhgwane Je pense que tu as encore des soucis avec ta config lint

cheikhgwane:
> @cheikhgwane Je pense que tu as encore des soucis avec ta config lint

Je suis en train de voir ça.

hexa/visualizations/admin.py
```diff
@@ -24,7 +24,7 @@ class IndexAdmin(admin.ModelAdmin):
     list_filter = ("content_type",)
     search_fields = ("label", "content")
 
-    def app_label(self, obj):
+    def app_id(self, obj):
```
qgerome, line 27: Why did you change `app_label` to `app_id` ?

hexa/visualizations/schema.py
```diff
@@ -0,0 +1,73 @@
...
+        .order_by("-updated_at")
+    )
+
+    return result_page(queryset=queryset, page=page, per_page=per_page)
+
+
+@dashboard_queries.field("externalDashboard")
+def resolve_external_dashboard(_, info, **kwargs):
+    try:
+        request = info.context["request"]
```
qgerome, line 63: This can be above the `try..catch` as it is not what you'd like to test

qgerome (changes requested):

hexa/visualizations/admin.py
cheikhgwane: It was  a test in order to make the label clickable on the admin. But I'll remove it from now.

hexa/visualizations/schema.py
```diff
@@ -0,0 +1,73 @@
...
+    return reverse(
+        "visualizations:dashboard_image", kwargs={"dashboard_id": dashboard.id}
+    )
+
+
+dashboard_queries = QueryType()
+
+
+@dashboard_queries.field("externalDashboards")
+def resolve_external_dashboards(_, info, page=1, per_page=15):
```
qgerome, line 49:
```suggestion
def resolve_external_dashboards(_, info, page=1, perPage=15):
```

cheikhgwane: It's updated

qgerome:
It would be nice to have a test where there are more than two pages of items

qgerome (approved):

PR #360, opened 2022-11-14
connector_s3: add MinIO support

ekatsah:
Small changes to the S3 connector to support MinIO blob storage server. For MinIO, the backend should inform the region and the endpoint to boto3 or fuse. This PR adds those fields in the data model.


hexa/plugins/connector_s3/api.py
```diff
@@ -119,6 +125,15 @@ def generate_sts_user_s3_credentials(
         5. Assume the team/pipeline/.. role
     """
 
+    if principal_credentials.endpoint_url:
+        # We are pointing to a MinIO instance, keep it simple
+        return (
```
pvanliefland, line 130: This would grant access to all buckets in the minio instance?

pvanliefland (commented):
Not sure I understand the changes and their impact. Could you add a bit more context?

hexa/plugins/connector_s3/api.py
ekatsah: Same as S3. Openhexa support only one blob storage provider at a given time

pvanliefland: My question was rather "don't we want to limit the buckets that the user can access" ? 

Especially because we are in the `generate_sts_user_s3_credentials`, which gives the impression that we are generating credentials that are specific to the user.

If it's a temporary solution to test minio, then ok by me, but let's add a TODO in the code and an issue in our backlog.

hexa/plugins/connector_s3/api.py
```diff
@@ -279,7 +294,11 @@ def _build_app_s3_client(*, principal_credentials: models.Credentials):
     )
     return boto3.client(
         "s3",
-        principal_credentials.default_region,
+        endpoint_url=(
```
pvanliefland, line 297: Do we really want to remove the region name here?

hexa/plugins/connector_s3/models.py
```diff
@@ -51,6 +51,7 @@ class Meta:
     )
     user_arn = models.CharField(max_length=200)
     app_role_arn = models.CharField(max_length=200)
+    endpoint_url = models.CharField(blank=True, max_length=200)
```
pvanliefland, line 54: Could you add a `help_text` to indicate that this is not needed for S3?

pvanliefland (commented):
It's nice to see that we can adapt our connector for minio. Let's discuss the "user sts credentials" for minio.

pvanliefland (approved):
Alright, let's start with this, and plan better credentials handling for later

PR #361, opened 2022-11-16
Add a scheduler for airflow job

ekatsah:
This PR add a background worker in the connector_airflow, which will trigger the start of a dag based on the schedule set by an openhexa admin. Previously, the schedule was forwarded by the sync to airflow, and airflow started a DAG based on its own rules. Unfortunately, this method was challenging from an architectural point of view because it means a DagRun can be 2 types: manual and scheduled. Now, every run are manual (from an airflow point of view), only some are triggered by users and others are called by the background worker.

## Changes

- Add a background worker
- Update cluster sync

## How/what to test

Schedule pipelines, check if they are launched

### DIFF
--- a/hexa/plugins/connector_airflow/api.py
+++ b/hexa/plugins/connector_airflow/api.py
@@ -42,14 +42,17 @@ class AirflowAPIClient:
         return response.text
 
     def trigger_dag_run(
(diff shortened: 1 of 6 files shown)


pvanliefland (approved):
Nice! LGTM 👍

( just looked at the code, didn't have the time to test it locally though)

PR #363, opened 2022-11-24
Remove ENABLE_GRAPHQL setting

mbayangdiop95:
GraphQL is not optional / experimental anymore.

## Changes

 - Remove ENABLE_GRAPHQLsetting  from `config/settings.py`
 - Remove ENABLE_GRAPHQL env variable from `docker-compose.yaml`

## How/what to test

- Build a new image
- Test on localhosting
- Deploy in demo and prod

### DIFF
--- a/config/urls.py
+++ b/config/urls.py
@@ -61,9 +61,6 @@ urlpatterns = [
         GraphQLView.as_view(
             schema=schema, playground_options={"request.credentials": "include"}
         )
-        if settings.ENABLE_GRAPHQL is True
-        else TemplateView.as_view(template_name="404.html"),
-        name="graphql",
     ),
 ]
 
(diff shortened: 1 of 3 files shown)


config/urls.py
```diff
@@ -61,9 +61,6 @@
         GraphQLView.as_view(
             schema=schema, playground_options={"request.credentials": "include"}
         )
-        if settings.ENABLE_GRAPHQL is True
-        else TemplateView.as_view(template_name="404.html"),
-        name="graphql",
```
qgerome, line 66: I think this should be kept

qgerome (changes requested):

config/urls.py
mbayangdiop95: I put it back

qgerome: I've fixed the line.

PR #365, opened 2022-12-02
feat(Catalog) : show/hide postgreSQL catalog url

cheikhgwane:
Display or hide PostgreSQL url on database  detail page.

## Changes

- add hide/show icon 

## How/what to test

Go to PostgreSQL detail on catalog view and check the hide/show feature works.

## Screenshots / screencast
image
image


pvanliefland (changes requested):
Let's get rid of the console.log and this can be merged

hexa/static/js/hexa.js
```diff
@@ -173,6 +173,7 @@ function Editable(key, value, originalValue) {
                 (originalValue !== "" ? originalValue : "-");
         },
         toggle() {
+            console.log("edit called")
```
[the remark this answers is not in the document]
cheikhgwane: My bad.

pvanliefland (approved):
LGTM!

hexa/ui/templates/ui/datacard/property_text.html
```diff
@@ -1,11 +1,30 @@
...
-    {% if markdown %}
-        <div class="prose prose-sm" {% if not translate %}translate="no"{% endif %}>
-            {{ text|default:"-" }}
-        </div>
+{% if markdown %}
+<div class="prose prose-sm" {% if not translate %}translate="no" {% endif %}>
+    {{ text|default:"-" }}
+</div>
+{% else %}
+<div x-data="SecretSection()">
```
qgerome, line 9: It's more `SecretProperty` than `SecretSection`

qgerome: I think it's better to not instantiate the `SecretSection` if the property is not a secret one. (So only called `SecretSection` if `{%secret%}`

hexa/static/js/hexa.js
```diff
@@ -423,3 +423,16 @@ function AdvancedSearch(initialQuery) {
...
 }
+
+/**
+ * 
+ * Display text property secret
+ */
+function SecretSection() {
+    return {
+        secret: true,
+        showSecret($event) {
```
qgerome, line 434: Based on the implementation, it's more a `toggleSecret` than a `showSecret` since it handles the two use cases.

hexa/ui/templates/ui/datacard/property_text.html
```diff
@@ -1,11 +1,30 @@
...
+    <p x-show="secret" class="flex w-fit">
+        <span class="text-md text-blue-600 hover:text-blue-500 focus:outline-none cursor-pointer"
+            @click="showSecret()">
+            {% include "ui/icons/eye.html" with w=4 h=4 ml=1 %}
+        </span>
+    </p>
+    <p x-show="!secret" class="flex gap-2" {% if not translate %}translate="no" {% endif %}>{{ text|default:"-" }}
+        <span class="text-md text-blue-600 hover:text-blue-500 flex focus:outline-none cursor-pointer"
+            @click="showSecret()">
+            {% include "ui/icons/eye_slash.html" with w=4 h=4 ml=1 %}
```
qgerome, line 20: There is already a margin left applied.

hexa/ui/templates/ui/icons/eye.html
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-6 h-6">
```
qgerome, line 1: Can you check how to make them customisable with variables please.

hexa/ui/templates/ui/icons/eye_slash.html
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-6 h-6">
```
qgerome, line 1: Can you check how to make them customisable with variables please.

hexa/ui/templates/ui/datacard/property_text.html
qgerome: Well, it's not since the variables are not passed to the icon. (Better to change the icons to have the same variables than the others.

qgerome (changes requested):
If the property has a value, I'd display "***********" instead of nothing when the field is in secret mode.

I'm not sure it's a good idea to use the primary color on the icon to show/hide the secret value. I'd prefer to have an other color but I can live with it.

Could you reduce the size of the icon, it's too big compared to the rest. (18 or 20 should be fine)
Could you add title attribute on the icon explaining what it does ?

The value is briefly shown on load of the page. This can be fixed if you wrap the secret part with the jinja.

I think this html is more appropriate:
```html
<div x-data="SecretSection()">
    {% if secret%}
    <p x-show="secret">
        <span class="text-md text-blue-600 hover:text-blue-500 focus:outline-none cursor-pointer"
            @click="showSecret()">
            {% include "ui/icons/eye.html" with w=4 h=4 ml=1 %}
        </span>
    </p>
    <p x-show="!secret" {% if not translate %}translate="no" {% endif %}>{{ text|default:"-" }}
        <button type="button" class="inline align-middle text-blue-600 hover:text-blue-500 focus:outline-none cursor-pointer"
            @click="showSecret()">
            {% include "ui/icons/eye_slash.html" with w=3 h=3 ml=1 %}
        </button>
    </p>

    {% else %}
    <p {% if not translate %}translate="no" {% endif %}>{{ text|default:"-" }}</p>
    {% endif %}
</div>
```
I've removed the flex and used an actual button to wrap the icon.

cheikhgwane:
> If the property has a value, I'd display "***********" instead of nothing when the field is in secret mode.
> 
> I'm not sure it's a good idea to use the primary color on the icon to show/hide the secret value. I'd prefer to have an other color but I can live with it.
> 
> Could you reduce the size of the icon, it's too big compared to the rest. (18 or 20 should be fine) Could you add title attribute on the icon explaining what it does ?
> 
> The value is briefly shown on load of the page. This can be fixed if you wrap the secret part with the jinja.
> 
> I think this html is more appropriate:
> 
> ```
> <div x-data="SecretSection()">
>     {% if secret%}
>     <p x-show="secret">
>         <span class="text-md text-blue-600 hover:text-blue-500 focus:outline-none cursor-pointer"
>             @click="showSecret()">
>             {% include "ui/icons/eye.html" with w=4 h=4 ml=1 %}
>         </span>
>     </p>
>     <p x-show="!secret" {% if not translate %}translate="no" {% endif %}>{{ text|default:"-" }}
>         <button type="button" class="inline align-middle text-blue-600 hover:text-blue-500 focus:outline-none cursor-pointer"
>             @click="showSecret()">
>             {% include "ui/icons/eye_slash.html" with w=3 h=3 ml=1 %}
>         </button>
>     </p>
> 
>     {% else %}
>     <p {% if not translate %}translate="no" {% endif %}>{{ text|default:"-" }}</p>
>     {% endif %}
> </div>
> ```
> 
> I've removed the flex and used an actual button to wrap the icon.

Oh thanks. You're right.  And for the color which do you think is more suitable?

hexa/ui/templates/ui/datacard/property_text.html
cheikhgwane: Okay, that's right, will update it.

hexa/static/js/hexa.js
```diff
@@ -423,3 +423,16 @@ function AdvancedSearch(initialQuery) {
         }
     }
 }
+
+/**
+ * 
+ * Display text property secret
+ */
+function SecretSection() {
```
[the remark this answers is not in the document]
cheikhgwane: Yes, you right

hexa/static/js/hexa.js
cheikhgwane: Yes, will change it.

hexa/ui/templates/ui/datacard/property_text.html
```diff
@@ -1,11 +1,30 @@
...
-            {{ text|default:"-" }}
-        </div>
+{% if markdown %}
+<div class="prose prose-sm" {% if not translate %}translate="no" {% endif %}>
+    {{ text|default:"-" }}
+</div>
+{% else %}
+<div x-data="SecretSection()">
+    {% if secret%}
+    <p x-show="secret" class="flex w-fit">
```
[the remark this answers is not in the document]
cheikhgwane: The clickable zone was taking all the width of the parent. It made this on order to reduce it.

hexa/ui/templates/ui/datacard/property_text.html
cheikhgwane: Okay. I'll have a look at it.

hexa/ui/templates/ui/icons/eye.html
cheikhgwane: Okay.

PR #369, opened 2022-12-09
Decouple connector_accessmod & connector_airflow from Pipeline module

ekatsah:
Decouple connector_accessmod & connector_airflow from Pipeline module. This is the groundwork to refactor the pipeline module without impacting those two downstream modules. 

## Changes

- make the pipeline model concrete
- Analysis, Project and DAG inherit from model and index from core instead of Datasource/Pipeline
- Add missing fields, migrations etc

## How/what to test

- migration is ok
- tests are ok


hexa/plugins/connector_accessmod/models.py
```diff
@@ -239,7 +231,7 @@ def owner(
         return self.project.owner
 
     def index_object(self):
```
pvanliefland, line 233: Maybe `index_object` could be removed altogether?

pvanliefland (approved):
LGTM, just a simple suggestion. 

Let me know when it's on demo, I can help test it.

hexa/plugins/connector_accessmod/models.py
ekatsah: I think it's required for the permission system :/

PR #373, opened 2022-12-16
clean up unused codes

cheikhgwane:
Remove a bunch of unused code/templates.

## Changes

- Add redirection to NEW_FRONTEND_DOMAIN
- Remove unused code.

## How/what to test

## Screenshots / screencast


hexa/user_management/tests/test_views.py
```diff
@@ -65,16 +65,14 @@ def test_tos(self):
         # without validation -> page should ask to accept tos
         response = self.client.get(reverse("core:dashboard"))
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(b"TEST-KEY: ACCEPT_TOS" in response.content, True)
```
qgerome, line 68: I'm not sure for this. I think we will have to implement it on the new frontend as well.

cheikhgwane: I have rollback the change for this. It thought it would break the test but it's not the case.

PR #378, opened 2023-01-02
Add upload pipeline mutation

ekatsah:
Add a new mutation for the pipeline module to upload the code of a pipeline. The code is versionned, but the versionning system is basic. This approch is pretty lean: no external deps, no external volume etc. It's a solid foundation: easy to change, easy to add pin etc.

## Changes

- Add a model for the pipeline content
- Add a mutation to add content to a pipeline

# How to test

Use the latest version of the CLI and upload a pipeline


pvanliefland (changes requested):
Cool ! 

I think we could take some time to discuss it tomorrow. I've already added a few comments.

I'm mostly wondering about two things :

- How to keep a link with git
- How do we determine which Docker image should run the code

ekatsah:
- Git support: let's add a git reference in PipelineVersion when supporting git oriented workflow
- which docker image to use: let's use LATEST of datascience et support more image when the concept is clearer in the workspace?

PR #381, opened 2023-01-04
Add pipeline2 scheduler + runner

ekatsah:
To run pipelines, the backend needs a background worker to start a kubernetes pods with the right parameters: this is the ```runner```. The ```runner``` is a while True loop querying the DB for QUEUED PipelineRun; when it founds one, it forks itself and starts the pipeline; the fork watching the execution of one pipeline, let's call that the ```watcher```. The ```watcher``` can starts the pipeline in a docker for development purpose, or as a kubernetes pod in production. The watcher collects statistics, logs, return code, ... Every few seconds, the watcher also update a heartbeat to prove that 1/ the watcher is still alive, 2/ the pipeline is still alive. It's used to reap crashed ```watcher``` by the ```runner```: every minutes, if a PipelineRun is in a state RUNNING with a last heartbeat bigger that 2 minutes, it will be marked as FAILED with a timeout message in its logs.

This worker can start & run pipeline created manually, but to enable the "scheduled" execution feature, we need another background worker: the ```scheduler```. Also a while True loop, querying all Pipeline objects, seeing if some execution is needed in the near future and create a PipelineRun in state QUEUED if it needs to be. That PipelineRun is then picked by the ```runner``` with the above process.

## Changes

- Rename webhook token -> access token
- Add runner worker
- Add scheduler worker
- Trigger type (scheduled, manual) is now an enum, not a string

## How/what to test

- Start the runner, create a pipeline run (with API or shell), watch it execute.
- Start the scheduler with a runner, create a pipeline with a schedule "*/10 * * * *", watch it run every 10 minutes


hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -0,0 +1,254 @@
...
+        if e.status != 404:
+            logger.exception("pod delete")
+
+    return remote_pod.status.phase == "Succeeded", "STDOUT\n%s\n\nMESSAGES\n%s" % (
+        stdout,
+        termination_message,
+    )
+
+
+def run_pipeline_local(run: PipelineRun, env_var: dict):
```
pvanliefland, line 157: `run_pipeline_docker` ?

hexa/pipelines/models.py
```diff
@@ -265,9 +254,9 @@ def status(self):
 
     @property
     def trigger_mode(self):
```
pvanliefland, line 256: I suggest storing the trigger mode directly in the database

pvanliefland (commented):
Nice work ! Here is a first batch of comments.

In addition to the comments in the code, some feedback :

- It feels strange for `PipelineRun` to have both `pipeline` and `pipeline_version` fields
- It seems that the system requires `Pipeline` and `PipelineRun` to work. Is it normal? Temporary until workspace integration?

Additionally, we need to figure out how to allow developers to run this setup locally. I'll create a separate task for that.

pvanliefland:
Also, as discussed, it seems that it is not yet possible to specify parameters for running the pipeline

PR #383, opened 2023-01-12
Connections

qgerome:
Add the implementation of the connections in workspaces.

Connection available:
* S3
* Postgresql
* GCS
* Custom
* DHIS2


hexa/workspaces/graphql/schema.graphql
```diff
@@ -125,6 +128,113 @@ extend type Mutation {
   inviteWorkspaceMember(input: InviteWorkspaceMemberInput!): InviteWorkspaceMemberResult! @loginRequired
 }
 
+### Connections
+
+enum WorkspaceConnectionType {
```
pvanliefland, line 133: We miss DHIS2 - this is probably the most important one

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+            raise PermissionDenied
+
+        connection = self.create(workspace=workspace, user=principal, **kwargs)
+        if fields is not None:
+            connection.set_fields(principal, fields)
+
+        return connection
+
+
+class WorkspaceConnectionType(models.TextChoices):
```
pvanliefland, line 160: Let's add DHIS2

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    workspace = models.ForeignKey(
+        Workspace, on_delete=models.CASCADE, related_name="connections"
+    )
+    user = models.ForeignKey(
+        "user_management.User",
+        on_delete=models.CASCADE,
+    )
+    name = models.CharField(max_length=255, null=False)
+    _slug = models.CharField(
```
pvanliefland, line 177: I would suggest to always explicitly store the slug instead of having a `@property` - this would prevent code from breaking should the name change at some point.

The slug could be set to the slugified name either in the frontend or in the backend but saved in an explicit `slug` field

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+        for field in fields:
+            if field.get("id") is None:
+                # Unknown field -> Create it
+                WorkspaceConnectionField.objects.create(
+                    connection=self, user=user, **field
+                )
+            elif field["id"] in fields_map:
+                # Update existing field
+                current_field = fields_map[field["id"]]
+                current_field.update_if_has_perm(user, **current_field)
```
pvanliefland, line 212: Not sure we need this level of permissions - maybe the `update_if_has_perm` on the connection itself is enough

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+
+
+class WorkspaceConnectionType(models.TextChoices):
+    S3 = "S3", _("S3 Bucket")
+    GCS = "GCS", _("GCS Bucket")
+    POSTGRESQL = "POSTGRESQL", _("PostgreSQL DB")
+    CUSTOM = "CUSTOM", _("Custom")
+
+
+class WorkspaceConnection(models.Model):
```
pvanliefland, line 167: Maybe `Connection` ? We are already in a workspace module

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+        return self.save()
+
+    def delete_if_has_perm(self, principal: User):
+        if not principal.has_perm("workspaces.delete_workspace_connection", self):
+            raise PermissionDenied
+
+        return self.delete()
+
+
+class WorkspaceConnectionField(models.Model):
```
pvanliefland, line 244: `ConnectionField`

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+    connection = models.ForeignKey(
+        WorkspaceConnection, on_delete=models.CASCADE, related_name="fields"
+    )
+    user = models.ForeignKey(
+        "user_management.User",
+        on_delete=models.CASCADE,
+    )
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    name = models.CharField(max_length=255, blank=False, null=False)
```
pvanliefland, line 255: Let's discuss this, maybe it should be a `code`

pvanliefland (commented):
Hey @qgerome , a few remarks, but I suggest we go over them tomorrow together - no need to change everything right away

hexa/workspaces/models.py
qgerome: On create, take the name and slugify it. On update, do nothing if slug is not passed.

hexa/workspaces/models.py
qgerome: And validates the slug in the backend (return a graphql error)

hexa/workspaces/models.py
```diff
@@ -133,3 +134,141 @@ class WorkspaceMembership(models.Model):
...
+    value = models.TextField(blank=True)
```
pvanliefland, line 256: We should use `EncryptedTextField` (like for other credentials)

pvanliefland (approved):

PR #395, opened 2023-01-19
fix(Workspaces): use UUID scalar for workspace

cheikhgwane:
Fix for https://github.com/orgs/BLSQ/projects/4/views/2?filterQuery=label%3Abug&pane=issue&itemId=18282101.

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast

### DIFF
--- a/hexa/workspaces/schema.py
+++ b/hexa/workspaces/schema.py
@@ -39,7 +39,7 @@ def parse_uuid_value(value):
         UUID(value, version=4)
         return str(value).upper()
     except (ValueError, TypeError):
-        raise ValueError(f'"{value}" is not a valid input')
+        raise ValueError(f'"{value}" is not a valid uuid')
 
 
 @me_permissions_object.field("createWorkspace")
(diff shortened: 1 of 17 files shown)


cheikhgwane:
> Would be nice to change all query fields that use a `id: String!` by this `UUID` type.

For the workspaces schema only ?

qgerome:
No, all of the ids.
Quentin Gérôme
+32 471 490 190

qgerome:
You can also update the mutations to take this UUID instead of the String when it makes sense.

hexa/workspaces/schema.py
```diff
@@ -21,9 +28,19 @@
...
+uuid_scalar = ScalarType("UUID")
+
+
+@uuid_scalar.value_parser
+def parse_uuid_value(value):
+    try:
+        UUID(value, version=4)
+        return str(value).upper()
+    except (ValueError, TypeError):
+        raise ValueError(f'"{value}" is not a valid input')
```
qgerome, line 42:
```suggestion
        raise ValueError(f'"{value}" is not a valid uuid')
```

qgerome (changes requested):

PR #399, opened 2023-01-26
Create a DB and a bucket when creating a workspace

ekatsah:
When an user create a new workspace with the API, it will also try to create a bucket and a DB. Unfortunately I didn't really write tests, the django test framework doesn't handle well the rollback of a DB creation and we don't have a real mock for GCS plugin.

The link between the workspace and the DB/bucket is probably to discuss? We probably won't have a one-to-one mapping based on the name, an FK is probably better. This is due to different constraints on workspace name vs db/bucket:
- bucket need to be globaly unique, without space and other special characters
- dbname need to be globaly unique, without space and other special characters and max 31 chars.

## Changes

- Add create DB in connector_postgresql
- Add create bucket in connector_gcs and connector_s3
- Call two of these functions in the schema.py of workspace module

## How/what to test

1. Configure the PG server hosting the workspace DB in django admin
2. User call the mutation create_workspace
3. Check that a new bucket and a new DB is created


pvanliefland (commented):
Thanks @ekatsah , I think the code in this PR is really useful. As you said,  we need to figure out some parts of the data model first (links between workspaces / DBs / buckets).

I suggest we keep this PR open for now, we'll integrate the code when we're ready on the workspace front.

pvanliefland:
This has been integrated in other PRs

PR #400, opened 2023-02-02
feat(Workspaces): add workspace database support

cheikhgwane:
Add GraphQL API for workspace database.

## Changes

- Create graphql types/queries/mutations for database 
- Add database field to workspace schema

## How/what to test

## Screenshots / screencast


hexa/databases/schema.py
```diff
@@ -0,0 +1,50 @@
...
+
+database_object = ObjectType("Database")
+database_table_object = ObjectType("DatabaseTable")
+database_queries = QueryType()
+
+
+@database_object.field("tables")
+def resolve_database_tables(database, info, **kwargs):
+    tables_summary = get_database_tables_summary(database=database)
+    paged_tables = result_page(
```
qgerome, line 22: Pagination has to be implemented in the `get_database_tables_summary`

hexa/databases/utils.py
```diff
@@ -0,0 +1,86 @@
+from typing import Dict, List, Tuple
+
+import psycopg2
+from psycopg2 import sql
+
+from hexa.plugins.connector_postgresql.models import Database
+
+
+def get_database_tables_summary(database: Database, table=None, limit_per_table=4):
```
qgerome, line 9: It's also better to handle the pagination of the result in this function than in the resolver

qgerome: `get_database_definition` is more appropriate
I think it's better to hide the implementation detail of the database layer and just pass a workspace instead of a database from the connector_postgresql (this will not be used in the future I think)

qgerome: the `def get_table_data(workspace: Workspace, tableName: str, rows: int=4)` has to be a different function

hexa/databases/utils.py
```diff
@@ -0,0 +1,86 @@
...
+                            sql.Identifier(data["table_name"])
+                        ),
+                    )
+                    response = cursor.fetchone()
+                    row_count = response["row_count"]
+                tables_summary.append(
+                    {
+                        "name": name,
+                        "total_count": row_count,
+                        "sample": sample_data,
```
qgerome, line 63: I'd replace `total_count` by `approx_count` or just `count` as it's just an estimation of the number of rows and not the actual count.

if the value is -1, you should replace it by None (meaning it's unknown)

hexa/databases/utils.py
qgerome: It's also better I think to handle the unique table case in the `get_table_definition` (where you would return {"name": "My_table", "columns": [], "row_count": 10}

hexa/databases/utils.py
```diff
@@ -0,0 +1,86 @@
...
+                    """,
+                (table,),
+            )
+
+            response: List[Tuple[str, str]] = cursor.fetchall()
+            columns: List[Dict[str, str]] = [
+                {"name": x[0], "type": x[1]} for x in response
+            ]
+
+    return columns
```
qgerome, line 86: Maybe return a named tuple with {name, columns, count}

hexa/workspaces/schema.py
```diff
@@ -73,6 +74,11 @@ def resolve_workspace_permissions(workspace: Workspace, info):
     return workspace
 
 
+@workspace_object.field("database")
```
qgerome, line 77: This should be in the workspaces module

hexa/workspaces/schema.py
```diff
@@ -73,6 +74,11 @@ def resolve_workspace_permissions(workspace: Workspace, info):
...
+def resolve_workspace_database(workspace: Workspace, info, **kwargs):
+    return Database.objects.get(database="hexa-explore-demo")
```
qgerome, line 79: We should hide the implementation details of this first version and not return the database but the workspace instead.
```suggestion
    return workspace
```

hexa/databases/graphql/schema.graphql
```diff
@@ -0,0 +1,30 @@
...
+type DatabaseTablePage {
+  pageNumber: Int!
+  totalPages: Int!
+  totalItems: Int!
+  items: [DatabaseTable!]!
+}
+
+
+extend type Query{
+  databaseTable(id:String!,databaseId:String!): DatabaseTable @loginRequired
```
qgerome, line 29: This should be included in the `Workspace` type and without the `databaseId` argument (since you would receive a workspace)

hexa/databases/graphql/schema.graphql
```diff
@@ -0,0 +1,30 @@
...
+}
```
qgerome, line 30:
```
type Database {
  tables(page, perPage): DatabaseTablePage!
  table(name: String!): DatabaseTable
}
extend type Workspace {
  database: Database!
}
```

hexa/workspaces/graphql/schema.graphql
```diff
@@ -5,6 +5,7 @@ type Workspace {
   members(page: Int, perPage: Int): WorkspaceMembershipPage!
   permissions: WorkspacePermissions!
   countries: [Country!]!
+  database: Database!
```
qgerome, line 8: Better to extend workspace in the databases schema

hexa/databases/graphql/schema.graphql
```diff
@@ -0,0 +1,30 @@
...
+  sample: [[String]]
+}
+
+
+type DatabaseTable{
+  name: String!
+  totalCount: Int!
+  columns: [TableColumn!]!
+  sample:[[String]]
+  database: Database!
```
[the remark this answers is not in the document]
cheikhgwane: When we try to resolve a databaseTableColumn, we need to have access to the db url in order to open a connection. I didn't have access to the database object in the context so that's why I added it here.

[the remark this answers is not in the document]
qgerome: Yes I remember but I think that if you implement it like I commented at the end of the review and if you return the workspace for the Database object you will be able to have it

qgerome (changes requested):

cheikhgwane:
Okay i'll update it @qgerome

hexa/databases/graphql/schema.graphql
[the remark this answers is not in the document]
cheikhgwane: Okay i'll update it @qgerome

hexa/databases/utils.py
```diff
@@ -0,0 +1,109 @@
...
+    res = []
+    for sample in samples:
+        res.append(
+            [{"name": columns[j], "value": sample[j]} for j in range(len(sample))]
+        )
+    return res
+
+
+def get_database_definition(
+    workspace: Workspace, limit_per_table=4, page=1, per_page=10
```
qgerome, line 20: You should not get the sample data with this function. It takes too much time.

hexa/workspaces/graphql/schema.graphql
```diff
@@ -48,6 +48,7 @@ type WorkspaceMembershipPage {
 extend type Query {
   workspace(id: UUID!): Workspace @loginRequired
   workspaces(page: Int, perPage: Int): WorkspacePage! @loginRequired
+  databaseTable(id:String!): DatabaseTable @loginRequired
```
qgerome, line 51: I think you don't need this.

hexa/workspaces/models.py
```diff
@@ -10,6 +10,7 @@
 
 from hexa.core.models import Base
 from hexa.core.models.base import BaseQuerySet
+from hexa.plugins.connector_postgresql.models import Database
```
qgerome, line 13: You can remove this I think

hexa/workspaces/schema.py
```diff
@@ -73,6 +74,12 @@ def resolve_workspace_permissions(workspace: Workspace, info):
     return workspace
 
 
+@workspace_object.field("database")
+def resolve_workspace_database(workspace: Workspace, info, **kwargs):
+    setattr(workspace, "database", Database.objects.get(database="hexa-explore-demo"))
```
qgerome, line 79: No need for this, it's a bit confusing.

hexa/workspaces/models.py
```diff
@@ -58,6 +59,7 @@ class Workspace(Base):
     description = models.TextField(blank=True)
     members = models.ManyToManyField(User, through="WorkspaceMembership")
     countries = CountryField(multiple=True, blank=True)
+    database = models.ForeignKey(Database, null=True, on_delete=models.SET_NULL)
```
qgerome, line 62: Since we haven't defined the model yet, I think it's better to not have this column. @pvanliefland WDYT?

hexa/workspaces/schema.py
qgerome: Better to do a database = Database.objects.get(database="hexa-explore-demo")) in each util functions

qgerome (changes requested):

hexa/databases/utils.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+
+            response: List[Tuple[str, str]] = cursor.fetchall()
+            if not response:
+                return None
+
+            columns: List[Dict[str, str]] = [
+                {"name": x[0], "type": x[1]} for x in response
+            ]
+            cursor.execute(
```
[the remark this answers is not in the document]
cheikhgwane: Okay! I'll revert it

hexa/databases/utils.py
cheikhgwane: You suggest to create a resolver for the sample field.

hexa/workspaces/models.py
pvanliefland: @qgerome @cheikhgwane I agree - not needed for now

cheikhgwane: Okay !

hexa/databases/utils.py
qgerome: Yup

cheikhgwane:
@qgerome I've added a resolver for the  databaseTable sample field but not that with the current implementation we will have the same issue that we had last time : how to access to the workspace object inside the resolver.

qgerome:
> 

Ok, If it's pushed I'll take a look at it

hexa/databases/utils.py
```diff
@@ -0,0 +1,120 @@
...
+                """
+            )
+            response: List[Tuple[str, str, int]] = cursor.fetchall()
+            tables: Dict[str, Dict] = {
+                x[0]: x for x in response if x[0] not in IGNORE_TABLES
+            }
+
+            for name, data in tables.items():
+                row_count = data["row_count"]
+                if data["row_count"] < 10_000:
```
qgerome, line 47: This part can be removed. The estimate is good enough plus querying a count for each table separately is not really a good idea as it will potentially trigger 1 query per table

hexa/databases/utils.py
```diff
@@ -0,0 +1,120 @@
...
+                    FROM information_schema.tables
+                    JOIN pg_class ON information_schema.tables.table_name = pg_class.relname
+                    WHERE table_schema = 'public'
+                    AND table_name=(%s);
+                """,
+                (table,),
+            )
+            res = cursor.fetchone()
+            row_count = res["row_count"]
+            if row_count < 10_000:
```
qgerome, line 94: This can be removed.

hexa/databases/tests/test_utils.py
```diff
@@ -0,0 +1,137 @@
...
+    @mock.patch("psycopg2.connect")
+    @mock.patch("hexa.databases.utils.Database")
+    def test_get_table_definition_not_found(self, mock_connect, mock_db):
+        mock_db.get.return_value = self.DB1
+        mock_connection = mock_connect.return_value
+        mock_context_object = mock_connection.__enter__.return_value
+        cursor = mock_context_object.cursor.return_value.__enter__.return_value
+        cursor.fetchall.return_value = []
+
+        result = get_database_definition(self.WORKSPACE)
```
qgerome, line 119: The name of the test makes you think you'll test `get_table_definition` but only `get_database_definition` is called.

hexa/databases/tests/test_utils.py
```diff
@@ -0,0 +1,137 @@
...
+        self.assertEqual(0, result["total_items"])
+
+    @mock.patch("psycopg2.connect")
+    def test_get_table_definition(self, mock_connect):
```
qgerome, line 123: It would be interesting to check what happens if you try to get details of a non-existent table.

qgerome (changes requested):

hexa/databases/utils.py
cheikhgwane: Yes, I agree.

cheikhgwane:
Great @qgerome :+1:

hexa/databases/tests/test_utils.py
```diff
@@ -0,0 +1,142 @@
...
+        cursor = mock_context_object.cursor.return_value.__enter__.return_value
+        cursor.fetchall.return_value = [schema]
+        cursor.fetchone.return_value = {"row_count": row_count}
+
+        result = get_table_definition(self.WORKSPACE, table_name)
+
+        self.assertEqual(table, result)
+
+    @mock.patch("psycopg2.connect")
+    def test_get_table_definition_not_found(self, mock_connect):
```
qgerome, line 133: @pvanliefland Is there a better way to mock a context manager ?

hexa/workspaces/schema/types.py
```diff
@@ -72,6 +72,11 @@ def resolve_workspace_countries(workspace: Workspace, info, **kwargs):
     return []
 
 
+@workspace_object.field("database")
```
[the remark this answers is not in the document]
cheikhgwane: To create a new schema.py or to put it the new schema folder?

hexa/databases/tests/test_utils.py
pvanliefland: Not that I know of, sorry

PR #401, opened 2023-02-07
Workspace slug

qgerome:
Add a generated slug on the workspace model.
The slug is the name of the workspace slugified with a random hash as suffix.
Workspace.id has been replaced by Workspace.slug in the graphql queries and mutations.

Link to the frontend PR: https://github.com/BLSQ/openhexa-frontend/pull/194


hexa/workspaces/migrations/0007_alter_workspace_slug.py
```diff
@@ -0,0 +1,41 @@
+# Generated by Django 4.1.3 on 2023-02-06 13:33
+
+import re
+
+import django.core.validators
+from django.db import migrations, models
+
+
+def migrate_workspace_slug(apps, schema_editor):
+    from hexa.workspaces.models import Workspace, create_workspace_slug
```
pvanliefland, line 10: You should use `apps.get_model` to prevent issues with model versions

hexa/workspaces/models.py
```diff
@@ -30,6 +38,7 @@ def create_if_has_perm(
             raise PermissionDenied
 
         create_kwargs = {"name": name, "description": description}
+        create_kwargs["slug"] = create_workspace_slug(name)
```
pvanliefland, line 41: Can we be more strict with slug creation and validation ? Make sure we only have lowercase letters, numbers and dashes?

pvanliefland (commented):
Looks good, just a couple of remarks

hexa/workspaces/models.py
qgerome: From my understanding of slugify, that's exactly what it does. It's makes the text lowercase, replace spaces by dashes ... It is enough no ?

hexa/workspaces/migrations/0007_alter_workspace_slug.py
qgerome: Ok and what about the create_workspace_slug ? Can we leave it like this or is there another way to import it ? @pvanliefland

pvanliefland: For create_workspace_slug it's alright

hexa/workspaces/models.py
pvanliefland: It also accepts underscores, which we don't want (not valid for bucket / db names)

qgerome: Regarding `slugify`, by default it does not.

```python
from slugify import slugify
slugify('a_string_with_underscores_')
'a-string-with-underscores'
```

The `validate_slug` of Django does accept it, I'll make a custom validator.

qgerome: Done

hexa/workspaces/tests/test_models.py
```diff
@@ -36,6 +38,27 @@ def test_create_workspace_regular_user(self):
...
+            )
+        self.assertEqual(workspace.slug, "this-is-a-very-long-wor-mock")
+        self.assertTrue(len(workspace.slug) <= 30)
+
+    def test_create_workspace_with_underscore(self):
+        with patch("secrets.token_hex", lambda _: "mock"):
+
+            workspace = Workspace.objects.create_if_has_perm(
+                self.USER_JULIA,
+                name="worksp?ace_wi😱th_und~ersc!/ore",
```
qgerome, line 57: I've added this to test the slug @pvanliefland

hexa/workspaces/migrations/0008_alter_workspace_slug.py
```diff
@@ -0,0 +1,32 @@
...
+        migrations.AlterField(
+            model_name="workspace",
+            name="slug",
+            field=models.CharField(
+                editable=False,
+                max_length=30,
+                unique=True,
+                validators=[
+                    django.core.validators.RegexValidator(
+                        re.compile("^[-a-zA-Z0-9]+\\Z"),
```
pvanliefland, line 25: We could remove the uppercase letters

qgerome:
Of course -_-‘

PR #410, opened 2023-02-15
feat(Workspaces): Add bucket creation, listing of objects & other actions

qgerome:
- Create a bucket on creation of a workspace.
- Expose the objects of the bucket in the graphql API with several actions (create folder, upload new objects, delete objects)


hexa/files/graphql/schema.graphql
```diff
@@ -0,0 +1,51 @@
...
+  PERMISSION_DENIED
+}
+
+type DeleteWorkspaceObjectResult {
+  success: Boolean!
+  errors: [DeleteWorkspaceObjectError!]!
+}
+
+extend type Mutation {
+  deleteWorkspaceObject(input: DeleteWorkspaceObjectInput!): DeleteWorkspaceObjectResult
```
pvanliefland, line 45: `deleteBucketObject` ?

hexa/files/graphql/schema.graphql
```diff
@@ -0,0 +1,51 @@
...
+
+type BucketObject {
+  name: String!
+  size: Int
+  updatedAt: DateTime
+  type: BucketObjectType!
+}
+
+type BucketObjectPage {
+  nextPageToken: String
```
pvanliefland, line 19: I would argue that we could keep our current pagination approach

pvanliefland (commented):
Nice, work, there are a few points that I'd like to discuss with you

hexa/files/graphql/schema.graphql
qgerome: Agreed

config/settings.py
```diff
@@ -409,3 +409,13 @@ def sentry_tracer_sampler(sampling_context):
 WORKSPACE_DATABASE_PASSWORD = os.environ.get("WORKSPACE_DATABASE_PASSWORD")
 WORKSPACE_DATABASE_HOST = os.environ.get("WORKSPACE_DATABASE_HOST")
 WORKSPACE_DATABASE_PORT = os.environ.get("WORKSPACE_DATABASE_PORT")
+
+# Filesystem configuration
+WORKSPACE_BUCKET_PREFIX = os.environ.get("WORKSPACE_BUCKET_PREFIX", "hexa-test-")
+WORKSPACE_BUCKET_REGION = os.environ.get("WORKSPACE_BUCKET_REGION", "EUROPE-WEST1")
```
pvanliefland, line 415: Are you sure about `EUROPE-WEST1`? In the docs, regions are lowercased (https://cloud.google.com/compute/docs/regions-zones)

config/settings.py
```diff
@@ -409,3 +409,13 @@ def sentry_tracer_sampler(sampling_context):
 WORKSPACE_DATABASE_PASSWORD = os.environ.get("WORKSPACE_DATABASE_PASSWORD")
 WORKSPACE_DATABASE_HOST = os.environ.get("WORKSPACE_DATABASE_HOST")
 WORKSPACE_DATABASE_PORT = os.environ.get("WORKSPACE_DATABASE_PORT")
+
+# Filesystem configuration
+WORKSPACE_BUCKET_PREFIX = os.environ.get("WORKSPACE_BUCKET_PREFIX", "hexa-test-")
```
pvanliefland, line 414: I would put the default value in the Docker compose file

config/settings.py
pvanliefland: And I wouldn't put a default value here - rather in the Docker compose file

hexa/files/api.py
```diff
@@ -0,0 +1,183 @@
...
+
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from google.api_core.exceptions import NotFound
+from google.cloud import storage
+from google.cloud.exceptions import Conflict
+from google.cloud.storage.blob import Blob
+from google.oauth2 import service_account
+
+WORKSPACE_BUCKET_PREFIX = settings.WORKSPACE_BUCKET_PREFIX or ""
```
pvanliefland, line 12: The default `""` could be in `settings.py`, and the dev prefix in the docker compose file

hexa/workspaces/models.py
```diff
@@ -85,7 +91,7 @@ class Workspace(Base):
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     name = models.TextField()
     slug = models.CharField(
-        max_length=30,
+        max_length=63,
```
pvanliefland, line 94: I think we might need to shorten this - let's discuss this

pvanliefland (changes requested):
@qgerome I think it's almost perfect, I still need to discuss the slug with you as I'm afraid we might run into issues with Kubernetes, let's talk about it tomorrow.

pvanliefland (approved):
LGTM!

PR #411, opened 2023-02-16
Notebooks in workspaces

pvanliefland:
This PR integrates the notebooks feature within workspaces. It allows users to launch one jupyter server by workspace. These workspace-specific notebooks servers will be launched with the credentials / env variables of the workspace only.

## Changes

- Added a GraphQL mutation that launches jupyter servers using the jupyterhub REST API (and named servers)
- Added a standard HTTP endpoint the the Jupyterhub spawner can call to get the workspace credentials

## How/what to test

This can be tested locally - you need to run `openhexa-notebooks` using Docker.

- Run the `launchNotebookServer` mutation
- Navigate to the URL returned by the mutation
- Make sure that the notebook server runs fine and has workspace-specific credentials

## Screenshots / screencast


hexa/notebooks/schema.py
```diff
@@ -1,18 +1,80 @@
...
+    try:
+        workspace = Workspace.objects.filter_for_user(request.user).get(
+            slug=workspace_slug
+        )
+    except Workspace.DoesNotExist:
+        return {"success": False, "server": None, "errors": ["NOT_FOUND"]}
+
+    if not request.user.has_perm(
+        "workspaces.launch_notebooks", workspace
+    ):  # TODO: check workspace
```
qgerome, line 40: What do you mean by this TODO ?

hexa/notebooks/schema.py
```diff
@@ -1,18 +1,80 @@
...
+    # Get user, create if needed
+    user_response = requests.get(
+        f"{settings.NOTEBOOKS_API_URL}/users/{request.user.email}", headers=headers
+    )
+    if user_response.status_code == 404:
+        user_response = requests.post(
+            f"{settings.NOTEBOOKS_API_URL}/users/{request.user.email}", headers=headers
+        )
+        if user_response.status_code != 201:
+            return {"success": False, "server": None, "errors": ["UNKNOWN_ERROR"]}
```
qgerome, line 56: FYI `None` values do not need to be in the return dict

hexa/notebooks/views.py
```diff
@@ -1,15 +1,66 @@
...
 from django.views.decorators.http import require_POST
 
 from hexa.app import get_hexa_app_configs
 from hexa.notebooks.credentials import NotebooksCredentials
 
 
 @require_POST
-@csrf_exempt  # TODO: we should remove this
+def authenticate(request: HttpRequest) -> HttpResponse:
+    """This API endpoint is called by the notebooks component to get authenticate the current user using Django
```
qgerome, line 10:
```suggestion
    """This API endpoint is called by the notebooks component to authenticate the current user using Django
```

qgerome:
The openhexa-notebooks is still building... I'll test it when it's ready.

hexa/notebooks/credentials.py
```diff
@@ -7,12 +7,12 @@ class NotebooksCredentials(HexaCredentials):
...
 
     def __init__(self, user):
-        self.user = user
+        self.user = user  # TODO: enforce user
         self.env: dict[str, str] = {}
         self.files: dict[str, bytes] = {}
 
     @property
-    def authenticated(self):
+    def authenticated(self):  # TODO: remove
```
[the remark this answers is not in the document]
pvanliefland: Because there is still a view (`credentials`) that uses it. Once I remove the view, the property can be removed as well

hexa/notebooks/graphql/schema.graphql
```diff
@@ -1,3 +1,23 @@
+type NotebookServer {
+    name: String!
+    url: String!
+}
+input LaunchNotebookServerInput {
+    workspaceSlug: String!
+}
+enum LaunchNotebookServerError {
+    UNKNOWN_ERROR
```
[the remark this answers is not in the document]
pvanliefland: I'm not sure I understand - all of our mutations return an error code as far as I know. Do we want to change that behaviour?

hexa/notebooks/schema.py
pvanliefland: It's a leftover - at one point I didn't check whether the user had any link to the workspace. I removed the comment.

hexa/notebooks/schema.py
pvanliefland: Good catch. I removed `None` values - and added the empty `errors` list in case of success, I had forgotten to do it.

hexa/notebooks/views.py
pvanliefland: Thanks, fixed

hexa/notebooks/graphql/schema.graphql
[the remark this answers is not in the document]
qgerome: The logic is more that you check the status_code of the response and if it's not 200 you return this UNKNOWN_ERROR without even logging it in the backend. We lose track of the existence of the error (except the user that will get something). I'd strongly prefer that we raise an exception in the backend if the requests fail and receives in the frontend a failed mutation. That's just my point of view and we can leave it like this.

cheikhgwane:
@pvanliefland I've added some tests but due to some authentication rules (login_middleware that redirect to login page when user request a url that's not on the ANONYMOUS_URL list ) I was not able to test some cases (eg: user not logged and trying to access /notebooks/default-credentials)

pvanliefland:
> @pvanliefland I've added some tests but due to some authentication rules (login_middleware that redirect to login page when user request a url that's not on the ANONYMOUS_URL list ) I was not able to test some cases (eg: user not logged and trying to access /notebooks/default-credentials)

@cheikhgwane you can actually test the redirect (using `self. assertRedirects`).

Can you also test the mutation?

cheikhgwane:
@pvanliefland should be good!

PR #412, opened 2023-02-17
feat(Workspaces) : create database workspace

cheikhgwane:
Create a database, role when creating a workspace. The created database and role use the workspace slug name as identifier.
The created role has CONNECT and CREATE privileges granted on the created database.
## Changes

- Add create_database function in api.py on database module
- Add create_database api on workspace model
- Replace and use the workspace slug name when querying for workspace database objects

## How/what to test
 1 - Configure env variables that hold the pg_server credentials
 2 - Call the create workspace mutation
 3 - Check that a database and a role with the workspace slug name are created


pvanliefland (changes requested):
Nice work 👍  I will try to test it asap.

Could you already try to "disconnect" workspace databases from the Postgres connector? It is not necessary to use the `Database` model anymore.

cheikhgwane:
> Nice work +1 I will try to test it asap.
> 
> Could you already try to "disconnect" workspace databases from the Postgres connector? It is not necessary to use the `Database` model anymore.

Okay, on it.

hexa/databases/api.py
```diff
@@ -0,0 +1,51 @@
...
+    role = credentials["role"]
+    password = credentials["password"]
+    host = credentials["host"]
+    port = credentials["port"]
+
+    url = f"postgresql://{role}:{password}@{host}:{port}"
+    try:
+        conn = psycopg2.connect(url)
+        conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+        with conn.cursor() as cursor:
```
pvanliefland, line 30: I would be extra careful about SQL injection here. Normally we would use execute parameters but this is not possible for CREATE DATABASE operations.

Could you use the `psycopg2.sql` submodule? You can have a look here: https://www.psycopg.org/docs/sql.html ?

cheikhgwane: Oh yes, I'll update it

cheikhgwane:
@pvanliefland If we remove the Database model, do we need to generate a password with the created role ? For now the created role cannot connect without a password.

pvanliefland:
> @pvanliefland If we remove the Database model, do we need to generate a password with the created role ? For now the created role cannot connect without a password.

Yes, you need to generate a password and store it on the workspace model using an `EncryptedTextField` field.

hexa/databases/api.py
```diff
@@ -0,0 +1,71 @@
...
+                sql.SQL("GRANT USAGE ON SCHEMA public TO {role_name};").format(
+                    role_name=sql.Identifier(db_name)
+                )
+            )
+            cursor.execute(
+                sql.SQL(
+                    "GRANT SELECT ON ALL TABLES IN SCHEMA public TO {role_name};"
+                ).format(role_name=sql.Identifier(db_name))
+            )
+    finally:
```
qgerome, line 69: Why not using a context manager for the connection as well ?

hexa/databases/api.py
```diff
@@ -0,0 +1,71 @@
...
+    credentials = get_db_server_credentials()
+
+    role = credentials["role"]
+    password = credentials["password"]
+    host = credentials["host"]
+    port = credentials["port"]
+
+    url = f"postgresql://{role}:{password}@{host}:{port}"
+
+    db_name = ""
```
qgerome, line 32: It's weird that you changes the name of the database you receive. How do the caller know that you maybe adapted the name of the database ? IMHO it should be a helper and it should only be validated here (and raises an error if it's not correct)

hexa/databases/utils.py
```diff
@@ -1,28 +1,33 @@
...
+    port = credentials["port"]
+
+    return f"postgresql://{workspace.slug}:{password}@{host}:{port}/{workspace.slug}"
 
-def get_database_definition(workspace: Workspace):
-    database = get_workspace_database(workspace)
 
-    with psycopg2.connect(database.url) as conn:
+def get_database_definition(workspace: Workspace):
+    url = get_workspace_database_url(workspace)
```
qgerome, line 19: I'd use the default database credentials to get db definition

qgerome (changes requested):

hexa/databases/api.py
cheikhgwane: You right

hexa/databases/api.py
cheikhgwane: Yes, you right,.

hexa/databases/utils.py
cheikhgwane: WDYT about it @pvanliefland ?

pvanliefland: I agree with @qgerome

hexa/databases/api.py
pvanliefland: Also, now that we use `sql.Identifier`, no need to be super paranoid about the dbname - it should prevent SQL injection. Using the workspace slug should be fine.

hexa/databases/utils.py
cheikhgwane: Okay works for me

hexa/databases/api.py
cheikhgwane: The create database command cannot be used inside a transaction block (see https://github.com/psycopg/psycopg2/issues/1305#issuecomment-866712961)

cheikhgwane: There were some issues about the CREATE DATABASE command and the with statement (cannot use it inside a transaction) and also another one related to closing connection (see https://www.psycopg.org/docs/usage.html#with-statement)

hexa/databases/api.py
```diff
@@ -0,0 +1,71 @@
...
+        "role": settings.WORKSPACE_DATABASE_ROLE,
+        "password": settings.WORKSPACE_DATABASE_PASSWORD,
+        "host": settings.WORKSPACE_DATABASE_HOST,
+        "port": settings.WORKSPACE_DATABASE_PORT,
+    }
+
+
+def check_db_name(name):
+    if not name:
+        raise ValueError("rmpty value for database name")
```
pvanliefland, line 18: Can you add typing info to your functions? Not only this one, in general

hexa/databases/utils.py
pvanliefland: I don't think you implemented what @qgerome was suggesting - should't you use the default credentials?

pvanliefland (commented):
Almost there 👍  - see comments.

Also, I don't see the CREATE EXTENSION statements, am I missing something?

hexa/databases/api.py
cheikhgwane: > Almost there +1 - see comments.
> 
> Also, I don't see the CREATE EXTENSION statements, am I missing something?

Oh not yet. I'll add it

hexa/databases/api.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    credentials = get_db_server_credentials()
+
+    role = credentials["role"]
+    password = credentials["password"]
+    host = credentials["host"]
+    port = credentials["port"]
+
+    url = f"postgresql://{role}:{password}@{host}:{port}"
+    check_db_name(name=db_name)
```
qgerome, line 45:
```python
# Start of file
import re
# ...

if (re.match(r"^[_a-z][a-z0-9_]{,30}$", db_name) is None:
  raise ValidationError("Name must contain only lowercase alphanumeric characters, start with a letter or a underscore and with a maximum length of 31 characters")

Comment (qgerome) [reply hexa/databases/api.py:45]: As per https://www.postgresql.org/docs/7.0/syntax525.htm

File: hexa/databases/utils.py (hunk 4 above)
Comment (qgerome) [reply hexa/databases/utils.py:20]: @cheikhgwane We can't give the permission to each role to query the default schema. Queries on `information_schema` have to be run with the user that creates the databases

File: hexa/workspaces/models.py [hunk 7]
```diff
@@ -55,6 +56,7 @@ def create_if_has_perm(
         if description is None:
             create_kwargs["description"] = "This is a workspace for {}".format(name)
 
+        create_database(create_kwargs["slug"])
```
qgerome, line 59: The slug as-is is not valid as a database name. You need to prepare the slug to match the required format.

I'd also store on the workspace model the db_name & db_password

hexa/databases/api.py
```diff
@@ -0,0 +1,92 @@
...
+    return {
+        "role": settings.WORKSPACE_DATABASE_ROLE,
+        "password": settings.WORKSPACE_DATABASE_PASSWORD,
+        "host": settings.WORKSPACE_DATABASE_HOST,
+        "port": settings.WORKSPACE_DATABASE_PORT,
+    }
+
+
+def format_db_name(requestedName: str):
+    dbName = ""
```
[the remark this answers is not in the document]
cheikhgwane: All right

hexa/databases/api.py
```diff
@@ -0,0 +1,92 @@
...
+        "role": settings.WORKSPACE_DATABASE_ROLE,
+        "password": settings.WORKSPACE_DATABASE_PASSWORD,
+        "host": settings.WORKSPACE_DATABASE_HOST,
+        "port": settings.WORKSPACE_DATABASE_PORT,
+    }
+
+
+def format_db_name(requested_name: str):
+    db_name = ""
+    for c in requested_name.lower().replace(" ", "_")[:31]:
```
qgerome, line 21: You have to check that the db starts with a letter or an underscore and the resulting db_name might be shorter than 31 characters if it includes unsupported chars. You need to limit the length after the processing.

hexa/databases/api.py
```diff
@@ -0,0 +1,92 @@
...
+                )
+            )
+
+    finally:
+        if conn:
+            conn.close()
+
+    # load extensions into the new db
+    try:
+
```
qgerome, line 83:
```suggestion
```

hexa/databases/utils.py
```diff
@@ -1,28 +1,34 @@
...
+
+    return f"postgresql://{role}:{password}@{host}:{port}/{database}"
 
-def get_database_definition(workspace: Workspace):
-    database = get_workspace_database(workspace)
 
-    with psycopg2.connect(database.url) as conn:
+def get_database_definition(workspace: Workspace):
+    url = get_database_url(workspace.db_name)
+    with psycopg2.connect(url) as conn:
```
qgerome, line 21: Isn't it a problem that we do not close the connection here ? (if we do it above with the creation ?) cc: @pvanliefland

hexa/databases/api.py
```diff
@@ -0,0 +1,92 @@
...
+        conn = psycopg2.connect(url)
+        conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+        with conn.cursor() as cursor:
+            cursor.execute(
+                sql.SQL("CREATE DATABASE {db_name};").format(
+                    db_name=sql.Identifier(db_name),
+                )
+            )
+            cursor.execute(
+                sql.SQL("CREATE ROLE {role_name} LOGIN;").format(
```
qgerome, line 64: You need to set a password for it no ?

hexa/workspaces/models.py
```diff
@@ -55,6 +60,11 @@ def create_if_has_perm(
         if description is None:
             create_kwargs["description"] = "This is a workspace for {}".format(name)
 
+        db_name = format_db_name(create_kwargs["slug"])
+        create_kwargs["db_name"] = db_name
+        create_kwargs["db_password"] = get_db_server_credentials()["password"]
```
qgerome, line 65: What I understand here is that you set the password to be the one used by the user that creates the database. 

Each workspace has to have a unique db and a unique password.

qgerome (changes requested):

hexa/databases/api.py
cheikhgwane: The check is done on a separate function (see  line 27)

hexa/databases/utils.py
cheikhgwane: Hum yes. Will need to add try ... finally block

hexa/databases/api.py
```diff
@@ -0,0 +1,96 @@
...
+def get_db_server_credentials():
+    return {
+        "role": settings.WORKSPACE_DATABASE_ROLE,
+        "password": settings.WORKSPACE_DATABASE_PASSWORD,
+        "host": settings.WORKSPACE_DATABASE_HOST,
+        "port": settings.WORKSPACE_DATABASE_PORT,
+    }
+
+
+def format_db_name(name: str):
```
qgerome, line 19: It would be great to have a few tests for this function

hexa/workspaces/tests/test_models.py
```diff
@@ -62,20 +66,22 @@ def test_create_workspace_with_underscore(self):
...
     def test_create_workspace_admin_user(self):
-        workspace = Workspace.objects.create_if_has_perm(
-            self.USER_JULIA,
-            name="Senegal Workspace",
-            description="This is test for creating workspace",
-        )
+        with patch("hexa.workspaces.models.create_database"):
+            workspace = Workspace.objects.create_if_has_perm(
+                self.USER_JULIA,
+                name="Senegal Workspace",
```
qgerome, line 72: Would be great to test it with different names (with other characters)

qgerome (approved):
We are getting there !

hexa/databases/tests/test_api.py
```diff
@@ -0,0 +1,26 @@
...
+class DatabaseAPITest(TestCase):
+    def test_format_db_name(self):
+        db_name = "RDC_POLIO_PROJECT"
+        self.assertEqual(len(db_name), len(format_db_name(db_name)))
+
+        # test with a name with length superior to 31
+        db_name = "THIS_IS_A_VERY_LONG_TEXT_WITH_MORE_THAN_31_LETTERS"
+        self.assertTrue(len(format_db_name(db_name)) <= 31)
+
+        # test with a name beginning with a character
```
qgerome, line 18: number

hexa/databases/tests/test_api.py
```diff
@@ -0,0 +1,26 @@
...
+        db_name = "1rwandaProject"
+        self.assertFalse(re.match("^[0-9]", format_db_name(db_name)))
```
qgerome, line 20: I think it's better to clearly write what you expect as a result
```
self.assertEqual(format_db_name(db_name), "_1rwandaProject")
```

hexa/databases/api.py
```diff
@@ -24,7 +24,7 @@ def format_db_name(name: str):
 
     # Ensure db_name do not start with a number
     if db_name[0] in "0123456789":
-        db_name.insert(0, "_")
+        db_name = db_name.replace(db_name[0], "_", 1)
```
qgerome, line 27: Doing it like this is over complicated and a bit strange. We do not want to replace a character, we want to add one to be sure to have the right format for PGSQL.
```python
db_name = "_" + db_name
```
It's easier to understand.

hexa/databases/tests/test_api.py
```diff
@@ -0,0 +1,26 @@
...
+from django.core.exceptions import ValidationError
+
+from hexa.core.test import TestCase
+from hexa.databases.api import create_database, format_db_name
+
+
+class DatabaseAPITest(TestCase):
+    def test_format_db_name(self):
+        db_name = "RDC_POLIO_PROJECT"
+        self.assertEqual(len(db_name), len(format_db_name(db_name)))
```
qgerome, line 12: Better to compare the result of format_db_name than the length of the result. You might end up with the same number of characters but completely different from what you'd like to have.

qgerome (changes requested):

hexa/databases/api.py
```diff
@@ -0,0 +1,96 @@
...
+    credentials = get_db_server_credentials()
+
+    role = credentials["role"]
+    password = credentials["password"]
+    host = credentials["host"]
+    port = credentials["port"]
+
+    url = f"postgresql://{role}:{password}@{host}:{port}"
+    validate_db_name(db_name)
+
```
qgerome, line 58:
```suggestion
    conn = None
```

If you have a problem to connect to the database `conn` is not instantiated and cannot be used in the finally clause

qgerome (changes requested):

cheikhgwane:
@qgerome should be good now

hexa/databases/api.py
```diff
@@ -0,0 +1,97 @@
...
+        "role": settings.WORKSPACE_DATABASE_ROLE,
+        "password": settings.WORKSPACE_DATABASE_PASSWORD,
+        "host": settings.WORKSPACE_DATABASE_HOST,
+        "port": settings.WORKSPACE_DATABASE_PORT,
+    }
+
+
+def format_db_name(name: str):
+    db_name = ""
+    for char in name.lower().replace(" ", "_"):
```
qgerome, line 21: Since you take the slug of the workspace, there is already no space in it. It makes more sense to replace `-` by `_` to keep the words

qgerome (changes requested):

hexa/databases/api.py
cheikhgwane: So you advice to not replace or change the replacement stirng?

PR #418, opened 2023-03-02
fix(Workspaces): fix psycopg2 connection error

cheikhgwane:
A connection error may occur using credentials that contain special characters like "@,%". 

## Changes

- Use keyword arguments for opening connection instead of URI string.
- Add function to return the database connection object.

## How/what to test

Generate new credentials with special characters and check if the connection is established without error(s).

### DIFF
--- a/hexa/databases/tests/test_api.py
+++ b/hexa/databases/tests/test_api.py
@@ -1,6 +1,6 @@
 import psycopg2
 from django.core.exceptions import ValidationError
-from psycopg2 import sql
+from psycopg2 import OperationalError, sql
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT, STATUS_READY
 
 from hexa.core.test import TestCase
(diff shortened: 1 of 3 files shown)


pvanliefland (changes requested):
@cheikhgwane good work, the approach looks fine to me.

Could you add a test for the `get_database_connection` function ? With special characters?

hexa/databases/tests/test_api.py
```diff
@@ -79,20 +77,3 @@ def test_create_database_raise_error(self):
         password = "password"
         with self.assertRaises(ValidationError):
             create_database(bad_input, password)
-
-    def test_create_database_not_access(self):
```
pvanliefland, line 83: Why did this test get deleted?

cheikhgwane: Oh I rename it

cheikhgwane: I'll revert it. I think it was during a rebase with main

cheikhgwane:
@pvanliefland it should be okay now.

PR #420, opened 2023-03-06
feat(Workspaces): inject database credentials into workspace notebooks

cheikhgwane:
Inject workspace database credentials (name, host, port...) into workspace notebooks

### DIFF
--- a/hexa/workspaces/tests/test_views.py
+++ b/hexa/workspaces/tests/test_views.py
@@ -1,6 +1,7 @@
 from django.urls import reverse
 
 from hexa.core.test import TestCase
+from hexa.databases.api import get_db_server_credentials
 from hexa.user_management.models import Feature, FeatureFlag, User
 from hexa.workspaces.models import (
     Connection,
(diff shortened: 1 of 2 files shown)


hexa/workspaces/tests/test_views.py
```diff
@@ -90,6 +91,21 @@ def test_workspace_credentials_200(self):
             )
         )
 
+        db_credentials = get_db_server_credentials()
+        db_credentials["host"]
```
pvanliefland, line 95: This has no effect, like the next line

hexa/workspaces/views.py
```diff
@@ -35,6 +36,18 @@ def credentials(request: HttpRequest, workspace_slug: str) -> HttpResponse:
 
     # TODO: Database / Filesystem credentials
 
+    # Database credentials
+    db_credentials = get_db_server_credentials()
+    env.update(
```
qgerome, line 41: I'd only add the `WORKSPACE_DB_URL` at first. @pvanliefland  WDYT ?

pvanliefland (changes requested):
Almost ok

hexa/workspaces/views.py
pvanliefland: The individual variables are needed for R - the typical postgres lib that our team uses cannot use a connection string url and need the individual parameters

PR #423, opened 2023-03-08
fix(Files): Use a base64 encoded string for the GCS credentials

qgerome:
Changing the set of env variables we had to connect to GCS by a b64 encoded string.


pvanliefland (changes requested):
Nice - would you mind using `GCS_SERVICE_ACCOUNT_KEY` as env variable / settings ?That's how those JSON files are called in GCP.

PR #424, opened 2023-03-08
fix(Workspaces): add default workspace default database settings

cheikhgwane:
Fix bug that occurs when connecting to postgres and  WORKSPACE_DATABASE_ROLE default database is not set  or doesn't exist.


docker-compose.yaml
```diff
@@ -35,6 +35,7 @@ x-app: &common
       - WORKSPACES_DATABASE_PASSWORD=hexa-app
       - WORKSPACES_DATABASE_HOST=db
       - WORKSPACES_DATABASE_PORT=5432
+      - WORKSPACES_DATABASE_DEFAULT_DB
```
pvanliefland, line 38: Could you add `hexa-app` as default here?

.env.dist
```diff
@@ -6,3 +6,8 @@ EMAIL_HOST_USER=
 EMAIL_USE_TLS=
 EMAIL_HOST_PASSWORD=
 GCS_SERVICE_ACCOUNT_KEY=
+WORKSPACES_DATABASE_ROLE=
```
pvanliefland, line 9: Is this needed? All the values can be in the `docker-compose.yaml` file

pvanliefland (changes requested):
2 small comments about the env variables for dev

.env.dist
cheikhgwane: Not really needed.

pvanliefland (approved):

PR #427, opened 2023-03-10
chore(Workspaces): add unique constraint for member workspace and user

cheikhgwane:
Add  unique constraint on user and workspace in the WorkspaceMembership model


pvanliefland (approved):
LGTM

PR #429, opened 2023-03-10
Fix workspaces files permissions

qgerome:
Only admins & editors can delete/create objects; all members can download


pvanliefland (approved):
LGTM, the `files/api.py` should probably be in another PR

PR #441, opened 2023-03-24
feat(Workspaces): add archive workspace mutation

cheikhgwane:
Add a mutation for archiving workspace

## Changes

- Add archived field on workspace model on query set manager


hexa/workspaces/models.py
```diff
@@ -140,8 +143,15 @@ def update_if_has_perm(self, *, principal: User, **kwargs):
     def delete_if_has_perm(self, *, principal: User):
         if not principal.has_perm("workspaces.delete_workspace", self):
             raise PermissionDenied
+
         delete_database(self.db_name)
-        self.delete()
```
pvanliefland, line 144: Why did you remove `self.delete()` here? My understanding is that we won't use `delete_if_has_perm` for now. But if you leave the method, then it makes sens that it still calls `self.delete()`

pvanliefland (changes requested):
Good work, almost there!

hexa/workspaces/models.py
cheikhgwane: It shouldn't be removed. I'll put it back

pvanliefland (approved):

qgerome:
Personally I would filter out the archived workspaces by default in the query set returned by the workspacemanager and add a method on the queryset to display the archived one to not have to change all the code to ensure we have a active workspace.

Example: Do we allow users to run a pipeline from an archived workspace ?

This needs to be updated as well: https://github.com/BLSQ/openhexa-app/blob/56e992eb1a1cd3b02323ba3dec556372425cae0f/hexa/pipelines/management/commands/pipelines_scheduler.py#L21

@pvanliefland @cheikhgwane  WDYT ?

pvanliefland:
> Personally I would filter out the archived workspaces by default in the query set returned by the workspacemanager and add a method on the queryset to display the archived one to not have to change all the code to ensure we have a active workspace.
> 
> Example: Do we allow users to run a pipeline from an archived workspace ?
> 
> This needs to be updated as well:
> 
> https://github.com/BLSQ/openhexa-app/blob/56e992eb1a1cd3b02323ba3dec556372425cae0f/hexa/pipelines/management/commands/pipelines_scheduler.py#L21
> 
> @pvanliefland @cheikhgwane WDYT ?

@qgerome that is definitely an option - in that case we need to adapt the admin class so that it uses another manager.

qgerome:
> > Personally I would filter out the archived workspaces by default in the query set returned by the workspacemanager and add a method on the queryset to display the archived one to not have to change all the code to ensure we have a active workspace.
> > Example: Do we allow users to run a pipeline from an archived workspace ?
> > This needs to be updated as well:
> > https://github.com/BLSQ/openhexa-app/blob/56e992eb1a1cd3b02323ba3dec556372425cae0f/hexa/pipelines/management/commands/pipelines_scheduler.py#L21
> > 
> > @pvanliefland @cheikhgwane WDYT ?
> 
> @qgerome that is definitely an option - in that case we need to adapt the admin class so that it uses another manager.

Yes, let's change the manager used for the admin view of the workspaces and make the archived=False the default behaviour. 

Users should not be able to see or run pipelines for archived workspaces.

cheikhgwane:
@qgerome for now what I've done is to filter the archived workspaces on filter_for_user and keep the default behavior for admin queryset (return all workspaces archived or not)

cheikhgwane:
@qgerome for now what I've done is to filter the archived workspaces on filter_for_user and keep the default behavior for admin queryset (return all workspaces archived or not)

qgerome (approved):

PR #444, opened 2023-03-27
feat(Databases): use hash for workspace database name

cheikhgwane:
Relative to [issue](https://github.com/BLSQ/openhexa-team/issues/271). Replace database name by a generated hash.

## Changes

- Use hash for database name
- Update test


hexa/workspaces/models.py
```diff
@@ -68,7 +68,9 @@ def create_if_has_perm(
             )
 
         db_password = User.objects.make_random_password(length=16)
-        db_name = format_db_name(create_kwargs["slug"])
+        db_name = format_db_name(
```
pvanliefland, line 71: @cheikhgwane this will work but after giving it some thoughts, I'd rather use a random uuid4 (without dashes) as the database name and role name.

pvanliefland (changes requested):

hexa/workspaces/models.py
cheikhgwane: Okay, all right.

hexa/workspaces/tests/test_models.py
```diff
@@ -86,7 +86,6 @@ def test_create_workspace_with_random_characters(self):
                 description="Description",
             )
         self.assertEqual(workspace.slug, "1workspace-with-random-mock")
-        self.assertEqual(workspace.db_name, "_1workspace_with_random_mock")
```
pvanliefland, line 89: Can you test the length of the string ? Should be 32 without underscores

hexa/workspaces/models.py
```diff
@@ -68,7 +68,7 @@ def create_if_has_perm(
             )
 
         db_password = User.objects.make_random_password(length=16)
-        db_name = format_db_name(create_kwargs["slug"])
+        db_name = format_db_name(str(uuid.uuid4()))
```
pvanliefland, line 71: We don't need `format_db_name` anymore - let's use `str(uuid.uuid4()).replace("-", "")`

pvanliefland (changes requested):

hexa/workspaces/models.py
cheikhgwane: Okay

cheikhgwane:
I've changed the MAX_LENGTH of database name  from 31 to 63 but still need to check that the generated name doesn't start with a number. @pvanliefland

cheikhgwane:
LGTM! @pvanliefland

pvanliefland (commented):
As discussed, I've switched to a random string

PR #454, opened 2023-04-07
fix(Connections):add database URI connection string

cheikhgwane:
Add database URI connection string


hexa/workspaces/views.py
```diff
@@ -35,6 +35,13 @@ def credentials(request: HttpRequest, workspace_slug: str) -> HttpResponse:
     env = {}
     for connection in connections:
         env.update(connection.env_variables)
+        if connection.connection_type == ConnectionType.POSTGRESQL:
```
qgerome, line 38: It's weird to have the special case for this connection type here. It should be in the property "env_variables" of the connection.

qgerome (changes requested):

qgerome (approved):

PR #455, opened 2023-04-10
feat(Pipelines): Add a code to Pipeline and resolver for the CLI

qgerome:

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast


config/settings.py
```diff
@@ -92,6 +92,7 @@
     "hexa.user_management.middlewares.TwoFactorMiddleware",
     "hexa.plugins.connector_airflow.middlewares.dag_run_authentication_middleware",
     "hexa.pipelines.middlewares.pipeline_run_authentication_middleware",
+    "hexa.workspaces.middlewares.workspace_authentication_middleware",
```
pvanliefland, line 95: Maybe we could be more explicit and call it `workspace_token_authentication_middleware`

hexa/workspaces/graphql/schema.graphql
```diff
@@ -176,6 +191,7 @@ extend type Mutation {
   inviteWorkspaceMember(input: InviteWorkspaceMemberInput!): InviteWorkspaceMemberResult! @loginRequired
   updateWorkspaceMember(input: UpdateWorkspaceMemberInput!): UpdateWorkspaceMemberResult! @loginRequired
   deleteWorkspaceMember(input: DeleteWorkspaceMemberInput!): DeleteWorkspaceMemberResult! @loginRequired
+  getWorkspaceToken(input: GetWorkspaceTokenInput!): GetWorkspaceTokenResult! @loginRequired
```
pvanliefland, line 194: Could we call this `generateWorkspaceToken`?

hexa/workspaces/middlewares.py
```diff
@@ -0,0 +1,38 @@
...
+
+from django.core.signing import BadSignature, Signer
+from django.http import HttpRequest
+
+from hexa.workspaces.models import WorkspaceMembership
+
+logger = getLogger(__name__)
+
+
+# FIXME: This needs to be changed as it allows the user to query everything with just a simple workspace token.
```
pvanliefland, line 12: Is this still the case?

hexa/workspaces/middlewares.py
```diff
@@ -0,0 +1,38 @@
...
+def workspace_authentication_middleware(get_response):
+    """This middleware allows an user to be authenticated through a simple token linked to a workspace"""
+
+    def middleware(request: HttpRequest):
+        try:
+            auth_type, token = request.headers["Authorization"].split(" ")
+            if auth_type.lower() == "bearer":
+                token = Signer().unsign_object(token)
+                membership = WorkspaceMembership.objects.get(access_token=token)
+                request.user = membership.user
```
pvanliefland, line 23: I think we should create a dedicated user class for workspace tokens. This user class would be linked to the user and workspace, and would respect the Django user interface.

This would ease the differentiation between logged-in users and CLI users when it matters.

It would require a bit of work in our workspace permission checks but I think it would be more robust than setting `workspace` and `bypass_two_factor` on a regular user.

hexa/workspaces/migrations/0027_remove_workspacemembership_token_and_more.py
pvanliefland, line 1: I think these three migrations can be merged in a single one - not important but nicer

hexa/workspaces/schema/mutations.py
```diff
@@ -265,3 +272,25 @@ def resolve_delete_workspace_connection(_, info, **kwargs):
 bindables = [
     workspace_mutations,
 ]
+
+
+@workspace_mutations.field("getWorkspaceToken")
+def resolve_get_workspace_token(_, info, **kwargs):
```
pvanliefland, line 278: `generate_workspece_token`?

pvanliefland (changes requested):
Nice work, the only think I'm really not convinced about is setting a regular user with extra properties on the request instance for authentication.

I also added a couple of minor renaming suggestions.

hexa/workspaces/middlewares.py
pvanliefland: Actually the user class might not be a good idea - it makes sense for a pipeline run but in this case it's really the user that is performing operations.

What should be changed IMHO is to set the workspace / bypass params on the request and not on the user.

hexa/workspaces/graphql/schema.graphql
qgerome: Only the first call of the mutation will create the token. The next ones will retrieve the value already in DB. That's why I think "generate" is not the appropriate verb. WDYT?

hexa/workspaces/middlewares.py
qgerome: It clearly is. That's the problem of setting a `request.user` :/

hexa/workspaces/graphql/schema.graphql
qgerome: I renamed it `generateWorkspaceToken`...

PR #456, opened 2023-04-12
Add missing keys in .env.dist & docker-compose

pvanliefland:
It seems that some required keys we missing from the docker compose file - and from `.env.dist` as well.


pvanliefland:
@qgerome @cheikhgwane can you check? Did I miss something?

cheikhgwane:
LGTM !

cheikhgwane (approved):

PR #462, opened 2023-04-24
chore(Docker): Create docker profiles to run the pipelines & datawork…

qgerome:
The goal of the PR is to make it easy for developers to run the optional services mentioned in the docker-compose file by using the `--profile` flag of docker-compose.

The Dockerfile has been updated to not include the docker-ce package by default in the app image.


pvanliefland (approved):
I'm not familiar with docker compose profiles, but looks good to me. Feel free to merge if it works fine on your environment.

PR #471, opened 2023-05-09
feat(Database): add database credentials

cheikhgwane:
Add database credentials (username, port, host and database name) to schema.

## Changes

- Add database credentials to schema


docker-compose.yaml
```diff
@@ -44,6 +44,7 @@ x-app: &common
       - WORKSPACES_DATABASE_HOST=db
       - WORKSPACES_DATABASE_PORT=5432
       - WORKSPACES_DATABASE_DEFAULT_DB=hexa-app
+      - WORKSPACES_DATABASE_PROXY_URL=explore.openhexa.org
```
qgerome, line 47: The proxy does not exist in dev but we could try with `db` (the name of the service in the docker-compose.yaml)

hexa/databases/graphql/schema.graphql
```diff
@@ -1,6 +1,11 @@
 type Database {
   tables(page:Int, perPage:Int): DatabaseTablePage!
   table(name: String!): DatabaseTable
+  name: String!
+  username: String!
+  host: String!
+  port: Int!
+  externalUrl: String!
```
qgerome, line 8: We would need to have the password here as well (otherwise it's not possible to connect to the database). I'd only return the password if the user is an admin of the workspace

qgerome (changes requested):

hexa/databases/graphql/schema.graphql
cheikhgwane: All right.

hexa/databases/schema.py
```diff
@@ -38,6 +40,35 @@ def resolve_database_table(workspace, info, **kwargs):
...
+def resolve_database_host(_, info, **kwargs):
+    return get_db_server_credentials()["host"]
+
+
+@database_object.field("password")
+def resolve_database_password(workspace, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    return (
+        workspace.db_password
+        if request.user.has_perm("workspaces.manage_members", workspace)
```
qgerome, line 58: The permission is not the right one. You can just check if the user is an admin of the workspace

config/settings.py
```diff
@@ -419,6 +419,7 @@ def sentry_tracer_sampler(sampling_context):
 WORKSPACES_DATABASE_HOST = os.environ.get("WORKSPACES_DATABASE_HOST")
 WORKSPACES_DATABASE_PORT = os.environ.get("WORKSPACES_DATABASE_PORT")
 WORKSPACES_DATABASE_DEFAULT_DB = os.environ.get("WORKSPACES_DATABASE_DEFAULT_DB")
+WORKSPACES_DATABASE_PROXY_URL = os.environ.get("WORKSPACES_DATABASE_PROXY_URL")
```
qgerome, line 422: Could you rename the PROXY_URL to PROXY_HOST ?

hexa/databases/graphql/schema.graphql
pvanliefland: - `host` should be `slug.explore.openhexa.org`
- `externalUrl` should be replace by `url` and should have the value of the connection string, like the one we use for SQLAlchemy

pvanliefland: Also, we should use have something like this

```graphql
type Database {
  tables(page:Int, perPage:Int): DatabaseTablePage!
  table(name: String!): DatabaseTable
  credentials: DatabaseCredentials
}

type DatabaseCredentials {
  dbName: String!
  username: String!
  password: String!
  host: String!
  port: Int!
  url: String!
}
```

hexa/databases/schema.py
```diff
@@ -38,6 +40,37 @@ def resolve_database_table(workspace, info, **kwargs):
...
+    return get_db_server_credentials()["port"]
+
+
+@database_object.field("host")
+def resolve_database_host(_, info, **kwargs):
+    return get_db_server_credentials()["host"]
+
+
+@database_object.field("password")
+def resolve_database_password(workspace, info, **kwargs):
```
pvanliefland, line 54: In the end, we will have a `resolve_database_credentials` resolver. 

It should call `has_perm("database.view_database_credentials")`. And in this new permission function, you should check whether the user is either admin or editor of the workspace.

pvanliefland (changes requested):
A few change requests :

hexa/databases/schema.py
```diff
@@ -38,6 +41,42 @@ def resolve_database_table(workspace, info, **kwargs):
     return get_table_definition(workspace, kwargs.get("name"))
 
 
+@database_object.field("credentials")
```
pvanliefland, line 44: @cheikhgwane this will work but it would be easier to return a simple dict (instead of returning the workspace and then writing a bunch of resolvers) with all the connection parameters.

pvanliefland: Also, could we encapsulate the different connection parameters in the workspace model? (thinking about port / url / host etc). They could be defined as @property methods.

pvanliefland (changes requested):
Hey @cheikhgwane , still a couple of remarks.

Could we also make sure that the environment variables for the workspace database use the same URL / host than the one provided by GraphQL?

cheikhgwane:
> Hey @cheikhgwane , still a couple of remarks.
> 
> Could we also make sure that the environment variables for the workspace database use the same URL / host than the one provided by GraphQL?

Got it.

hexa/databases/schema.py
cheikhgwane: Like the username, define properties like `workspace_db_host`?

pvanliefland: `db_host`, `db_port`, `db_url`... That way the logic is encapsulated in the workspace and does not need to be copied / pasted elsewhere.

cheikhgwane: Yes, make sense.

pvanliefland (approved):
LGTM!

PR #477, opened 2023-05-22
fix(Workspaces): fix workspace member role update

cheikhgwane:
Fix broken update on workspace's member role update.

## Changes

- Change filter_for_user query
- Ensure that a user cannot update/delete his own membership

## How/what to test

- Test update of another user's role.
- Check if you can delete/update your own role


cheikhgwane:
@pvanliefland For the point `Make sure that there is always an admin (i.e admins can't downgrade their role or remove themselves if there is no other admin)` if we take these assumptions :

PR #490, opened 2023-06-22
fix(GCS): Add a content-disposition:download on object download url

qgerome:
When users wanted to download files, some triggered a the browser to download the object directly while other opened a new windows with the content of the file. 

We now ask to have a "download" content disposition set on the generated download url created by google.


pvanliefland (approved):
Nice! Just a small suggestion - feel free to ignore.

pvanliefland (approved):
👍

PR #493, opened 2023-07-05
Feat/workspace external members

cheikhgwane:
Allow workspace admin to add users outside of openhexa users.

## Changes

- Update inviteMember mutation to allow external users. 
- Generate a timestamped token for invitation
- Send mail with link to account creation

## Screenshots / screencast


hexa/workspaces/tests/test_schema/test_workspace.py
```diff
@@ -815,3 +783,50 @@ def test_generate_workspace_token_non_authorized(self):
...
+                {
+                    "input": {
+                        "workspaceSlug": self.WORKSPACE.slug,
+                        "userEmail": user_email,
+                        "role": WorkspaceMembershipRole.VIEWER,
+                    }
+                },
+            )
+
+            invitation = WorkspaceInvitation.objects.filter(email=user_email).first()
```
qgerome, line 821: Since you should only have one invitation, I'd recommend to use `WorkspaceInvitation.objects.get(workspace=self.WORKSPACE, email=user_email)`

hexa/workspaces/tests/test_schema/test_workspace.py
```diff
@@ -815,3 +783,50 @@ def test_generate_workspace_token_non_authorized(self):
...
+            self.assertIsNotNone(invitation)
```
qgerome, line 822: This test is useless since you access it directly after

hexa/workspaces/templates/workspaces/mails/invite_external_user.txt
```diff
@@ -0,0 +1,5 @@
+You have been invited to join the workspace {{workspace}} by {{owner}}.
+
+Use the following link proceed {{ workspace_signup_url }}
```
qgerome, line 3:
```suggestion
Create your account using this link: {{ workspace_signup_url }}
```

hexa/workspaces/templates/workspaces/mails/invite_external_user.txt
```diff
@@ -0,0 +1,5 @@
...
+
```
qgerome, line 4: Add a blank line

hexa/workspaces/templates/workspaces/mails/invite_external_user.html
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% autoescape off %}
+{% blocktranslate %}You have been invited to join the workspace <b>{{workspace}}</b> by {{owner}}.{% endblocktranslate %}
+
+{% translate "Use the following link to proceed:" %}
+
+<a href="{{ workspace_signup_url }}">{% translate "Accept invitation" %}</a>
```
qgerome, line 7: Since we do not handle yet the case where you accept the invitation, I think it would be better to write "Create your account" or something like this

hexa/workspaces/schema/mutations.py
```diff
@@ -114,23 +118,60 @@ def resolve_create_workspace_member(_, info, **kwargs):
...
-        )
-        send_mail(
-            title=gettext_lazy(f"You've been added to the workspace {workspace.name}"),
-            template_name="workspaces/mails/invite_member",
-            template_variables={
-                "workspace": workspace.name,
-                "owner": request.user.display_name,
-                "workspace_url": request.build_absolute_uri(
-                    f"//{settings.NEW_FRONTEND_DOMAIN}/workspaces/{workspace.slug}"
+        _user = User.objects.filter(email=input["userEmail"])
```
qgerome, line 121: Could you rename `def resolve_create_workspace_member` to the correct mutation name ?

hexa/workspaces/schema/mutations.py
```diff
@@ -114,23 +118,60 @@ def resolve_create_workspace_member(_, info, **kwargs):
...
+            }
+        else:
+            invitation = WorkspaceInvitation.objects.create_if_has_perm(
+                principal=request.user,
+                workspace=workspace,
+                email=input["userEmail"],
+                role=input["role"],
+            )
+            signer = TimestampSigner()
+            token = base64.b64encode(
```
qgerome, line 154: I think it would be better to move this on the model as a method you can call

hexa/workspaces/models.py
```diff
@@ -284,6 +285,57 @@ def delete_if_has_perm(self, *, principal: User):
...
+        )
+
+
+class WorkspaceInvitation(Base):
+    email = CIEmailField()
+    workspace = models.ForeignKey(
+        Workspace,
+        on_delete=models.CASCADE,
+    )
+    inviter = models.ForeignKey(
```
qgerome, line 325:
```suggestion
    invited_by = models.ForeignKey(
```

hexa/workspaces/models.py
```diff
@@ -284,6 +285,57 @@ def delete_if_has_perm(self, *, principal: User):
...
+        "user_management.User",
+        on_delete=models.CASCADE,
```
qgerome, line 327: I'd prefer to set this to null instead of removing the row

hexa/workspaces/schema/mutations.py
```diff
@@ -114,23 +118,60 @@ def resolve_create_workspace_member(_, info, **kwargs):
...
+                signer.sign(invitation.id).encode("utf-8")
+            ).decode()
+            user_email = parse.quote(input["userEmail"])
+            params = {
+                "title": gettext_lazy(
+                    f"You've been invited to join the workspace {workspace.name}"
```
qgerome, line 160:
```suggestion
                    f"You've been invited to join the workspace {workspace.name} on OpenHexa"
```

hexa/workspaces/schema/mutations.py
```diff
@@ -114,23 +118,60 @@ def resolve_create_workspace_member(_, info, **kwargs):
...
+                principal=request.user,
+                workspace=workspace,
+                email=input["userEmail"],
+                role=input["role"],
+            )
+            signer = TimestampSigner()
+            token = base64.b64encode(
+                signer.sign(invitation.id).encode("utf-8")
+            ).decode()
+            user_email = parse.quote(input["userEmail"])
```
qgerome, line 157: Since you store It in your model, you can safely use the value without using `parse.quote`/

hexa/workspaces/schema/mutations.py
```diff
@@ -114,23 +118,60 @@ def resolve_create_workspace_member(_, info, **kwargs):
...
+                    "workspace": workspace.name,
+                    "owner": request.user.display_name,
+                    "workspace_signup_url": request.build_absolute_uri(
+                        f"//{settings.NEW_FRONTEND_DOMAIN}/workspaces/{workspace.slug}/signup?user={user_email}&token={token}"
+                    ),
+                },
+                "recipient_list": [input["userEmail"]],
+            }
+
+        send_mail(**params)
```
qgerome, line 173: I'd prefer to see the two cases having their send_mail instead of building the params dictionary

hexa/workspaces/schema/mutations.py
qgerome: I think I would handle this using a try..catch (User.DoesNotExist) block

qgerome (changes requested):
In addition to the schema tests, could you write tests for the models. The test for a already existing user would be a nice addition as well

hexa/workspaces/templates/workspaces/mails/invite_external_user.html
cheikhgwane: Yes, you're right

pvanliefland (approved):
@cheikhgwane nice work! I agree with @qgerome 's remarks, and it seems that most or all of them are already implemented.

hexa/workspaces/schema/mutations.py
```diff
@@ -107,55 +113,150 @@ def resolve_archive_workspace(_, info, **kwargs):
...
+    request: HttpRequest = info.context["request"]
+    input = kwargs["input"]
+
+    try:
+        signer = TimestampSigner()
+        decoded_value = b64decode(input["token"]).decode("utf-8")
+        # the token is valid for 48h
+        invitation_id = signer.unsign(decoded_value, max_age=48 * 3600)
+
+        invitation = WorkspaceInvitation.objects.get(id=invitation_id)
```
qgerome, line 206: I don't see a `catch WorkspaceInvitation.DoesNotExist`, I think it would be needed

hexa/workspaces/schema/mutations.py
```diff
@@ -107,55 +113,150 @@ def resolve_archive_workspace(_, info, **kwargs):
...
+        if invitation.status == WorkspaceInvitationStatus.ACCEPTED:
+            raise AlreadyExists
+        workspace = invitation.workspace
```
qgerome, line 209: nit pick: just use invitation.workspace below

hexa/workspaces/schema/mutations.py
```diff
@@ -107,55 +113,150 @@ def resolve_archive_workspace(_, info, **kwargs):
...
+
+        if WorkspaceMembership.objects.filter(
+            user__email=invitation.email, workspace=workspace
+        ).exists():
+            raise AlreadyExists(
+                f"Already got a membership for {invitation.email} and workspace {workspace.name}"
+            )
+
+        if input["password"] != input["confirmPassword"]:
+            raise ValidationError("The two passwords didn't match.")
```
qgerome, line 219:
```suggestion
            raise ValidationError("The two passwords do not match.")
```

qgerome (changes requested):
I think that in terms of test, there are a few paths that are not tested (membership already approved, user already member of the workspace, ...)

hexa/workspaces/graphql/schema.graphql
```diff
@@ -184,6 +184,28 @@ enum GenerateWorkspaceTokenError {
...
+  lastName: String!
+  password: String!
+  confirmPassword: String!
+}
+
+type JoinWorkspaceResult {
+  success: Boolean!
+  token: String
+  workspace: Workspace
+  errors: [String!]!
```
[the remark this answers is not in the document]
cheikhgwane: Oh yes, I forgot to change it. Thanks

hexa/workspaces/graphql/schema.graphql
```diff
@@ -184,6 +184,28 @@ enum GenerateWorkspaceTokenError {
...
+  token: String!
+  firstName: String!
+  lastName: String!
+  password: String!
+  confirmPassword: String!
+}
+
+type JoinWorkspaceResult {
+  success: Boolean!
+  token: String
```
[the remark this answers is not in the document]
cheikhgwane: No , I will remove it

cheikhgwane:
> I think that in terms of test, there are a few paths that are not tested (membership already approved, user already member of the workspace, ...)

The test for an membership already accepted is https://github.com/BLSQ/openhexa-app/blob/cf021afb38b303dfec6c7c729963ab2ea9013354/hexa/workspaces/tests/test_schema/test_workspace.py#L949

The one for the user already member of the workspace : https://github.com/BLSQ/openhexa-app/blob/cf021afb38b303dfec6c7c729963ab2ea9013354/hexa/workspaces/tests/test_schema/test_workspace.py#L995

WDYT about them?

qgerome (approved):
Great !

hexa/workspaces/graphql/schema.graphql
```diff
@@ -9,8 +9,31 @@ type Workspace {
...
+  items: [WorkspaceInvitation!]!
+}
+
+type WorkspaceInvitation {
+  id: UUID!
+  email: String!
+  workspace: Workspace!
+  role: WorkspaceMembershipRole!
+  status: WorkspaceInvitationStatus!
+  invited_by: User!
```
qgerome, line 28: Since the user of a invitation can be deleted (user is deleted from the database), this field is nullable

hexa/workspaces/schema/types.py
```diff
@@ -96,6 +102,26 @@ def resolve_workspace_members(workspace: Workspace, info, **kwargs):
     )
 
 
+@workspace_object.field("invitations")
+def resolve_workspace_invitations(workspace: Workspace, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+
+    status = [WorkspaceInvitationStatus.ACCEPTED, WorkspaceInvitationStatus.PENDING]
+    if kwargs.get("status"):
```
qgerome, line 110: I would prefer to have a second filter on the query set in case there is a condition on the status

```python
if kwargs.get("status"):
  qs = qs.filter(status=kwargs["status"])



```

hexa/workspaces/schema/types.py
```diff
@@ -96,6 +102,26 @@ def resolve_workspace_members(workspace: Workspace, info, **kwargs):
...
+
+    qs = (
+        WorkspaceInvitation.objects.filter_for_user(request.user)
+        .filter(status__in=status)
+        .order_by("-updated_at")
+    )
+    return result_page(
+        queryset=qs,
+        page=kwargs.get("page", 1),
+        per_page=kwargs.get("perPage", qs.count()),
```
qgerome, line 121: I don't think it's a good idea to return by default the whole list if the users do not give the perPage argument.
Better to have a default value (like "15")

qgerome (changes requested):

hexa/workspaces/schema/types.py
cheikhgwane: Yes, you're right .

qgerome (approved):

PR #497, opened 2023-07-06
fix(Connections): Returns the secret value for members that can edit them

qgerome:
When a user edited any field of a connection that had secret fields, those secret fields were reset to an empty value

## Changes

We now return the value of the secret fields when the user is at least an member with the editor role

## How/what to test

- Create a connection with at least 2 fields (one secret and one non secret).
- Edit the non secret field
- The secret field value should not have changed


cheikhgwane (approved):
LGTM!

PR #500, opened 2023-07-12
chore(Workspaces): use proxy for db_url

cheikhgwane:
Use proxy for WORKSPACE_DATABASE_URL.

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast

### DIFF
--- a/hexa/workspaces/models.py
+++ b/hexa/workspaces/models.py
@@ -152,8 +152,7 @@ class Workspace(Base):
 
     @property
     def db_url(self):
-        host = get_db_server_credentials()["host"]
-        return f"postgresql://{self.db_name}:{self.db_password}@{host}:{self.db_port}/{self.db_name}"
+        return f"postgresql://{self.db_name}:{self.db_password}@{self.db_host}:{self.db_port}/{self.db_name}"
 
     def update_if_has_perm(self, *, principal: User, **kwargs):
         if not principal.has_perm("workspaces.update_workspace", self):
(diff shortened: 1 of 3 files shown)


hexa/workspaces/models.py
```diff
@@ -152,8 +152,7 @@ def db_port(self):
 
     @property
     def db_url(self):
-        host = get_db_server_credentials()["host"]
-        return f"postgresql://{self.db_name}:{self.db_password}@{host}:{self.db_port}/{self.db_name}"
+        return f"{self.slug}.{settings.WORKSPACES_DATABASE_PROXY_HOST}"
```
pvanliefland, line 155: @cheikhgwane only the IP should be replaced by xyz.explore.openhexa.org.

By the way, `f"{self.slug}.{..."` could be replaced by `self.db_host` ;)

pvanliefland (changes requested):
@cheikhgwane, see my comment about the url generation.

hexa/workspaces/models.py
cheikhgwane: Ah yes, my bad.

cheikhgwane: Something like this ?
```python
f"postgresql://{self.db_name}:{self.db_password}@{settings.WORKSPACES_DATABASE_PROXY_HOST}:{self.db_port}/{self.db_name}"
```

pvanliefland (approved):
LGTM!

PR #505, opened 2023-07-18
fix(Connections): Add a suffix to connection slug if it exists

qgerome:
Before creating a connection from the slug provided by the API of the name, we check for the existence of a workspace connection with this slug In case we find one, we add a random suffix to the slug of the new connection.


pvanliefland (approved):
I'm ok with this, just wondering if we should always add the suffix so that connection identifiers always look the same.

What do you think?

qgerome:
My thoughts were : If the user adds a single connection for a db in his workspace, would it make sense to always add a random part to it or can we just use the name to create a nice and easy to remember slug ?

So I’d leave it as it is peronally

PR #507, opened 2023-07-19
feat(Pipelines): prevent push of version with params for scheduled pipeline.

cheikhgwane:
Prevent user to push a new version with parameters for a scheduled pipeline.
## Changes

- Reject upload of a new version with params if current version is scheduled
- write data migration to unschedule all previous scheduled pipelines with parameters.

## How/what to test

## Screenshots / screencast


hexa/pipelines/models.py
```diff
@@ -188,6 +192,10 @@ def last_run(self) -> "PipelineRun":
 
     def upload_new_version(self, user: User, zipfile, parameters):
         if self.last_version:
+            if self.schedule and parameters:
+                raise InvalidDefinition(
```
pvanliefland, line 196: In this case you should raise a more specific exception, like `PipelineDoesNotSupportParametersError`. Otherwise how would you know later which error code to return?

pvanliefland: The other option would be to have an InvalidDefinitionError exception class with an error code enum but I don't see the added value compared to an explicit exception class.

pvanliefland (changes requested):
Nice work - the last thing to fix imho is the exception. Also requesting @qgerome 's opinion on this specific question.

hexa/pipelines/models.py
cheikhgwane: Indeed, you're right.I'll go with the first option.

pvanliefland (approved):
LGTM

PR #508, opened 2023-07-20
feat(Workspaces): add workspace invitations list to django admin

cheikhgwane:
Setup workspace invitations on django admin view.

## Changes

- Register WorkspaceInvitation to django admin. 

## How/what to test

## Screenshots / screencast

image

### DIFF
--- a/hexa/workspaces/admin.py
+++ b/hexa/workspaces/admin.py
@@ -1,6 +1,12 @@
 from django.contrib import admin
 
-from .models import Connection, ConnectionField, Workspace, WorkspaceMembership
+from .models import (
+    Connection,
+    ConnectionField,
+    Workspace,
+    WorkspaceInvitation,
+    WorkspaceMembership,
+)
 
 
 @admin.register(Workspace)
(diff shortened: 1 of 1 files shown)


hexa/workspaces/admin.py
```diff
@@ -42,3 +48,16 @@ class ConnectionFieldAdmin(admin.ModelAdmin):
         "connection",
         "code",
     )
+
+
+@admin.register(WorkspaceInvitation)
+class WorkspaceInvitationAdmin(admin.ModelAdmin):
+    list_display = (
+        "id",
```
pvanliefland, line 56: We can probably skip the `id` and keep `email` as the first column

pvanliefland (changes requested):
See comment about `id` - otherwise, looking good

hexa/workspaces/admin.py
cheikhgwane: Okay.

pvanliefland (approved):
👍

PR #511, opened 2023-07-24
feat(Workspaces): set demo data optional

cheikhgwane:
Makes optional demo data loaded when creating a workspace.

## Changes

- update schema, mutation and model to handle to make demo data optional.

## How/what to test

## Screenshots / screencast

### DIFF
--- a/hexa/workspaces/models.py
+++ b/hexa/workspaces/models.py
@@ -66,6 +66,7 @@ class WorkspaceManager(models.Manager):
         name: str,
         description: str = None,
         countries: typing.Sequence[Country] = None,
+        load_sample_data: bool = False,
     ):
         if not principal.has_perm("workspaces.create_workspace"):
             raise PermissionDenied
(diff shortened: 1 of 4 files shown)


hexa/workspaces/models.py
```diff
@@ -83,14 +84,15 @@ def create_if_has_perm(
...
         create_kwargs["db_name"] = db_name
-
         create_database(db_name, db_password)
-        load_database_sample_data(db_name)
 
         bucket = create_bucket(settings.WORKSPACE_BUCKET_PREFIX + slug)
-        load_bucket_sample_data(bucket.name)
         create_kwargs["bucket_name"] = bucket.name
 
+        if load_sample_data:
```
qgerome, line 92: Could you add a test for this please ?

qgerome (changes requested):

hexa/workspaces/models.py
cheikhgwane: Yes !

PR #515, opened 2023-07-31
feat(Pipelines): define timeout for pipeline run

cheikhgwane:
Relative to [issue](https://github.com/BLSQ/openhexa-team/issues/485). We introduce a timeout mechanism that will stop  pods that continue runs  even after finishing their job.

## Changes

- Set a default activeDeadlineSeconds (timeout) to 4hours 
- Update schema 

## How/what to test

## Screenshots / screencast


docker-compose.yaml
```diff
@@ -30,6 +30,8 @@ x-app: &common
       - NEW_FRONTEND_DOMAIN=localhost:3000
       - PIPELINE_SCHEDULER_SPAWNER=docker
       - PIPELINE_API_URL=http://app:8000
+      - PIPELINE_RUN_DEFAULT_TIMEOUT=14400
+      - PIPELINE_RUN_MAX_TIMEOUT=43200
```
pvanliefland, line 34: This is not necessary, the defaults in `settings.py` will do the job

hexa/pipelines/schema.py
```diff
@@ -270,13 +271,22 @@ def resolve_create_pipeline(_, info, **kwargs):
         }
 
     try:
+        if input.get("timeout") and input.get("timeout") > int(
+            os.environ.get("PIPELINE_RUN_MAX_TIMEOUT")
```
pvanliefland, line 275: You should use the Django settings instead of directly accessing the env variable (`from django.conf import settings`)

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -33,6 +38,12 @@ def run_pipeline_kube(run: PipelineRun, env_var: dict):
         + "-"
         + exec_time_str
     )
+    pipeline_timeout = (
+        run.pipeline.timeout
+        if run.pipeline.timeout
+        else os.environ.get("PIPELINE_RUN_DEFAULT_TIMEOUT")
```
pvanliefland, line 44: Please use Django settings instead of accessing the env variable directly (`from django.cong import settings`)

hexa/pipelines/schema.py
```diff
@@ -270,13 +271,22 @@ def resolve_create_pipeline(_, info, **kwargs):
...
+        ):
+            raise Exception("Pipeline timeout greater than the maximum allowed value.")
```
pvanliefland, line 277: I think we should return a specific error code for this, that the SDK can understad

pvanliefland (changes requested):
Hey @cheikhgwane , nice work, here are a couple of initial comments

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -18,6 +19,10 @@
 logger = getLogger(__name__)
 
 
+class PodTerminationReason(Enum):
```
pvanliefland, line 22: It seems that this class is not used - can you confirm?

pvanliefland (commented):
Hey @cheikhgwane , I still have a question, could you check?

hexa/pipelines/management/commands/pipelines_runner.py
cheikhgwane: Yes, I've just added a check for the termination reason to see if it was due to timeout and adapt the stdout.

pvanliefland (approved):
LGTM!

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -178,6 +190,9 @@ def run_pipeline_kube(run: PipelineRun, env_var: dict):
     except Exception:
         logger.exception("get logs")
         stdout = ""
+    # check termination reason
+    if remote_pod.status.reason == PodTerminationReason.DeadlineExceeded.value:
+        stdout = f"Timeout killed run {run.pipeline.name} #{run.id}"
```
pvanliefland, line 195: I'm not sure we should discard the previous value of stdout.

Also, in that case, wouldn't stdout already contain something about the timeout?

pvanliefland (commented):
I think we are almost there. See my comment and questions about stdout...

hexa/pipelines/management/commands/pipelines_runner.py
cheikhgwane: If I remember correctly the deadlineExceeded will not be on the pod logs, but I'll double check to be more sure about it.

pvanliefland: Otherwise, could we try to append to `stdout` ?

cheikhgwane: Yes, it's safer.

cheikhgwane: @pvanliefland should be good now

PR #517, opened 2023-08-04
feat(Files): Show/hide hidden files & directories

qgerome:
- It adds a parameter on bucket.objects() to hide/display hidden objects (starting with ".").
- A client.list_blobs mock has been implemented as well


cheikhgwane (approved):

pvanliefland (changes requested):
As discussed

pvanliefland (approved):
LGTM!

PR #523, opened 2023-08-09
feat(Workspaces): improve invitation management

cheikhgwane:
Improve workspace invitation management.

## Changes

- Allow admin to delete pending invitations.
- Re-use existing invitation (Pending) if admin want to invite the same user.

## How/what to test

## Screenshots / screencast


hexa/workspaces/schema/mutations.py
```diff
@@ -150,13 +150,19 @@ def resolve_invite_workspace_member(_, info, **kwargs):
...
             }
         except User.DoesNotExist:
-            # If the user does not exist, we create an invitation to create an account and join the workspace
-            invitation = WorkspaceInvitation.objects.create_if_has_perm(
-                principal=request.user,
-                workspace=workspace,
-                email=input["userEmail"],
-                role=input["role"],
-            )
+            # If the user does not exist we first
```
qgerome, line 153: Better to let the code crash thanks to a unique constraint

qgerome (changes requested):
From what I see in the task, we should also have a "resend invitation" mutation.

hexa/workspaces/schema/mutations.py
```diff
@@ -196,7 +195,7 @@ def resolve_invite_workspace_member(_, info, **kwargs):
 
 
 @workspace_mutations.field("joinWorkspace")
-def resolver_join_workspace(_, info, **kwargs):
+def resolve_join_workspace(_, info, **kwargs):
```
qgerome, line 198: thanks

hexa/workspaces/schema/mutations.py
```diff
@@ -266,8 +265,57 @@ def resolver_join_workspace(_, info, **kwargs):
...
+def resolve_resend_workspace_invitation(_, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    input = kwargs["input"]
+
+    try:
+        invitation = WorkspaceInvitation.objects.get(id=input["invitationId"])
+        if not request.user.has_perm("workspaces.manage_members", invitation.workspace):
+            raise PermissionDenied
+
+        if invitation.status == WorkspaceInvitationStatus.ACCEPTED:
```
qgerome, line 278: I think you can use the same use case than the invitation not found above

hexa/workspaces/schema/mutations.py
```diff
@@ -266,8 +265,57 @@ def resolver_join_workspace(_, info, **kwargs):
         }
 
 
+@workspace_mutations.field("resendWorkspaceInvitation")
+def resolve_resend_workspace_invitation(_, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    input = kwargs["input"]
+
+    try:
+        invitation = WorkspaceInvitation.objects.get(id=input["invitationId"])
```
qgerome, line 274:
```suggestion
        invitation = WorkspaceInvitation.objects.filter(status=WorkspaceInvitationStatus.PENDING).get(id=input["invitationId"])
```

hexa/workspaces/schema/mutations.py
```diff
@@ -266,8 +265,57 @@ def resolver_join_workspace(_, info, **kwargs):
...
+        if not request.user.has_perm("workspaces.manage_members", invitation.workspace):
+            raise PermissionDenied
+
+        if invitation.status == WorkspaceInvitationStatus.ACCEPTED:
+            raise AlreadyExists("Invitation already accepted.")
+
+        token = invitation.generate_invitation_token()
+        send_mail(
```
qgerome, line 282: If it's 99% the same code as the initial invite, it's better to make a function of this and call it at the different places.

hexa/workspaces/schema/mutations.py
```diff
@@ -266,8 +265,57 @@ def resolver_join_workspace(_, info, **kwargs):
...
+        return {
+            "success": False,
+            "errors": ["PERMISSION_DENIED"],
+        }
+    except WorkspaceInvitation.DoesNotExist:
+        return {
+            "success": False,
+            "errors": ["INVITATION_NOT_FOUND"],
+        }
+    except AlreadyExists:
```
qgerome, line 310: This can be can be removed then

qgerome (changes requested):

qgerome (approved):

qgerome (changes requested):
In fact, I don't see the constraint on the workspace and email address that would be useful to not having multiple invitations with the same email addresses

qgerome (changes requested):
You still need to catch the error thrown by Postgresql and return a proper error to the frontend (AlreadyExists)

cheikhgwane:
> You still need to catch the error thrown by Postgresql and return a proper error to the frontend (AlreadyExists)

I've updated the resolver inviteWorkspaceMember. For the resend we do no try to create a new one.

qgerome (approved):

PR #525, opened 2023-08-10
fix(files): Break when we have enough objects (and before loading the next page)

qgerome:
The pagination was broken when the user had a certain number of files in his buckets. It was displayed that no next page existed but it was a LIE.

The problem came from `current_page = next(page)` that was always called and could result in a `StopIteration` error.

## Changes

I check in the loop if we have enough objects and break the loop early

### DIFF
--- a/hexa/files/api.py
+++ b/hexa/files/api.py
@@ -173,7 +173,8 @@ def list_bucket_objects(
             res = _prefix_to_dict(bucket_name, prefix)
             if not ignore_hidden_files or not res["name"].startswith("."):
                 objects.append(res)
-        while len(objects) <= max_items:
+
+        while True:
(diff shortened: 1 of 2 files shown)


hexa/files/api.py
```diff
@@ -196,6 +203,7 @@ def list_bucket_objects(
     except StopIteration:
         # We reached the end of the list of pages. Let's return what we have and set the
         # has_next_page to false
+        print("StopIteration", flush=True)
```
pvanliefland, line 206: This should probably be removed

pvanliefland (commented):
@qgerome I think that we should have a test for that specific bug

pvanliefland (approved):
👍

PR #533, opened 2023-08-18
feat(Datasets): Data model

qgerome:
First implementation of the workspace's datasets. Users can interact with the datasets using the web interface or the SDK from their pipelines or notebooks.

The bucket where the dataset are stored is managed by Terraform.


hexa/datasets/models.py
```diff
@@ -0,0 +1,170 @@
...
+            user,
+            models.Q(dataset_version__workspace__members=user)
+            | models.Q(
+                dataset__in=Dataset.objects.filter(shares__workspace__members=user)
+            ),
+            return_all_if_superuser=False,
+        )
+
+
+class DatasetVersionObject(Base):
```
pvanliefland, line 150: I'm still wondering about `File` vs `Object`. `Object` feels like an implementation detail.

What would you think about `File` and `uri` vs `Object` and `key`?

hexa/datasets/models.py
```diff
@@ -0,0 +1,170 @@
...
+    name = models.TextField(null=False, blank=False)
+    key = models.TextField(null=False, blank=False)
```
pvanliefland, line 152: `key` should be unique

pvanliefland (commented):
Hey @qgerome , I think this is fine 👍

I'm just wondering whether we should decouple our data model from the bucket/object paradigm. Not sure about this, so feel free to ignore.

hexa/datasets/models.py
pvanliefland: Not 100% certain, so it's probably not worth changing what you implemented. In any case this can be changed later.

hexa/datasets/models.py
```diff
@@ -0,0 +1,219 @@
...
+        blank=False,
+        on_delete=models.CASCADE,
+        related_name="files",
+    )
+
+    objects = DatasetVersionFileQuerySet.as_manager()
+
+    @property
+    def full_uri(self):
+        return f"{self.dataset_version.dataset.id}/{self.dataset_version.id}/{self.uri}"
```
pvanliefland, line 175: Something to consider: maybe storing the full URI would be more robust - to discuss

hexa/files/api.py
```diff
@@ -253,8 +253,10 @@ def generate_download_url(bucket_name: str, target_key: str, force_attachment=Fa
     client = get_storage_client()
     gcs_bucket = client.get_bucket(bucket_name)
     blob: Blob = gcs_bucket.get_blob(target_key)
+    filename = blob.name.split("/")[-1]
```
pvanliefland, line 256: I'm curious about this change, why is it needed?

hexa/workspaces/models.py
```diff
@@ -141,6 +142,9 @@ class Workspace(Base):
         null=True,
     )
     archived = models.BooleanField(default=False)
+    shared_datasets = models.ManyToManyField(
```
pvanliefland, line 145: Not sure about the name. But I don't have a better idea at this point :)

pvanliefland:
Also, this PR deserves a description!

hexa/datasets/models.py
qgerome: Yes indeed

hexa/datasets/tests/test_models.py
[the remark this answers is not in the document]
qgerome: I think an intermediate step would be to not have a class of 1000 lines of tests but split them by context

hexa/files/api.py
qgerome: otherwise the name of the filename of the download file is my_dir-dir-file.jpg and not just file.jpg

qgerome:
> Also, this PR deserves a description!

It will have one

PR #534, opened 2023-08-18
fix(Pipelines): Do not throw an error when an output does not exist anymore

qgerome:
Sometimes the table or bucket object that was stored as output of a pipeline run has been deleted. Let's not include it in the outputs then.

## How/what to test

- Run a pipeline that creates a output.
- Delete this output from jupyter
- Try to open the page of the run


pvanliefland (approved):
I think it's ok for a quickfix, but ideally I think we should avoir catch-all clauses such as `except Exception` and handle exceptions explicitly.

PR #537, opened 2023-08-22
fix(Workspaces): Add the 'workspaces' feature to legacy users invited in a workspace

qgerome:
If a legacy user was invited, the workspaces feature flag was not attached to the user and thus he cannot access the workspace anyway.

This case was handled when inviting external users

## Changes

We check if the feature has to be activated for the user


cheikhgwane (approved):

pvanliefland (approved):
LGTM!

PR #538, opened 2023-08-22
fix(Workspaces): filter invitations by user and workspace

cheikhgwane:
Fix bug that allowed user from one workspace to be able to see invitations for others workspaces.

## Changes

- Add filter by workspace and Queryset result.

### DIFF
--- a/hexa/workspaces/tests/test_schema/test_workspace.py
+++ b/hexa/workspaces/tests/test_schema/test_workspace.py
@@ -97,6 +97,13 @@ class WorkspaceTest(GraphQLTestCase):
             role=WorkspaceMembershipRole.VIEWER,
             status=WorkspaceInvitationStatus.ACCEPTED,
         )
+        cls.INVITATION_BAZ = WorkspaceInvitation.objects.create(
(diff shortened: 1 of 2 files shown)


hexa/workspaces/tests/test_schema/test_workspace.py
```diff
@@ -97,6 +97,13 @@ def setUpTestData(cls):
             role=WorkspaceMembershipRole.VIEWER,
             status=WorkspaceInvitationStatus.ACCEPTED,
         )
+        cls.INVITATION_BAR = WorkspaceInvitation.objects.create(
```
qgerome, line 100: Shouldn't it use another name than the one just above ?

pvanliefland, line 100: It's a bit risky to reuse the same class variable here - maybe `cls.INVITATION_BAZ`?

qgerome (changes requested):

hexa/workspaces/tests/test_schema/test_workspace.py
cheikhgwane: Yes it should be BAZ. My bad

PR #541, opened 2023-09-07
fix(Connections): abort connection creation if an error occurs

cheikhgwane:
Fix relative to [issue](https://github.com/BLSQ/openhexa-team/issues/534) about connection that get created when one of his field is invalid.

## Changes

- Wrap connection creation logic into atomic transaction.
- Add test

## How/what to test

Try to add a connection with an invalid field name, the connection should be not created.

### DIFF
--- a/hexa/workspaces/schema/mutations.py
+++ b/hexa/workspaces/schema/mutations.py
@@ -5,7 +5,7 @@ from django.contrib.auth import authenticate, login
 from django.contrib.auth.password_validation import validate_password
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.core.signing import BadSignature, SignatureExpired, Signer
-from django.db import IntegrityError
+from django.db import IntegrityError, transaction
 from django.http import HttpRequest
 from django.utils.translation import gettext_lazy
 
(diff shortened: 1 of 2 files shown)


hexa/workspaces/schema/mutations.py
```diff
@@ -349,14 +349,15 @@ def resolve_create_workspace_connection(_, info, **kwargs):
     mutation_input = kwargs["input"]
 
     try:
-        workspace = Workspace.objects.filter_for_user(request.user).get(
-            slug=mutation_input.pop("workspaceSlug")
-        )
-        mutation_input["connection_type"] = mutation_input.pop("type")
+        with transaction.atomic():
+            workspace = Workspace.objects.filter_for_user(request.user).get(
```
qgerome, line 353: This can be done outside of the transaction no ?

cheikhgwane: Yes, it's possible.

qgerome: My point is just that since we do not try to protect `Workspace.objects.filter...` by an atomic transaction, it's better to move it two lines above.

cheikhgwane: Oh yes, my bad. Thx

cheikhgwane:
Voila @qgerome

qgerome (approved):

PR #542, opened 2023-09-11
feat(Pipelines): allow scheduling if all params have default value

cheikhgwane:
Relative to [issue](https://github.com/BLSQ/openhexa-team/issues/539) :  allow scheduling for pipelines if all parameters have default value.

## Changes

- Add  check on pipeline parameters default value

### DIFF
--- a/hexa/pipelines/models.py
+++ b/hexa/pipelines/models.py
@@ -129,7 +129,7 @@ class PipelineVersion(models.Model):
     @property
     def is_schedulable(self):
         return len(self.parameters) == 0 or all(
-            [parameter.get("default") for parameter in self.parameters]
+            [parameter.get("default") is not None for parameter in self.parameters]
         )
 
     @property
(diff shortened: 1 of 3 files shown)


qgerome (changes requested):

hexa/pipelines/models.py
```diff
@@ -126,6 +126,12 @@ class Meta:
 
     objects = PipelineVersionQuerySet.as_manager()
 
+    @property
+    def is_schedulable(self):
+        return len(self.parameters) == 0 or all(
+            [parameter.get("default") for parameter in self.parameters]
```
qgerome, line 132: That's not correct. What if it's a boolean ? False is a valid default

qgerome (changes requested):

hexa/pipelines/models.py
pvanliefland: This is a good example of something that could be tested @cheikhgwane

PR #552, opened 2023-09-21
Fix(Workspaces): ensure the presence of a token / server hash in memberships

pvanliefland:
## Changes

- Make sure that `access_token` (and `notebooks_server_hash` while we're at it) cannot be blank nor null in `WorkspaceMembership`
- Removes `generate_token` method - handle everythin in `save()` (that's what we were already doing for notebook server hash)
- Added an `RunPython` on he added migration that calls `save()` on every membership record without token so that it gets a token

## How/what to test

Create a workspace, check in the admin that it has a token and a notebook server hash.


pvanliefland (commented):
Note: let's put back the `blank=True` option, and add an admin action for regenerating the access token

PR #556, opened 2023-09-25
Fix datasets permissions management to be able to use them in the pipelines

qgerome:
This is a hack as we will revamp the way we manage the permissions of the users and pipeline' runs.


pvanliefland (approved):
Ok for me - let's just add FIXME in every change

PR #558, opened 2023-09-27
fix(Databases): get_table_rows has to take the page_size from the db return

qgerome:
`get_table_rows` always removed the last item of the items returned by the DB. In case we have the exact number of items, we should not remove the last one and just take the page size requested by the user instead.


cheikhgwane (approved):
LGTM!

pvanliefland (approved):
LGTM!

PR #562, opened 2023-09-30
fix: Do not allow slugs for workspace, datasets & connection to have double dash

qgerome:
This is to prevent the creation of slugs like `my-dataset--za21`


pvanliefland (approved):
It looks good, but I'm beginning to see some duplication here. I'll add a task about this in an upcoming sprint.

PR #564, opened 2023-10-06
Fix a wrong or incomplete instructions in README

toch:
When working on BLSQ/openhexa-team#565, I've found a few incorrect, incomplete, or misleading instructions.
The present changes bring a few minor quick corrections to those.

There are other changes to address those issues, but they'll require more work (e.g. fix the permission issue with test coverage reports). This is out of the current scope.

## Changes

- `README.md` :
  - The used Docker Compose command is deprecated
  - The host URL to access the app isn't mentioned
  - The way to get the test coverage report is incomplete, it requires to be in the container, also the issue with permission is not mentioned
  - It is not explained how you could run a single test

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,19 +93,19 @@ The following steps will get you up and running:
 ```bash
 cp .env.dist .env
 docker network create openhexa
-docker-compose build
-docker-compose run app fixtures
-docker-compose run app manage tailwind install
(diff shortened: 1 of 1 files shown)


pvanliefland (approved):
Perfect, thanks!

PR #566, opened 2023-10-09
fix(Connections): rename IASO connection api_url to url

cheikhgwane:
An error occurs when we  try to use  IASO connection on notebooks : there's a mismatch between what the sdk is looking for (IDENTIFIER_URL) and the field name (API_URL).


pvanliefland (changes requested):
@cheikhgwane could you add the reverse operation (`RunPython.noop()`)? I know that we have other migrations that miss it, but I'd prefer if we take the habit of having it.

cheikhgwane:
> @cheikhgwane could you add the reverse operation (`RunPython.noop()`)? I know that we have other migrations that miss it, but I'd prefer if we take the habit of having it.

Okay.

PR #572, opened 2023-10-13
fix(Pipelines): allow scheduling of pipelines

cheikhgwane:
Relative to issue (https://github.com/BLSQ/openhexa-team/issues/573). This PR improves pipelines with parameters scheduling by adding more flexibility on the way we allow/disallow scheduling.

## Changes

- Allow pipelines to be scheduled when all parameters are not required 
- Prevent scheduling of pipelines with required params without default value.

### DIFF
--- a/hexa/pipelines/models.py
+++ b/hexa/pipelines/models.py
@@ -129,16 +129,10 @@ class PipelineVersion(models.Model):
 
     @property
     def is_schedulable(self):
-        if len(self.parameters) == 0:
-            return True
-
-        for parameter in self.parameters:
-            if (parameter.get("required") is True) and (
-                parameter.get("default") is None
-            ):
-                return False
-
-        return True
+        return all(
+            parameter.get("required") is False or parameter.get("default") is not None
+            for parameter in self.parameters
+        )
 
     @property
     def display_name(self):
@@ -216,12 +210,8 @@ class Pipeline(models.Model):
         return self.pipelinerun_set.first()
 
     def upload_new_version(self, user: User, zipfile, parameters, timeout: int = None):
-        if self.last_version:
-            newnumber = self.last_version.number + 1
-        else:
-            newnumber = 1
-
-        version = PipelineVersion.objects.create(
+        newnumber = self.last_version.number + 1 if self.last_version else 1
+        version = PipelineVersion(
             user=user,
             pipeline=self,
             number=newnumber,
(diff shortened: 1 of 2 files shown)


hexa/pipelines/models.py
```diff
@@ -129,9 +129,16 @@ class Meta:
 
     @property
     def is_schedulable(self):
-        return len(self.parameters) == 0 or all(
-            [parameter.get("default") is not None for parameter in self.parameters]
-        )
+        if len(self.parameters) == 0:
```
pvanliefland, line 132: All of this could be expressed as a single `all` or `not any` statement

pvanliefland, line 132: All of this could be expressed as a single `all` or `not any` statement

hexa/pipelines/models.py
```diff
@@ -210,10 +217,6 @@ def last_run(self) -> "PipelineRun":
 
     def upload_new_version(self, user: User, zipfile, parameters, timeout: int = None):
         if self.last_version:
-            if self.schedule and parameters:
```
pvanliefland, line 213: We should still check `is_schedulable` here

hexa/pipelines/tests/test_schema.py
```diff
@@ -1061,46 +1195,6 @@ def test_pipelines_permissions_schedule_with_params_false(self):
             r["data"]["pipelineByCode"],
         )
 
-    def test_upload_pipeline_parameters_not_supported(self):
```
pvanliefland, line 1064: This test is still needed

pvanliefland (changes requested):
A few things to wrap up before merging this PR

hexa/pipelines/models.py
```diff
@@ -129,9 +129,19 @@ class Meta:
...
+                or (
+                    parameter.get("required") is True
+                    and parameter.get("default") is not None
+                )
+                for parameter in self.parameters
+            ]
+        ):
+            return True
+
+        return False
```
pvanliefland, line 144: All of this should be the same as

```python
return all(parameter.get("required") is False or parameter.get("default") is not None for parameter in self.parameters)
```

pvanliefland (changes requested):
Hey @cheikhgwane, a couple of remarks, some purely cosmetic, and a more important one about handling of unschedulable versions.

hexa/pipelines/tests/test_schema.py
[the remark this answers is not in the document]
pvanliefland: The rule is simple : if there is a `last_version` (which here mean the current version) and if it has a schedule, then the new version must be schedulable.

pvanliefland (commented):
@cheikhgwane we're getting there, I still have a suggestion regarding tests, let's discuss it tomorrow morning

PR #573, opened 2023-10-16
chore(Workspaces): add tests for connections env variables

cheikhgwane:
Add tests for workspace connections env variables.


pvanliefland (changes requested):
Looks good, let's add DHIS2 I would say

pvanliefland (approved):
Great, LGTM!

PR #578, opened 2023-10-18
fix: Add default value for AWS_DEFAULT_REGION in env dist file

pvanliefland:
As of now, if you follow the instructions in the README.md file to create your dev setup, some test from the `connector_s3` module will fail.

This PR adds a default value for AWS_DEFAULT_REGION in the `env.dist` file so that tests can pass on a fresh local install.

The value is not sensitive, and this setting will disappear once we deprecate the legacy features, so I think this is the most practical way to deal with it for now.


toch:
LGTM :+1:

toch (approved):
LGTM :+1:

PR #581, opened 2023-10-23
fix: Use blsq/openhexa-base-notebook:latest by default for pipelines

pvanliefland:
We no longer use openhexa-pipelines


cheikhgwane (approved):
LGTM

PR #584, opened 2023-10-30
chore : remove django-tailwind

cheikhgwane:
Remove django-tailwind. (See https://github.com/BLSQ/openhexa-team/issues/598) for reference.

## Changes

- Add generated css file to static folder.
- Delete static_src folder.


Dockerfile
```diff
@@ -24,14 +24,6 @@ CMD start
 
 FROM deps as app
```
pvanliefland, line 25: It seems that this whole build stage can be removed - could you try? We should probably rename deps to app.

pvanliefland (changes requested):
@cheikhgwane nice work, could you just check whether we can get rid of one of the Dockerfile build step.

Dockerfile
cheikhgwane: Okay.

pvanliefland (approved):
LGTM, let's test it!

PR #591, opened 2023-11-16
fix: Fix permission check for dataset download

pvanliefland:
Seems that we weren't using the proper check for the download dataset version file mutation.

## Changes

- Fix permission check
- Add test
- Add missing pieces to the GCS mock setup

## How/what to test

Download dataset files as a viewer.


cheikhgwane (approved):
LGTM

PR #600, opened 2023-12-04
fix(Pipelines): run pipeline with cloudrun argument

cheikhgwane:
The run of pipelines in local with docker was not working properly.

## Changes

- Use `cloudrun` argument instead of `run` in container launch cmd.

### DIFF
--- a/hexa/pipelines/management/commands/pipelines_runner.py
+++ b/hexa/pipelines/management/commands/pipelines_runner.py
@@ -207,7 +207,7 @@ def run_pipeline_kube(run: PipelineRun, env_var: dict):
 def run_pipeline_docker(run: PipelineRun, env_var: dict):
     from subprocess import PIPE, STDOUT, Popen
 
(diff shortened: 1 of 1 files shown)


hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -207,7 +207,7 @@ def run_pipeline_kube(run: PipelineRun, env_var: dict):
 def run_pipeline_docker(run: PipelineRun, env_var: dict):
     from subprocess import PIPE, STDOUT, Popen
 
-    docker_cmd = f'docker run --privileged -e HEXA_ENVIRONMENT=CLOUD_PIPELINE -e HEXA_RUN_ID={env_var["HEXA_RUN_ID"]} -e HEXA_SERVER_URL={env_var["HEXA_SERVER_URL"]} -e HEXA_TOKEN={env_var["HEXA_TOKEN"]} -e HEXA_WORKSPACE={env_var["HEXA_WORKSPACE"]} --network openhexa --platform linux/amd64 --rm {env_var["HEXA_PIPELINE_IMAGE"]} run'
```
qgerome, line 210: Why did you remove the `--rm` ?

pvanliefland:
> The run command is supposed to work with a mounted volume containing the pipeline and is used with the CLI of openhexa-sdk-python. "cloud run" is the one where we download the pipeline from the server and run it in a pod/container depending of the architecture. Since it is supposed to mimick what we'll have on the production servers, I think cloudrun is the correct command to run.

Ok, so I guess that's it's more a naming issue imho. I'll write that down somewhere else.

PR #608, opened 2023-12-15
feat(Pipelines): Implement a REST endpoint to run pipelines

qgerome:
2 content types are accepted : JSON & Form

## Changes

Add a simple endpoint on `/pipelines/run` that allows a user having a workspace token to trigger a pipeline run

## How/what to test

Make a few calls to run pipelines with parameters


hexa/pipelines/tests/test_views.py
pvanliefland, line 1: I'd suggest a test for the `application/x-www-form-urlencoded` content type

hexa/pipelines/urls.py
```diff
@@ -7,6 +7,7 @@
 app_name = "pipelines"
 
 urlpatterns = [
+    path("run", views.run_pipeline, name="run"),
```
pvanliefland, line 10: I think that the URL should contain the pipeline ID, and optionally the version.

It's easier for the webhook use case, especially for the use case that triggered the development of this feature (SMS -> Twilio number > webhook), that way the pipeline ID does not have to be sent in the SMS payload.

hexa/pipelines/views.py
```diff
@@ -87,10 +94,94 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+    elif content_type == "application/json":
+        try:
+            data = json.loads(request.body)
+        except json.JSONDecodeError:
+            return JsonResponse({"error": "Invalid JSON"}, status=400)
+    else:
+        return JsonResponse({"error": "Unsupported content type"}, status=400)
+
+    try:
+        print(request.POST, flush=True)
```
pvanliefland, line 132: I guess this should be removed

hexa/pipelines/views.py
```diff
@@ -87,10 +94,94 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+        pipeline = (
+            Pipeline.objects.filter_for_user(request.user)
+            .filter(workspace=request.workspace)
+            .get(id=data["id"])
```
pvanliefland, line 136: Pipeline id should be in URL

hexa/pipelines/views.py
```diff
@@ -87,10 +94,94 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+        )
+        version = pipeline.last_version
+        if "version" in data:
+            version = PipelineVersion.objects.get(
+                pipeline=pipeline, number=data["version"]
```
pvanliefland, line 141: Version number in URL as well

hexa/pipelines/views.py
```diff
@@ -87,10 +94,94 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+    except Pipeline.DoesNotExist:
+        return JsonResponse({"error": "Pipeline not found"}, status=404)
+    except PipelineVersion.DoesNotExist:
+        return JsonResponse({"error": "Pipeline version not found"}, status=404)
+
+    # Check if the user can run the pipeline
+    if not request.user.has_perm("pipelines.run_pipeline", pipeline):
+        return JsonResponse({"error": "Forbidden"}, status=403)
+
+    send_mail_notifications = data.get("send_mail_notifications", False)
```
pvanliefland, line 152: I would consider putting this as a query string option in the URL

hexa/pipelines/views.py
```diff
@@ -87,10 +94,94 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+    if content_type == "application/x-www-form-urlencoded":
+        for parameter in version.parameters:
+            if parameter["code"] not in data:
+                continue
+            if parameter["multiple"]:
+                config[parameter["code"]] = data.getlist(parameter["code"])
+            else:
+                config[parameter["code"]] = data.get(parameter["code"])
+    else:  # We have a JSON payload
+        config = data.get("config", {})
```
pvanliefland, line 165: If we move pipeline id and version to the URL, the `config` key is not necessary anymore and keys could be parameter codes.

pvanliefland (changes requested):
Nice, my only major remark is about the webhook URL.

hexa/pipelines/graphql/schema.graphql
```diff
@@ -137,6 +137,8 @@ type Pipeline {
   runs(page: Int, perPage: Int, orderBy: PipelineRunOrderBy): PipelineRunPage!
   versions(page: Int, perPage: Int): PipelineVersionPage!
   recipients: [PipelineRecipient!]!
+  hasPublicWebhook: Boolean!
```
pvanliefland, line 140: Small suggestion: `webhookEnabled`? Shouldn't be public for too long

hexa/pipelines/migrations/0032_pipeline_has_public_webhook.py
```diff
@@ -0,0 +1,18 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("pipelines", "0031_pipelinerun_timeout"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="pipeline",
+            name="has_public_webhook",
```
pvanliefland, line 15: Same idea here, `webhook_enabled`?

hexa/pipelines/urls.py
```diff
@@ -15,4 +15,10 @@
     ),
     path("credentials/", views.credentials, name="credentials"),
     path("status", views.pipelines_status, name="status"),
+    path(
```
pvanliefland, line 18: When I look at the rest of the code, I'm starting to think that this could be a query string parameter as well.

hexa/pipelines/views.py
```diff
@@ -87,10 +94,79 @@ def credentials(request: HttpRequest) -> HttpResponse:
...
+            return JsonResponse({"error": "Pipeline has no version"}, status=400)
+    except PipelineVersion.DoesNotExist:
+        return JsonResponse({"error": "Pipeline version not found"}, status=404)
+
+    # Get the data from the request
+    content_type = request.META.get("CONTENT_TYPE")
+    send_mail_notifications = False
+    config = {}
+    if content_type == "application/x-www-form-urlencoded":
+        send_mail_notifications = request.POST.get("send_mail_notifications", False)
```
pvanliefland, line 136: Shouldn't this be `request.GET.get()`?

pvanliefland (changes requested):
Nice, this looks good!

I think there is a small mistake regarding the email notification parameters (should use `request.GET` instead of `request.POST`.

My other comments are mostly cosmetic, feel free to take them into account or not 😉

hexa/pipelines/views.py
qgerome: Nope, We have to do it on GET to have the query params

pvanliefland (approved):
LGTM!

PR #611, opened 2023-12-20
feat(Pipelines): Improve webhook support

qgerome:
* Remove default support (leave this to the pipeline runner)
* Cast parameters to the correct type when using urlencoded form
* More tests
* Add a WEBHOOK trigger for the pipeline run


pvanliefland (approved):
LGTM!

PR #613, opened 2023-12-26
User can register & see their workspace's invitations

qgerome:
Review the flow to accept pending invitations for workspaces


hexa/user_management/schema.py
```diff
@@ -321,6 +326,55 @@ def resolve_logout(_, info, **kwargs):
...
 
+@identity_mutations.field("register")
+def resolve_register(_, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    mutation_input = kwargs["input"]
+
+    if request.user.is_authenticated:
+        return {"success": False, "errors": ["ALREADY_LOGGED_IN"]}
+
+    # Check that the invitation token is valid
```
pvanliefland, line 337: SO IIRC our plan is to require a workspace invitation token for registration for now, but let the user confirm that he wants to join the workspace after registering.

I think your implementation is fine but we should expand your comment to explain this approach.

pvanliefland (approved):
Nice 👍

PR #616, opened 2023-12-27
fix(Pipelines): Fix boolean -> bool for the pipeline webhook

qgerome:
The type of parameter used to convert string values to a real bool was not correct


pvanliefland (approved):
LGTM!

PR #617, opened 2023-12-27
Mestachs wip minio gcp

mestachs:
Try to allow to use s3/minio and still allow gcp 

## Todo

- [x] stop hardcoding credentials (discuss with some dev if the existing AWS settings can be reused) 
- [ ] the "hexa.files.tests.test_api" should test all api/exception 
    - [x] should implement correctly list_bucket_objects (paging is not working and path are not well handled)
    - [x] should implement and test
        - [x] generate_download_url
        - [x] get_bucket_object
        - [x] generate_upload_url
        - [x] delete_object
        - [x] load_bucket_sample_data
    - [ ] test the all thing
        - perhaps have 2 test per implementation (or at least parameterized one, the same tests run against the s3 gcp and mock implementation)
- [ ] the token return by s3 should be used in conjunction of the aws_... keys (see test_short_lived_downscoped_access_token)
    - [x] adapt pipelines to use "correctly" the get_short_lived_downscoped_access_token
       - see PR : https://github.com/BLSQ/openhexa-notebooks/pull/58
       - but we probably need to adapt the "newer" or have a variant of it for s3 : https://github.com/BLSQ/openhexa-docker-images/blob/main/images/base/Dockerfile
    - [x] adapt notebooks to use "correctly" the get_short_lived_downscoped_access_token
- [x] find a way to enforce the "api" contract (same signatures)
    - [x] probably stop exposing different functions but return the `Client`
    - [ ] harmonized the Exception (ex validation of bucket name are different) 
    - [ ] don't expose/leak underlying implementation : ex  
       - `google.api_core.exceptions.NotFound`
       - bucket.blob(...)
- [ ] Allow to use s3 also (for the moment some STS things don't pass the ARN)
- [x] find a way to share some code between the 2 implementations (ex ObjectsPage, exceptions,...)

for the mock in "tests" have a look at another strategy ?
- from botocore.stub import Stubber
- or http://docs.getmoto.org/en/latest/

## Changes

- moved the "existing" hexa/files/api.py to gcp.py, introduce a GCPClient
- introduced a s3.py with a S3Client
- changed the api.py to either export the GCPClient or the S3Client functions

## How/what to test

## Screenshots / screencast


docker-compose.yaml
```diff
@@ -128,6 +128,19 @@ services:
     depends_on:
       - db
 
+  minio:
+    image: docker.io/bitnami/minio:2022
```
mestachs, line 132: see if there's a official minio image

hexa/files/api.py
```diff
@@ -1,161 +1,90 @@
...
+from .gcp import GCPClient
+from .s3 import S3Client
 
-@dataclass
-class ObjectsPage:
-    items: typing.List[any]
-    has_next_page: bool
-    has_previous_page: bool
-    page_number: int
+default_mode = settings.WORKSPACE_STORAGE_ENGINE
```
pvanliefland, line 8: Small suggestion: move `settings.WORKSPACE_STORAGE_ENGINE` to `get_storage()`, I think that doing it in the module itself (vs in a function) could be a problem for tests that override Django settings (https://docs.djangoproject.com/en/5.0/topics/testing/tools/#django.test.SimpleTestCase.modify_settings).

If you set he value of `default_mode` outside a function, then the value would be set at import time and you won't be able to override it later.

hexa/files/api.py
```diff
@@ -1,161 +1,90 @@
...
-                                    "inRole:roles/storage.objectAdmin"
-                                ],
-                            }
-                        ]
-                    }
-                }
-            ),
-        },
+def create_bucket(bucket_name):
+    warn(
```
pvanliefland, line 22: Actually this is fine - why not let it as is without the warnings?

hexa/files/gcp.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def get_credentials():
+    decoded_creds = base64.b64decode(settings.GCS_SERVICE_ACCOUNT_KEY)
+    json_creds = json.loads(decoded_creds, strict=False)
+    return service_account.Credentials.from_service_account_info(json_creds)
+
+
+def _get_short_lived_downscoped_access_token(bucket_name):
+    token_lifetime = 3600
```
pvanliefland, line 28: TODO: handle read-only mode.

hexa/files/gcp.py
pvanliefland, line 1: Let's remove the private functions and move everything in the engine class

hexa/files/s3.py
pvanliefland, line 1: Let's remove the private functions and move everything in the engine class

config/settings.py
```diff
@@ -435,6 +435,7 @@ def sentry_tracer_sampler(sampling_context):
 # Filesystem configuration
 WORKSPACE_BUCKET_PREFIX = os.environ.get("WORKSPACE_BUCKET_PREFIX", "")
 WORKSPACE_BUCKET_REGION = os.environ.get("WORKSPACE_BUCKET_REGION")
+WORKSPACE_STORAGE_ENGINE = os.environ.get("WORKSPACE_STORAGE_ENGINE", "gcp")
```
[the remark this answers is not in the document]
pvanliefland: And the default should be `s3`(minio)

hexa/workspaces/views.py
```diff
@@ -96,13 +98,37 @@ def credentials(request: HttpRequest, workspace_slug: str = None) -> HttpRespons
...
-    token, _ = get_short_lived_downscoped_access_token(workspace.bucket_name)
-    env.update(
-        {
-            "WORKSPACE_BUCKET_NAME": workspace.bucket_name,
-            "GCS_TOKEN": token,
-        }
+    token, _expires_in, engine = get_short_lived_downscoped_access_token(
+        workspace.bucket_name
     )
+    if engine == "s3":
```
pvanliefland, line 104: This could be moved in a `get_environment_variables` method or something similar in the engine classes

pvanliefland (changes requested):
Nice! As discussed, a series of suggestions

PR #621, opened 2024-01-02
feat(Files): Filter files matching a substring

qgerome:
Allow users to filter the objects in a bucket by filtering on the name of the objects.

## How/what to test

Try with the frontend pull request or via the graphiql interface


pvanliefland (approved):
LGTM!

You could maybe already add / update the typing annotations for the API module but this can be merged as is.

PR #625, opened 2024-01-09
core: remove dashboard entry

cheikhgwane:
Remove dashboard entry link.


qgerome:
If you remove the entry in the list, you can remove the code of the page as well no ?

cheikhgwane:
> If you remove the entry in the list, you can remove the code of the page as well no ?

I think it was already deleted. (cf : https://github.com/BLSQ/openhexa-app/pull/373/files#diff-81a23207261ad31ef9a5a3cbcddd361f955d4ff437b06fc8d12816a23ec75e1b)

qgerome:
Indeed ! 👌

PR #631, opened 2024-01-17
feat(Datasets): link datasetVersion to pipelinRun

cheikhgwane:
Allow PipelineRunUser to link dataset_versions to a PipelineRun


hexa/datasets/models.py
```diff
@@ -144,6 +145,12 @@ def create_if_has_perm(
             description=description,
         )
 
+        if isinstance(principal, PipelineRunUser):
+            pipeline_run_version = PipelineRunDatasetVersion(
```
qgerome, line 149: `PipelineRunDatasetVersion.objects.create` would directly create the object instead of these two lines

hexa/datasets/models.py
```diff
@@ -282,3 +289,11 @@ def delete_if_has_perm(self, *, principal: User):
...
     class Meta:
         unique_together = ("dataset", "workspace")
+
+
+class PipelineRunDatasetVersion(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    pipeline_run = models.ForeignKey("pipelines.PipelineRun", on_delete=models.CASCADE)
+    dataset_version = models.ForeignKey(
+        "datasets.DatasetVersion", on_delete=models.CASCADE
+    )
```
qgerome, line 299: You could add a ORM relationship to make it easier to query them

hexa/pipelines/graphql/schema.graphql
```diff
@@ -85,6 +85,7 @@ type PipelineRun {
   code: String!
   sendMailNotifications: Boolean!
   timeout: Int
+  datasetVersion: [DatasetVersion!]!
```
qgerome, line 88: I'm not a fan of datasetVersion but I cannot think of another name... (Let's put it in plural)

hexa/pipelines/schema.py
```diff
@@ -310,6 +311,14 @@ def resolve_pipeline_run_outputs(run: PipelineRun, info, **kwargs):
     return result
 
 
+@pipeline_run_object.field("datasetVersion")
+def resolve_pipeline_run_dataset_version(run: PipelineRun, info, **kwargs):
+    return [
+        run_version.dataset_version
+        for run_version in PipelineRunDatasetVersion.objects.filter(pipeline_run=run)
```
qgerome, line 318: That's where it could be useful to have a relation between the two

qgerome (changes requested):

hexa/datasets/models.py
cheikhgwane: Like ?

qgerome:
We do not need a m2m relationship between DatasetVersion & PipelineRun since one PipelineRun can create a unique DatasetVersion

cheikhgwane:
> We do not need a m2m relationship between DatasetVersion & PipelineRun since one PipelineRun can create a unique DatasetVersion

I was thinking about the case where inside one run we try to create multiple versions for, let's say, different dataset

qgerome:
It works without it

DatasetY:DatasetVersion:1.run_id = run_x
DatasetY:DatasetVersion:2.run_id = run_x
DatasetZ:DatasetVersion:3.run_id = run_x

qgerome (approved):

PR #634, opened 2024-01-23
fix(PipelineRun): check if table/file_path exist

cheikhgwane:
Fix bugs that allowed user to specify a non-existent bucker file (path) or database name to be added as output.

## Changes

- Use get_bucket_object to check if file path exist
- Use get_table_definition to check if table_name exist


qgerome (commented):
Could you write a few tests for this please ?

qgerome (approved):

PR #636, opened 2024-01-26
fix(Notebooks): Logout the user from jupyterhub on logout

qgerome:
We call jupyterhub before logging out the user to also log him out of the hub.


pvanliefland (approved):
👍

PR #640, opened 2024-02-05
feat(Workspaces): do not add existing users to workspace by default

cheikhgwane:
Instead of auto-adding existing user to workspace, use the same process as the external user and let them manage invitations on account page.
See https://github.com/BLSQ/openhexa-team/issues/710#issue-2100169832


hexa/workspaces/schema/mutations.py
```diff
@@ -122,57 +122,61 @@ def resolve_invite_workspace_member(_, info, **kwargs):
...
             user = User.objects.get(email=input["userEmail"])
-            workspace_membership = WorkspaceMembership.objects.create_if_has_perm(
-                principal=request.user,
-                workspace=workspace,
-                user=user,
-                role=input["role"],
-            )
+            # If the user already exists, check if his not already a member of the workspace
+            if WorkspaceMembership.objects.filter(
+                user__email=input["userEmail"], workspace=workspace
```
qgerome, line 128: Since we found the user, let's use it to look for the membership
```suggestion
                user=user, workspace=workspace
```

hexa/workspaces/schema/mutations.py
```diff
@@ -122,57 +122,61 @@ def resolve_invite_workspace_member(_, info, **kwargs):
...
-                return {
-                    "success": True,
-                    "errors": [],
-                }
-            except IntegrityError:
-                return {
-                    "success": False,
-                    "errors": ["ALREADY_EXISTS"],
-                }
+            invitation = WorkspaceInvitation.objects.create_if_has_perm(
```
qgerome, line 164: This is exactly what you had above. It would be nice to refactor this mutation to simplify it.

`send_workspace_invitation_email` has to be used for user and non-user. You can do the check for the user in the body of the function (to set the correct language if possible)

hexa/workspaces/schema/mutations.py
```diff
@@ -294,7 +298,25 @@ def resolve_resend_workspace_invitation(_, info, **kwargs):
         invitation.updated_at = datetime.utcnow()
         invitation.save()
 
-        send_workspace_invitation_email(invitation)
+        query_set = User.objects.filter(email=invitation.email)
```
qgerome, line 301: This can be in `send_workspace_invitation_email` to not have two cases here. 
Plus, I'd prefer to have a `user = User.objects.filter(email=invitation.email).first()` than this, it's harder to read.

hexa/workspaces/templates/workspaces/mails/invite_member.html
```diff
@@ -1,11 +1,11 @@
 {% load i18n %}
 {% autoescape off %}
-{% blocktranslate %}You have been added to the workspace <b>{{workspace}}</b> by {{owner}}.{% endblocktranslate %}
+{% blocktranslate %}You have been invited to join the workspace <b>{{workspace}}</b> by {{owner}}.{% endblocktranslate %}
```
qgerome, line 3: IMO, the two mails can be merged (`invite_member` & `invite_external_user`). You could pass the user to the context of the mail to differentiate the two sentences.

qgerome (changes requested):

PR #653, opened 2024-02-13
chore(Pipelines): Remove the temporary status page

qgerome:
This page was only known by OH devs and it's not used.


cheikhgwane:
@qgerome failing tests.

cheikhgwane (approved):

PR #654, opened 2024-02-15
Activate the bucket versioning and set a lifecycle policy

qgerome:
As of now we had no way to recover from a file/folder deletion on a workspace bucket.
This pull request addresses that by enabling the versioning and setting a policy for the lifecycle of the files.

The policy will move progressively the files to a colder storage based on their last access date time.


cheikhgwane (approved):
LGTM!

PR #658, opened 2024-02-19
Fix docker image python

qgerome:
The python version used in the image was not the one we wanted to use...


cheikhgwane (approved):
Tested in local, works fine !

PR #663, opened 2024-02-20
feat(Pipelines): enable soft deletion of pipeline

cheikhgwane:
Enable soft delete of Pipeline.

## Changes

- Add a deleted field on pipeline.


cheikhgwane:
@qgerome  I've add the deleted field on Pipeline constraints list to allow user to be able to reuse a pipeline code.

qgerome:
Could you extract the logic and make it a mixin we could reuse on all the other models as needed ?
Check the libraries that already exist to have some examples/ideas.

hexa/core/models/soft_delete.py
```diff
@@ -0,0 +1,67 @@
...
+
+    """
+
+    class Meta:
+        abstract = True
+
+    deleted_at = models.DateTimeField(default=None, blank=True, null=True)
+    restored_at = models.DateTimeField(default=None, blank=True, null=True)
+
+    def delete(self):
```
qgerome, line 46: Could you add a is_deleted & is_restored ?

hexa/core/models/soft_delete.py
```diff
@@ -0,0 +1,67 @@
...
+from django.db import models
+from django.utils import timezone
+
+
+class DefaultSoftDeletedManager(models.Manager):
+    """
+    This manager will return all objects no marked as  soft deleted
+    """
+
+    def get_queryset(self):
```
qgerome, line 12: I would still create a `SoftDeleteQuerySet` to handle the bulk deletion (by default, it will delete the rows from DB and not soft delete them)

qgerome (changes requested):

qgerome (changes requested):

hexa/core/models/soft_delete.py
```diff
@@ -0,0 +1,93 @@
...
+    def delete(self):
+        """
+        Soft deletes the object.
+
+        """
+        self.deleted_at = timezone.now()
+        self.restored_at = None
+        self.save()
+
+    def hard_delete(self, using=None, keep_parents=False):
```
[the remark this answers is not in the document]
cheikhgwane: It was more to keep the same signature as the Model.delete method, but not used for now.

> Sometimes with [multi-table inheritance](https://docs.djangoproject.com/en/5.0/topics/db/models/#multi-table-inheritance) you may want to delete only a child model’s data. Specifying keep_parents=True will keep the parent model’s data.

qgerome:
Ok ! Thanks for the clarification!

hexa/core/tests/soft_delete/tests.py
```diff
@@ -0,0 +1,68 @@
...
+        with connection.schema_editor() as schema_editor:
+            schema_editor.delete_model(TestMyModel)
+
+        super().tearDownClass()
+
+    def test_soft_delete_model(self):
+        my_model = TestMyModel.objects.create()
+
+        assert my_model.is_deleted is False
+
```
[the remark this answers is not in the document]
qgerome: You didn't add this

qgerome:
I think it's a good idea to add a filter in the admin to filter out by default the deleted pipelines

hexa/pipelines/admin.py
```diff
@@ -19,15 +20,36 @@ def restore_pipelines(modeladmin, request, queryset):
         obj.restore()
 
 
+class DeletedPipelineFilter(admin.SimpleListFilter):
```
qgerome, line 23: Could you please take some inspiration from https://github.com/san4ezy/django_softdelete/blob/master/django_softdelete/admin.py & https://github.com/san4ezy/django_softdelete/blob/master/django_softdelete/filters.py to not have to write this piece of code for every SoftDeleteModel ?

qgerome (changes requested):

hexa/core/filters.py
```diff
@@ -0,0 +1,20 @@
...
+        return (
+            ("true", _("Soft deleted")),
+            ("false", _("Not Deleted")),
+        )
+
+    def queryset(self, request, queryset):
+        if not self.value():
+            return queryset
+
+        value = {"true": False, "false": True, "all": "all"}[self.value() or "false"]
```
qgerome, line 19: In which case is the `"all": "all"`  called ? What happens when you filter with deleted_at= "all" ?

qgerome (changes requested):

hexa/core/filters.py
cheikhgwane: I forgot to remove that entry, it should not be there.

PR #669, opened 2024-03-25
fix(Files): Do not log missing output files in Sentry

qgerome:
We haven't seen this NotFound error in Sentry since the refactoring of Stephan. I added a catch for this exception in the outputs resolver to ignore it.


cheikhgwane (approved):
LGTM!

PR #670, opened 2024-03-25
feat(Pipelines): allow user to stop a pipeline

cheikhgwane:
Allow user to stop pipeline run.

## Changes

- Add two new states : TERMINATING (indicates that this pipeline should be stopped) and STOPPED.
- Save user who stopped the run.


hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -225,11 +250,15 @@ def run_pipeline_docker(run: PipelineRun, image: str, env_vars: dict):
...
+        # we stop the running process when the run state is a terminating
+        if run.state == PipelineRunState.TERMINATING:
+            proc.kill()
+            break
 
         proc.poll()
         if proc.returncode is not None:
             break
-        sleep(5)
+        sleep(10)
```
qgerome, line 261: Why ?

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -281,7 +310,10 @@ def run_pipeline(run: PipelineRun):
     run.refresh_from_db()
     run.duration = timezone.now() - time_start
     run.run_logs = "\n".join([base_logs, container_logs])
-    if success:
+
+    if run.stopped_by is not None:
```
qgerome, line 314: run.state == PipelineRunState.Terminating

hexa/pipelines/models.py
```diff
@@ -404,6 +411,14 @@ def add_output(self, uri: str, output_type: str, name: typing.Optional[str]):
         )
         self.save()
 
+    def stop_if_has_perm(self, principal: User):
```
qgerome, line 414: it can simply be `def stop(self, principal: User)` and the check is made inside it. I think we only use the "if_has_perm" for basic CRUD operations

hexa/plugins/connector_airflow/graphql/schema.graphql
qgerome, line 1: I think you forgot to add "stoppedBy"

qgerome (changes requested):

hexa/pipelines/management/commands/pipelines_runner.py
cheikhgwane: For testing, will remote it.

hexa/pipelines/models.py
cheikhgwane: Copy that.

qgerome (approved):

PR #685, opened 2024-04-16
fix(Pipelines): enable nulls value check on Pipeline model constraint

cheikhgwane:
It was possible to have duplicate entry for pipeline with the same code. The constraint, composed of code, workspace and deleted_at (which is null by default) was not correctly implemented . This PR intends to fix that by enforcing null values check on Pipeline model constraint. (See https://docs.djangoproject.com/en/5.0/ref/models/constraints/#condition)

## Changes

- Set condition UniqueConstraint


qgerome:
The problem is that we are only in Postgres 12 :/

cheikhgwane:
> The problem is that we are only in Postgres 12 :/

Ouch 🫤. All right, let me look for another solution (maybe model clean method)

PR #688, opened 2024-04-19
Feat/notebooks pipelines

cheikhgwane:
Allow user to create a pipeline from a notebook.

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast


hexa/pipelines/models.py
```diff
@@ -195,6 +200,14 @@ class Meta:
     memory_limit = models.CharField(blank=True, max_length=32)
     recipients = models.ManyToManyField(User, through="PipelineRecipient")
 
+    type = models.CharField(
+        max_length=200,
+        blank=False,
+        choices=PipelineType.choices,
+        default=PipelineType.DEFAULT,
+    )
+    notebook = models.CharField(max_length=200, null=True, blank=True)
```
qgerome, line 209: `notebookPath` seems like a better idea. 200 chars seems too narrow for this. It's better to use `TextField`

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +53,33 @@ def resolve_create_pipeline(_, info, **kwargs):
...
+                name=input.get("name"),
+                workspace=workspace,
+                type=PipelineType.NOTEBOOK,
+                notebook=notebook,
+            )
+            # automatically create a new version
+            pipeline.upload_new_version(
+                user=request.user,
+                name=input.get("name"),
+                zipfile=base64.b64decode("".encode("ascii")),
```
qgerome, line 70: I don't see why you have to fill this since it's a notebook type

hexa/pipelines/schema/mutations.py
```diff
@@ -28,6 +29,15 @@ def get_bucket_object(bucket_name, file):
     return get_storage().get_bucket_object(bucket_name, file)
 
 
+def create_pipeline_object(code: str, name: str, workspace: Workspace, **kwargs):
```
qgerome, line 32: Why is there the this function here ?

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +53,33 @@ def resolve_create_pipeline(_, info, **kwargs):
...
+                user=request.user,
+                name=input.get("name"),
+                zipfile=base64.b64decode("".encode("ascii")),
+                parameters=[],
+            )
+        else:
+            pipeline = create_pipeline_object(
+                code=input["code"],
+                name=input.get("name"),
+                workspace=workspace,
```
qgerome, line 77: I would still add the type as well

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +53,33 @@ def resolve_create_pipeline(_, info, **kwargs):
...
 
     try:
-        pipeline = Pipeline.objects.create(
-            code=input["code"],
-            name=input.get("name"),
-            workspace=workspace,
-        )
+        notebook = input.get("notebook")
+        if notebook:
+            get_bucket_object(workspace.bucket_name, notebook)
```
qgerome, line 58: A comment explaining why you added this line might be a good idea

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -23,6 +23,30 @@ class PodTerminationReason(Enum):
     DeadlineExceeded = "DeadlineExceeded"
 
 
+def build_run_env(target, env_vars: dict):
```
qgerome, line 26: If you call this function in the two cases and you clearly have a if to handle those exact two cases, it's better to leave the code in the corresponding caller function

hexa/files/s3.py
```diff
@@ -276,12 +277,19 @@ def list_bucket_objects(
...
         pages = paginator.paginate(
             Bucket=bucket_name,
-            Delimiter="/",
+            Delimiter="" if ignore_delimiter else "/",
             Prefix=prefix,
         )
 
         def is_object_match_query(obj):
-            if ignore_hidden_files and obj["name"].startswith("."):
+            is_hidden = obj["name"].startswith(".") or any(
```
qgerome, line 285: This has to be tested

hexa/files/s3.py
```diff
@@ -276,12 +277,19 @@ def list_bucket_objects(
...
+                [
+                    p.startswith(".")
+                    for p in obj["key"].split("/")[: len(obj["key"]) - 1]
+                ]
+            )
+            if ignore_hidden_files and is_hidden:
+                return False
```
qgerome, line 292: There are two ``return False`

hexa/files/s3.py
qgerome: I think this is correct:
```python
is_hidden = obj["name"].startswith(".") or "/." in obj["key"]
```

hexa/pipelines/schema/mutations.py
cheikhgwane: Avoid duplication

hexa/pipelines/schema/mutations.py
cheikhgwane: The zipFile field was required in the PipelineVersion model, but yeah it's better to modify the binaryField and allow  it to be null. WDYT?

qgerome:
Yes, it’s better

hexa/files/gcp.py
```diff
@@ -248,7 +249,12 @@ def list_bucket_objects(
         objects = []
 
         def is_object_match_query(obj):
-            if ignore_hidden_files and obj["name"].startswith("."):
+            is_hidden = (
+                obj["name"].startswith(".")
```
qgerome, line 253: This one is covered by the last case no ? 
```
.hidden
a/b/c/.hidden
a/.b/c/hidden
.a/b/c/hidden
```

hexa/pipelines/graphql/schema.graphql
```diff
@@ -138,6 +138,11 @@ type PipelineRecipient {
   pipeline: Pipeline!
 }
 
+enum PipelineType {
+  DEFAULT
```
qgerome, line 142: You haven't changed the PipelineType to another value. Is it what you wanted to push ?

hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -236,7 +244,10 @@ def run_pipeline_kube(run: PipelineRun, image: str, env_vars: dict):
 def run_pipeline_docker(run: PipelineRun, image: str, env_vars: dict):
     from subprocess import PIPE, STDOUT, Popen
 
-    docker_cmd = f'docker run --privileged -e HEXA_ENVIRONMENT=CLOUD_PIPELINE -e HEXA_RUN_ID={env_vars["HEXA_RUN_ID"]} -e HEXA_SERVER_URL={env_vars["HEXA_SERVER_URL"]} -e HEXA_TOKEN={env_vars["HEXA_TOKEN"]} -e HEXA_WORKSPACE={env_vars["HEXA_WORKSPACE"]} --network openhexa --platform linux/amd64 --rm {image} pipeline cloudrun'
+    env_vars.pop("HEXA_PIPELINE_NAME", "")
```
qgerome, line 247: Why is it needed ?

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +44,35 @@ def resolve_create_pipeline(_, info, **kwargs):
         }
 
     try:
-        pipeline = Pipeline.objects.create(
-            code=input["code"],
-            name=input.get("name"),
-            workspace=workspace,
-        )
+        notebook = input.get("notebookPath")
```
qgerome, line 47: It's better to clearly state what is in your variable
```suggestion
        notebook_path = input.get("notebookPath")
```

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +44,35 @@ def resolve_create_pipeline(_, info, **kwargs):
...
+                name=input.get("name"),
+                workspace=workspace,
+                type=PipelineType.NOTEBOOK,
+                notebookPath=notebook,
+            )
+            # automatically create a new version
+            pipeline.upload_new_version(
+                user=request.user,
+                name=input.get("name"),
+                zipfile=base64.b64decode("".encode("ascii")),
```
qgerome, line 62: I thought you would remove this by setting the field `zipfile` as non mandatory ?

qgerome (changes requested):

hexa/pipelines/graphql/schema.graphql
cheikhgwane: It should be ZipFile.

hexa/pipelines/management/commands/pipelines_runner.py
cheikhgwane: I've seen that we really don't use it for the run_docker, but I can keep it (for future usage ?)

qgerome:
Better to keep it

hexa/files/gcp.py
cheikhgwane: Yes, I've added  it to handle hidden files in the root dir ( when ignore_delimiter=True and the file key`.file.ext`)

qgerome:
I think we are going to remove the code that allows the user to search for objects since it would required us to also implement this feature in `s3.py`. It's slow and might raise our costs.

hexa/pipelines/schema/mutations.py
```diff
@@ -43,11 +44,24 @@ def resolve_create_pipeline(_, info, **kwargs):
...
+        }
+        if input.get("notebookPath", None) is not None:
+            data["type"] = PipelineType.NOTEBOOK
+            data["notebook_path"] = input["notebookPath"]
+            # we need to check if the notebook path exist in the workspace bucket
+            get_bucket_object(workspace.bucket_name, data["notebook_path"])
+        else:
+            data["type"] = PipelineType.ZIPFILE
+
+        pipeline = Pipeline.objects.create(**data)
```
cheikhgwane, line 60: I think we need to associate the pipeline(type=Notebook) with a PipelineVersion @qgerome

PR #699, opened 2024-05-15
feat(Pipelines): validate parameter types upon PipelineVersion run

nazarfil:
## Changes

Enforces pipeline type arguments to contain only valid parameter types

## How/what to test

Try submitting pipeline version with invalid types


hexa/pipelines/graphql/schema.graphql
```diff
@@ -36,10 +36,24 @@ enum PipelineRunStatus {
   terminating
 }
 
+enum ParameterTypes {
```
qgerome, line 39: The other enums are singular, I think it's better to keep the singular to be consistant.

hexa/pipelines/graphql/schema.graphql
```diff
@@ -36,10 +36,24 @@ enum PipelineRunStatus {
...
+  s3
+  gcs
+  custom
+}
+
 type PipelineParameter {
   code: String!
   name: String!
-  type: String!
+  type: ParameterTypes
```
qgerome, line 56: `type` is required

qgerome (changes requested):
It would be nice to have tests with incorrect types.

qgerome:
Don't forget to edit the PR description to reflect your changes

nazarfil:
Valid points I have updated the pr

qgerome (approved):

PR #700, opened 2024-05-15
feat(Workspaces): allow user to update workspace image  and rename PIPELINE_IMAGE env

cheikhgwane:
This PR intends to allow users to edit the docker image used to run pipelines and jupyterlab for  the workspace.

### DIFF
--- a/hexa/workspaces/schema/mutations.py
+++ b/hexa/workspaces/schema/mutations.py
@@ -62,6 +62,9 @@ def resolve_update_workspace(_, info, **kwargs):
         if input.get("description", None):
             args["description"] = input["description"]
 
+        if input.get("dockerImage", None) is not None:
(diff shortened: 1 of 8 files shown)


hexa/workspaces/schema/mutations.py
```diff
@@ -62,6 +62,8 @@ def resolve_update_workspace(_, info, **kwargs):
         if input.get("description", None):
             args["description"] = input["description"]
 
+        args["docker_image"] = input.get("dockerImage", "")
```
qgerome, line 65: This is different than the other cases around this one where we check if there is an input given in order to avoid to override the dockerImage when it's not given explicitly by the user

qgerome (changes requested):

PR #704, opened 2024-05-16
feat(Workspaces): add feature flag to allow users the creation of wor…

nazarfil:
Changes the permission on who can create workspaces

In addition to superuser as it was allowed previously, we will allow users with feature flag workspaces.create to be able to create workspaces


qgerome (changes requested):
Could you write a test that only test the `workspace.create` feature flag ?

> User x can't create a workspace, I add the feature flag and now he can.

qgerome (approved):

PR #713, opened 2024-05-31
Feat pipeline version configration nf

nazarfil:

## Changes

Adds config field to pipeline [version](https://bluesquare.atlassian.net/browse/HEXA-938)
Allows pipeline to run in scheduled mode if parameter is in config


hexa/pipelines/models.py
```diff
@@ -262,6 +267,7 @@ def upload_new_version(
             description=description,
             external_link=external_link,
             zipfile=zipfile,
+            config=config if config else self.config,
```
qgerome, line 270: `Pipeline.config` is only used for the former system of pipelines using Airflow. For our implementation we can ignore that.

hexa/pipelines/schema/mutations.py
```diff
@@ -185,7 +185,7 @@ def resolve_run_pipeline(_, info, **kwargs):
             user=request.user,
             pipeline_version=version,
             trigger_mode=PipelineRunTrigger.MANUAL,
-            config=input.get("config", {}),
+            config=input.get("config", version.config),
```
qgerome, line 188: No need to use `version.config` as the default for the config. This logic has to be in the `run` method of the `PipelineVersion` in itself.

qgerome (changes requested):
Already a great start. There is a behaviour missing in this PR though: When a user launches a pipeline run, we have to take the config provided via the graphql and merge it with the configuration on the pipeline version before running it. 

Also, I don't think the scheduler has been modifier to take into consideration that it can takes the config on the `PipelineVersion` to run the pipeline.

hexa/pipelines/schema/mutations.py
nazarfil: Ok, changed it to merge both

hexa/pipelines/models.py
qgerome: Are you sure ? It's in the `upload_new_version` method that is on `Pipeline`

hexa/pipelines/models.py
```diff
@@ -323,6 +328,15 @@ def get_token(self):
             }
         )
 
+    def merge_pipeline_config(
+        self,
+        config: typing.Mapping[typing.Dict, typing.Any],
+        pipeline_version_config: typing.Mapping[typing.Dict, typing.Any],
+    ):
+        provided_config = config if config else self.config
+        merged_config = {**pipeline_version_config, **provided_config}
```
qgerome, line 337: I'm wondering if we should only takes the keys where their value is != than None?

hexa/pipelines/tests/test_schema/test_pipelines.py
```diff
@@ -1471,6 +1471,55 @@ def test_pipelines_permissions_schedule_required_param_with_no_default_value(sel
             r["data"]["pipelineByCode"],
         )
 
+    def test_pipelines_permissions_schedule_required_param_with_no_default_value_with_version_config(
```
qgerome, line 1474: Could you write more tests to check how the config is merged when you run a pipeline with a version's config?

qgerome (changes requested):

qgerome (changes requested):
You have to check the pipelines_scheduler.py file. This PR will have an effect on the `Pipeline.is_schedulable` (since you can change a version's config to remove values that may be required by a pipeline to run using a schedule)

As of now, the scheduler takes all the pipelines that have a schedule set. It should take into consideration only the ones that can be scheduled

nazarfil:
> You have to check the pipelines_scheduler.py file. This PR will have an effect on the `Pipeline.is_schedulable` (since you can change a version's config to remove values that may be required by a pipeline to run using a schedule)

Indeed, valid point. For the case where the pipeline was uploaded as schedulable, and if we remove a required parameter that was set previously in config - pipeline won't be schedulable anymore. I will add a check to forbid users uploading version config with required parameters that don't have default values and are removed.

qgerome:
> > You have to check the pipelines_scheduler.py file. This PR will have an effect on the `Pipeline.is_schedulable` (since you can change a version's config to remove values that may be required by a pipeline to run using a schedule)
> 
> Indeed, valid point. For the case where the pipeline was uploaded as schedulable, and if we remove a required parameter that was set previously in config - pipeline won't be schedulable anymore. I will add a check to forbid users uploading version config with required parameters that don't have default values and are removed.

I'm not sure to understand what you are going to do, are you proposing to prevent users to remove the values of the version's config when the pipeline is scheduled and these parameters are marked as required in the configuration ?

nazarfil:
> I'm not sure to understand what you are going to do, are you proposing to prevent users to remove the values of the version's config when the pipeline is scheduled and these parameters are marked as required in the configuration

Yes, preventing submission if a pipelineVersion that has config with a required parameter removed, when there is no default.

hexa/pipelines/models.py
```diff
@@ -131,16 +133,31 @@ def update_if_has_perm(self, principal: User, **kwargs):
         if not principal.has_perm("pipelines.update_pipeline_version", self):
             raise PermissionDenied
 
-        for key in ["name", "description", "external_link"]:
+        for key in ["name", "description", "external_link", "config"]:
             if key in kwargs and kwargs[key] is not None:
+                if key == "config" and self.is_schedulable:
```
qgerome, line 138: You can move that piece of code outside the for loop and treat the config directly.

qgerome: `is_schedulable` does not mean that the pipeline has a `schedule` ⚠️ It's better to check `pipeline.schedule`.

qgerome (changes requested):

PR #714, opened 2024-06-05
feat(Pipelines): generate webhook url

cheikhgwane:
The goal of this PR is to add an authentication layer to the webhook for pipelines. Users can now generate a new webhook url in case the previous one is compromised.

## Changes

- Add webhook_token attribute on Pipeline model
- Add mutation to generate a new webhook url and erase the previous one (no longer usable)
- Python migration to update pipelines webhook token


hexa/pipelines/models.py
```diff
@@ -288,6 +290,13 @@ def update_if_has_perm(self, principal: User, **kwargs):
             if key in kwargs:
                 setattr(self, key, kwargs[key])
 
+        if (
```
qgerome, line 293: That's not easy to read and is not optimal. In the case where you have disabled the webhook and enabled it again it wouldn't recreate a token.

What do you think of a `set_webhook_state(enabled: bool)` on the `Pipeline` ? You could call this method if the `webhook_enabled` is in the kwargs

hexa/pipelines/migrations/0043_generate_webhook_token.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 5.0.3 on 2024-06-04 20:59
+import base64
+
+from django.core.signing import TimestampSigner
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    def create_webhook_token(apps, schema_editor):
```
qgerome, line 9: Some people are using the webhooks already. How will you handle a smooth transition for them? I think in that case we have to keep the url as it is and warn them (export a list of the editor/admin members of the pipelines' workspaces)

hexa/pipelines/schema/mutations.py
```diff
@@ -213,6 +214,37 @@ def resolve_pipelineToken(_, info, **kwargs):
...
 
+@pipelines_mutations.field("generatePipelineWebhookUrl")
+def resolve_pipeline_generate_webhook_url(_, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    input = kwargs["input"]
+    try:
+        pipeline = Pipeline.objects.filter_for_user(request.user).get(
+            id=input.get("id")
+        )
+        if pipeline.webhook_enabled is False:
```
qgerome, line 225: This has to come after the check of permissions.

qgerome (changes requested):

hexa/pipelines/migrations/0043_generate_webhook_token.py
cheikhgwane: So return the keep old URL(/pipelines/uuid/run) and be able to use it alongside the new system and deprecated later?

qgerome: Yes or leave the url as it is and set the uuid as the webhook_token

hexa/pipelines/views.py
```diff
@@ -102,7 +102,7 @@ def run_pipeline(
         HttpResponse: Returns a dict with the `run_id` key containing the ID of the created run.
     """
     try:
-        pipeline = Pipeline.objects.get(id=id)
+        pipeline = Pipeline.objects.get(webhook_token=id)
```
qgerome, line 105: The type of the `id` argument is not the right one. It can be a string as well.

qgerome: I guess there is the same problem with the urls.py.

hexa/pipelines/models.py
```diff
@@ -199,6 +199,7 @@ class Meta:
     memory_request = models.CharField(blank=True, max_length=32)
     memory_limit = models.CharField(blank=True, max_length=32)
     recipients = models.ManyToManyField(User, through="PipelineRecipient")
+    webhook_token = models.UUIDField(null=True, blank=True)
```
qgerome, line 202: Your initial solution was good. You have to use a TimestampSigner for this BUT you still have to support the case where it's an UUID during the migration phase (that can last 2 weeks)

hexa/pipelines/views.py
qgerome: This is not what I had in ming. You have to support the two solutions to run pipelines, with an uuid and a generated signed token. We have to give time to people to switch to the new webhook token. The implementation in this file would be to try to validate the token and if it's not correct, try to get the pipeline with the id

qgerome (changes requested):

hexa/pipelines/views.py
cheikhgwane: Okay

hexa/pipelines/schema/mutations.py
```diff
@@ -213,6 +213,34 @@ def resolve_pipelineToken(_, info, **kwargs):
...
+    request: HttpRequest = info.context["request"]
+    input = kwargs["input"]
+    try:
+        pipeline = Pipeline.objects.filter_for_user(request.user).get(
+            id=input.get("id")
+        )
+        if not request.user.has_perm("pipelines.update_pipeline", pipeline):
+            raise PermissionDenied
+
+        if pipeline.webhook_enabled is False:
```
qgerome, line 227: Isn't there another check needed on the feature flag "web hooks" of the user ?

hexa/pipelines/views.py
```diff
@@ -102,9 +105,20 @@ def run_pipeline(
         HttpResponse: Returns a dict with the `run_id` key containing the ID of the created run.
     """
     try:
-        pipeline = Pipeline.objects.get(id=id)
+        # first we try to retrieve by id in case the pipeline webhook_url has not been migrated
```
qgerome, line 108: Could you add a #TODO to explain that we have to remove this after the next production release and once our users have been warned.

qgerome (changes requested):

PR #717, opened 2024-06-17
feat(Dataset): generate upload url independetly

nazarfil:
move generation of file uploadUrl to a separate mutation
deprecate uploadUrl in DatasetVersionFile

### DIFF
--- a/hexa/datasets/schema/mutations.py
+++ b/hexa/datasets/schema/mutations.py
@@ -193,6 +193,7 @@ def resolve_generate_upload_url(_, info, **kwargs):
         if get_blob(full_uri) is not None:
             return {"success": False, "errors": ["ALREADY_EXISTS"]}
         upload_url = generate_upload_url(full_uri, mutation_input["contentType"])
+
         return {"success": True, "errors": [], "upload_url": upload_url}
     except ValidationError:
         return {"success": False, "errors": ["INVALID_URI"]}
--- a/hexa/datasets/schema/mutations.py
+++ b/hexa/datasets/schema/mutations.py
@@ -5,7 +5,6 @@ from django.db import IntegrityError, transaction
 from hexa.pipelines.authentication import PipelineRunUser
 from hexa.workspaces.models import Workspace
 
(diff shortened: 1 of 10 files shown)


hexa/datasets/schema/mutations.py
```diff
@@ -178,6 +178,33 @@ def resolve_delete_dataset_share(_, info, **kwargs):
         return {"success": False, "errors": ["PERMISSION_DENIED"]}
 
 
+@mutations.field("generateUploadUrl")
+def resolve_generate_upload_url(_, info, **kwargs):
```
qgerome, line 182: I'm pretty sure there is a permission that need to be checked here ("update" or something like that)

qgerome (changes requested):
I think the part where you create the DatasetVersionFile after the upload is missing (in the tests and in the mutation resolver.

hexa/datasets/schema/types.py
```diff
@@ -189,6 +194,16 @@ def resolve_version_permissions_delete(obj: DatasetVersion, info, **kwargs):
     )
 
 
+@dataset_version_file_result_object.field("uploadUrl")
+def resolve_upload_url(obj, info, **kwargs):
+    try:
+        upload_url = generate_upload_url(obj.uri, obj.content_type)
+        return upload_url
+    except BucketObjectAlreadyExists as exc:
```
qgerome, line 202: Nice

hexa/datasets/schema/mutations.py
qgerome: @nazarfil Did you address this ?

hexa/datasets/schema/types.py
```diff
@@ -6,13 +8,16 @@
 from hexa.workspaces.models import Workspace
 from hexa.workspaces.schema.types import workspace_object, workspace_permissions
 
+from ...files.basefs import BucketObjectAlreadyExists
```
qgerome, line 11: Better to write `hexa.files.basefs`

nazarfil: True!

PR #719, opened 2024-06-18
feat(Dataset): add initial background queue, task, work and trigger t…

nazarfil:
Creates initial background queue set-up


docker-compose.yaml
```diff
@@ -89,6 +89,15 @@ services:
     depends_on:
       - db
 
+  dataset_file_metadata:
```
qgerome, line 92: `datasets_worker` may be better since we are going to do more things than just metadata extraction

qgerome (changes requested):

docker-compose.yaml
nazarfil: Yes, it is a better name

hexa/datasets/schema/mutations.py
```diff
@@ -252,11 +253,19 @@ def resolve_create_version_file(_, info, **kwargs):
                     uri=version.get_full_uri(mutation_input["uri"]),
                     content_type=mutation_input["contentType"],
                 )
-                return {
-                    "success": True,
-                    "errors": [],
-                    "file": file,
+            dataset_file_metadata_queue.enqueue(
```
qgerome, line 256: I think we should add a `load_metadata` on the DatasetVersionFile and call that on the save() method if there are no metadata attached.

qgerome (changes requested):

docker-compose.yaml
```diff
@@ -89,6 +89,15 @@ services:
     depends_on:
       - db
 
+  dataset_worker:
+    <<: *common
+    command: "manage dataset_worker"
+    restart: unless-stopped
+    profiles:
```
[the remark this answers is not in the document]
nazarfil: Added it to  the task PR, will move it here

hexa/datasets/schema/mutations.py
nazarfil: So on creation of DatasetVersionFile you suggest to change the flow to trigger generation of metadata file on save:
- DatasetVersionFile.save() -> load_metadata() -> enqueue(task)
Where load_metadata will enqueue the sample generation task, which does all the job to download, link with FK to DatasetVersionFile and save in db the sample

qgerome:
Yes

qgerome (approved):

PR #721, opened 2024-06-24
fix(Pipelines): handle case of notebook pipeline

cheikhgwane:
Notebook pipeline doesn't have version and thus we want to run a notebook pipeline it will fail.


hexa/pipelines/models.py
```diff
@@ -256,7 +256,11 @@ def run(
             trigger_mode=trigger_mode,
             execution_date=timezone.now(),
             state=PipelineRunState.QUEUED,
-            config=self.merge_pipeline_config(config, pipeline_version.config),
+            config=(
```
cheikhgwane, line 259: @nazarfil  can you check this please?

hexa/pipelines/views.py
```diff
@@ -126,14 +126,15 @@ def run_pipeline(
...
+    pipeline_version = None
     try:
         pipeline_version = pipeline.last_version
         if version_id is not None:
             pipeline_version = PipelineVersion.objects.get(
                 pipeline=pipeline, id=version_id
             )
 
-        if pipeline_version is None:
+        if pipeline.type == PipelineType.ZIPFILE and pipeline_version is None:
```
qgerome, line 137: It's weird to let the lines above run if you know that it does not going to work. The logic can be improved.

qgerome (changes requested):

qgerome (changes requested):
I've noticed there are no tests for this. Could you add tests for the webhook run of Notebook pipelines please ?

qgerome (approved):

PR #724, opened 2024-06-25
fix(Datasets): use dataset last version to track update

cheikhgwane:
Last updated should match the date of the version upload, [see](https://bluesquare.atlassian.net/jira/software/c/projects/HEXA/boards/330?assignee=712020%3A1d0055ff-af59-47b7-b48d-4d37e2a0d0b9&selectedIssue=HEXA-995). 

## Changes

- Return dataset last_version updated_at field if it exists.

### DIFF
--- a/hexa/datasets/schema/types.py
+++ b/hexa/datasets/schema/types.py
@@ -101,6 +101,15 @@ def resolve_dataset_permissions(obj: Dataset, info, **kwargs):
     return obj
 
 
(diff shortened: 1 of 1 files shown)


cheikhgwane:
> I like the idea but we have to be more concise than that. What if the last version is older than a change of the dataset's name or description ?

I htink we should take the most recent between dataset.update_at and dataversion.update_at.

qgerome (approved):

PR #761, opened 2024-07-03
feat/analytics :  OpenHEXA v1 analytics

cheikhgwane:
Draft PR for mixpanel integration.
## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast


qgerome (commented):
* `mixpanel` and not `mix_panel`
* no need of a get_mix_panel_client
* mixpanel.py -> analytics.py
* def create_event -> def track
* The check on the user has to be made in the track() and not before calling it obviously!
* The `distinct_id` is one of the most important property since it's the user identifier for Mixpanel
* You need to fill some of the properties directly in the track()
* I would not use a feature flag but a bool on the user to mark the opt-out
* Using a BufferedConsumer looks like a better idea as well

As a general note, do not hesitate to go deeper in the documentation or in the lib code.

qgerome:
You'd have to check this as well https://docs.mixpanel.com/docs/tracking-best-practices/server-side-best-practices

cheikhgwane:
@qgerome can you check for the default buffer size and also is it possible to have the app_version when we do a release ?

qgerome:
Could you please change the name of the PR please?

config/settings/base.py
```diff
@@ -395,3 +395,7 @@
 AWS_PERMISSIONS_BOUNDARY_POLICY_ARN = os.environ.get(
     "AWS_PERMISSIONS_BOUNDARY_POLICY_ARN", ""
 )
+
+# MIX_PANEL
+MIX_PANEL_TOKEN = os.environ.get("MIX_PANEL_TOKEN")
+MIX_PANEL_BUFFER_SIZE = os.environ.get("MIX_PANEL_BUFFER_SIZE", 50)
```
qgerome, line 401: You still haven't renamed the variables to `MIXPANEL`.
No need to be able to configure the buffer size. The default is good and I don't see cases where we could want to have different values depending on the instance.

hexa/core/analytics.py
```diff
@@ -0,0 +1,66 @@
+from datetime import datetime
+from uuid import uuid4
+
+from django.http import HttpRequest
+from mixpanel import BufferedConsumer, Mixpanel, MixpanelException
+from sentry_sdk import capture_exception
+from ua_parser import user_agent_parser
+
+from config.settings.base import MIX_PANEL_BUFFER_SIZE, MIX_PANEL_TOKEN
```
qgerome, line 9: Nope.

`from django.conf import settings`

hexa/core/analytics.py
```diff
@@ -0,0 +1,66 @@
...
+
+mixpanel_consumer = BufferedConsumer(max_size=MIX_PANEL_BUFFER_SIZE)
+mixpanel = Mixpanel(token=MIX_PANEL_TOKEN, consumer=mixpanel_consumer)
```
qgerome, line 12: Are you sure the mixpanel_token will always be there?

hexa/core/analytics.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def track(
```
qgerome, line 15: What do you do if the analytics are not enabled for the instance?

hexa/core/analytics.py
cheikhgwane: It's the auto-import, sorry!

hexa/core/analytics.py
cheikhgwane: Nope, well handle this.

hexa/core/analytics.py
cheikhgwane: I'm rewriting this part to thrown an exception or return if the token is not available

PR #764, opened 2024-07-08
feat : add mixpanel tracking

cheikhgwane:
Track events and users action using mixpanel.


hexa/datasets/tests/test_schema.py
```diff
@@ -328,7 +330,8 @@ class DatasetVersionTest(GraphQLTestCase, DatasetTestMixin):
     def setUpTestData(cls):
         get_storage().create_bucket(settings.WORKSPACE_DATASETS_BUCKET)
 
-    def test_create_dataset_version(self):
+    @mock.patch("hexa.datasets.schema.mutations.track")
```
qgerome, line 333: I think you'll have to patch the analytics module globally.

hexa/user_management/migrations/0020_user_analytics_enabled.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("user_management", "0019_user_language"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="analytics_enabled",
+            field=models.BooleanField(default=False),
```
qgerome, line 15: The default is TRUE

hexa/user_management/schema.py
```diff
@@ -314,6 +315,7 @@ def resolve_login(_, info, **kwargs):
         user_candidate.otp_device = device
 
     login(request, user_candidate)
+    track(user=user_candidate, event="user_login", request=request)
```
qgerome, line 318: Pretty sure there is a better way to track that (like `user_logged_in`)

qgerome (changes requested):
It's not going to be ok to patch all tests because some point in the loop we are going to track an event. It does not feel natural does it ?

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
+from datetime import datetime
+
+from django.conf import settings
+from django.http import HttpRequest
+from mixpanel import BufferedConsumer, Mixpanel
+from sentry_sdk import capture_exception
+from ua_parser import user_agent_parser
+
+mixpanel_consumer = BufferedConsumer()
+
```
qgerome, line 10:
```suggestion
mixpanel = None

if settings.MIXPANEL_TOKEN:
  mixpanel = Mixpanel()....
```

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
...
+
+def track(
+    user,
```
qgerome, line 13: Request is more important than user (you can get the user from the request if needed)

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
...
+    event: str,
+    properties: dict = {},
+    request: HttpRequest = None,
```
qgerome, line 16: Your typing is not correct. You typed the request to be a HttpRequest to which you set the default value to None and not a HttpRequest instance 🤷‍♂️

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
...
+    if isinstance(user, PipelineRunUser):
+        tracked_user = user.pipeline_run.user
+
+    if tracked_user is None or tracked_user.analytics_enabled:
+        try:
+            mixpanel = Mixpanel(
+                token=settings.MIXPANEL_TOKEN, consumer=mixpanel_consumer
+            )
+            # for request coming from openHEXA pipelines we don't need to store those info
+            if not isinstance(user, PipelineRunUser) and request:
```
qgerome, line 47: What we care is if we have a request to fill the metadata

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
...
+                    {
+                        "$browser": parsed["user_agent"]["family"],
+                        "$device": parsed["device"]["family"],
+                        "$os": parsed["os"]["family"],
+                        "ip": request.META["REMOTE_ADDR"],
+                    }
+                )
+            properties.update(
+                {
+                    "timestamp": datetime.now().timestamp(),
```
qgerome, line 59: Why do you need the timestamp? Isn't it already included in mixpanel ?

hexa/core/analytics.py
```diff
@@ -0,0 +1,87 @@
...
+                }
+            )
+
+            id = str(tracked_user.id) if tracked_user else None
+            mixpanel.track(
+                distinct_id=id,
+                event_name=event,
+                properties=properties,
+            )
+            if tracked_user:
```
qgerome, line 70: I think this could be in a post auth signal/middleware/authentication backend instead of being in the track()

hexa/files/schema/mutations.py
```diff
@@ -38,6 +39,12 @@ def resolve_prepare_download_object(_, info, **kwargs):
         download_url = get_storage().generate_download_url(
             workspace.bucket_name, object_key, force_attachment=True
         )
+        track(
+            request.user,
+            "file_downloaded",
```
qgerome, line 44: I'd prefer something that represents the truth such as : `files.generate_download_url`

qgerome (changes requested):

hexa/core/analytics.py
```diff
@@ -0,0 +1,84 @@
...
+    ----------
+    request : HttpRequest
+    event : str
+        An identifier for the event to track.
+    properties : dict
+       A dictionary holding the event properties
+    """
+    from hexa.pipelines.authentication import PipelineRunUser
+
+    mixpanel = None
```
qgerome, line 29: Bon. This is outside of the track function

qgerome (changes requested):

hexa/core/analytics.py
```diff
@@ -0,0 +1,84 @@
...
+            mixpanel.track(
+                distinct_id=str(tracked_user.id) if tracked_user else None,
+                event_name=event,
+                properties=properties,
+            )
+        except Exception as e:
+            capture_exception(e)
+
+
+def track_user(user: User):
```
qgerome, line 64: `set_user_properties` makes more sense no ?

hexa/core/analytics.py
```diff
@@ -0,0 +1,84 @@
...
+            mixpanel = Mixpanel(
+                token=settings.MIXPANEL_TOKEN, consumer=mixpanel_consumer
+            )
+            mixpanel.people_set_once(
+                distinct_id=str(user.id),
+                properties={
+                    "$email": user.email,
+                    "$name": user.display_name,
+                    "staff_status": user.is_staff,
+                    "superuser_status": user.is_superuser,
```
qgerome, line 76: is_superuser,
is_staff.
Why the renaming?

qgerome (changes requested):

hexa/core/analytics.py
cheikhgwane: Oh it was to respect the naming on the excel file, but I"ve changed it according to your comments.

hexa/core/analytics.py
cheikhgwane: Yes, updated.

qgerome (changes requested):

qgerome:
Could you change the README.md to include this please ?

qgerome (commented):
I've changed the code myself. 

Could you adapt the readme please?

hexa/core/analytics.py
```diff
@@ -0,0 +1,77 @@
+from django.conf import settings
+from django.http import HttpRequest
+from mixpanel import BufferedConsumer, Mixpanel
+from sentry_sdk import capture_exception
+from ua_parser import user_agent_parser
+
+from hexa.user_management.models import AnonymousUser, User
+
+mixpanel_consumer = BufferedConsumer(max_size=1)
```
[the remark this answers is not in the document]
cheikhgwane: Oh my bad, it was during testing, forgot to remove it thx.

README.md
```diff
@@ -145,6 +145,9 @@ As an example, use the following command to run the migrations:
 docker compose run app migrate
 ```
 
+### Analytics
+We use [Mixpanel](https://mixpanel.com/home/) to track users and their actions. If you want to enable it, set the `MIXPANEL_TOKEN` environment variable with the token from your Mixpanel project and restart the application.
```
cheikhgwane, line 149: @qgerome

qgerome:
Ok my bad I didn’t see it in the review panel

PR #773, opened 2024-07-19
feat(Analytics): add an analytics module

cheikhgwane:
Add an analytics module.

## Changes

- Move the track function from core to analytics module.
- Add a track endpoint for creating event from the frontend.


qgerome (changes requested):

hexa/analytics/views.py
```diff
@@ -0,0 +1,30 @@
...
+@require_POST
+@csrf_exempt
+def track_event(request: HttpRequest) -> HttpResponse:
+    """This API endpoint is called by the frontend to track events."""
+    payload = json.loads(request.body.decode("utf-8"))
+
+    if "event" not in payload:
+        return JsonResponse({"Bad request": "event name is required."}, status=400)
+
+    if isinstance(request.user, AnonymousUser) or request.user.analytics_enabled:
```
qgerome, line 21: I think it would make sense to check this condition before loading the json.

qgerome (changes requested):

hexa/analytics/views.py
qgerome: Well, this condition is already checked in the track() in fact...

cheikhgwane: Yes,  I thought it would be better to let the user knows the reason why. WDYT?

PR #781, opened 2024-08-08
Metadata  calculation

nazarfil:
Adds profiling to metadata

Example

```
data: [{'column_names': 'name', 'data_types': 'string', 'missing_values': 0, 'unique_values': 2, 'distinct_values': 2, 'constant_values': False}, {'column_names': 'surname', 'data_types': 'string', 'missing_values': 0, 'unique_values': 2, 'distinct_values': 2, 'constant_values': False}]
pandata:   column_names data_types  missing_values  unique_values  distinct_values  constant_values
0         name     string               0              2                2            False
1      surname     string               0              2                2            False
```


hexa/datasets/graphql/schema.graphql
```diff
@@ -104,8 +104,10 @@ enum FileMetadataStatus{
 Metadata for dataset file
 """
 type DatasetFileMetadata {
-    sample: JSON!
+    sample: JSON
+    profiling: JSON
     status: FileMetadataStatus!
+    status_reason: String
```
qgerome, line 110: In graphl, it's camelCased.

hexa/datasets/models.py
```diff
@@ -271,7 +271,7 @@ class DatasetFileMetadata(Base):
         (STATUS_FAILED, _("Failed")),
         (STATUS_FINISHED, _("Finished")),
     ]
-
+    profiling = JSONField(default=dict, null=True, blank=True)
```
qgerome, line 274: I think it would make sense to have None as the default value

hexa/datasets/queue.py
```diff
@@ -39,7 +39,7 @@ def download_file_as_dataframe(
     )
     download_url = generate_download_url(dataset_version_file)
     if mime_type == "text/csv":
-        return pd.read_csv(download_url)
+        return pd.read_csv(download_url, low_memory=False)
```
qgerome, line 42: What does that mean ? Maybe add a comment explaining why we set it as low_memory no ?

hexa/datasets/queue.py
```diff
@@ -52,6 +52,33 @@ def download_file_as_dataframe(
         return pd.read_parquet(download_url)
 
 
+def metadata_profiling(file_content: pd.DataFrame) -> list:
```
qgerome, line 55:
```suggestion
def metadata_profiling(dataframe: pd.DataFrame) -> list:
```

qgerome: It's not related to this function that it's a file_content

hexa/datasets/queue.py
```diff
@@ -52,6 +52,33 @@ def download_file_as_dataframe(
...
+    unique_values = file_content.nunique().to_dict()
+    distinct_values = file_content.apply(lambda x: x.nunique(dropna=False)).to_dict()
+    constant_values = (
+        file_content.apply(lambda x: x.nunique() == 1).astype("bool").to_dict()
+    )
+
+    metadata_per_column = [
+        {
+            "column_names": column,
+            "data_types": data_types.get(column, "-"),
```
qgerome, line 70: On the backend, having `"-"` as the representation of a None value is not optimal. Better to leave that to None.

hexa/datasets/queue.py
```diff
@@ -52,6 +52,33 @@ def download_file_as_dataframe(
         return pd.read_parquet(download_url)
 
 
+def metadata_profiling(file_content: pd.DataFrame) -> list:
+    for col in file_content.select_dtypes(include=["object"]).columns:
+        file_content[col] = file_content[col].astype("string")
+
+    data_types = file_content.dtypes.apply(str).to_dict()
```
qgerome, line 59: In the case of a CSV, will it be able to return a float / int / bool type based on the values found in the column ?

qgerome:
```suggestion
    data_type = file_content.dtypes.apply(str).to_dict()
```

hexa/datasets/queue.py
qgerome, line 70:
```suggestion
            "data_type": data_type.get(column, "-"),
```

hexa/datasets/queue.py
```diff
@@ -90,6 +117,9 @@ def generate_dataset_file_sample_task(
                 replace=True,
             )
             dataset_file_metadata.sample = file_sample.to_json(orient="records")
+            dataset_file_metadata.profiling = json.dumps(
```
qgerome, line 120: Not sure you need to dumps the json since you defined the column as a JSON field. I would hope they do that for us.

qgerome (changes requested):

hexa/datasets/queue.py
nazarfil: Yes it is detecting:  float, int, bool and string.

hexa/datasets/queue.py
nazarfil: Still need it, tested in tests

PR #790, opened 2024-08-22
fix(Workspace): adds pipeline user check to fetch pipelines

nazarfil:
workaround for PipelineUser permissions to fetch workspace

### DIFF
--- a/hexa/workspaces/schema/queries.py
+++ b/hexa/workspaces/schema/queries.py
@@ -2,6 +2,7 @@ from ariadne import QueryType, convert_kwargs_to_snake_case
 
 from hexa.core.graphql import result_page
 
+from ...pipelines.authentication import PipelineRunUser
 from ..models import (
     Connection,
     Workspace,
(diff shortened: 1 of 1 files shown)


hexa/workspaces/schema/queries.py
```diff
@@ -25,7 +26,13 @@ def resolve_workspaces(_, info, query=None, page=1, perPage=15):
 def resolve_workspace(_, info, **kwargs):
     request = info.context["request"]
     try:
-        return Workspace.objects.filter_for_user(request.user).get(slug=kwargs["slug"])
+        # TODO: add pipeline user permission check
```
qgerome, line 29: Why is this needed since we already check this in the `filter_for_user` method of the Queryset ?

qgerome: I was referencing the body of your fix that is already implemented in Workspace.filter_for_user.

nazarfil: It was not working.
Similar to what was added to the workspace `workspace.filter_for_user` for `createDataset` mutation. 
If it is expected from  the check inside `filter_for_user` to perform this for PipelineUser, then I think it is worth searching for the reason why it is not working in here

PR #793, opened 2024-08-22
fix(Analytics): add support for PipelineRunUser and Ano

cheikhgwane:
The set_user_properties didn't check if User is defined before checking if the analytics are enabled, and  PipelineRunUser was not taken into account.

## Changes

- Retrieve the concrete user on analytics_middleware if user is an instance of PipelineRunUser model.
-  Check if user is defined on set_user_properties.

### DIFF
--- a/hexa/analytics/api.py
+++ b/hexa/analytics/api.py
@@ -62,7 +62,7 @@ def track(
 
 
 def set_user_properties(user: User):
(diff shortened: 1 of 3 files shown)


qgerome:
You should not have to check it it's a PipelineRunUser. That's what I already asked you to remove when you implemented the analytics.

PR #800, opened 2024-08-26
 feat(Metadata): add generic metadata and link with dataset, version, files

nazarfil:
Add dataset module
  Models for generic metadata  and attributes

Calculation of system related metadata attributes 

GraphQL API to fetch the metadata attributes, add custom attribute, delete custom attribute

*wip*:
- [x] copy attributes fro existing to the new version file 
- [x] add permissions check
- [x] fix tests


hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+
+extend type Dataset {
+  metadata: GenericMetadata
+}
+
+"""
+Input to add or edit a custom attribute
+"""
+input MetadataAttributeInput {
+  id: ID!
```
qgerome, line 35: Not clear what `id` represents here.

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+  key: String!
+  value: String
```
qgerome, line 37: This is in contradiction with your schema above where the value is mandatory

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+}
+
+"""
+Input to delete custom attribute
+"""
+input DeleteMetadataAttributeInput {
+  id: ID!
```
qgerome, line 44: what does it represent ?

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+  key: String!
+}
+
+"""
+Errors that can occur when creating an attribute.
+"""
+enum MetadataCreateAttributeError {
+  DATASET_VERSION_FILE_NOT_FOUND
```
qgerome, line 52: This should not be a type of error defined in this package.

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+  METADATA_OBJECT_NOT_FOUND
+  DUPLICATE_KEY
+}
+
+"""
+Errors that can occur when deleting an attribute.
+"""
+enum MetadataDeleteAttributeError {
+  DATASET_VERSION_FILE_NOT_FOUND
```
qgerome, line 61: Same here

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+  METADATA_OBJECT_NOT_FOUND
+  METADATA_ATTRIBUTE_NOT_FOUND
+}
+
+"""
+Errors that can occur when editing an attribute.
+"""
+enum MetadataEditAttributeError {
+  DATASET_VERSION_FILE_NOT_FOUND
```
qgerome, line 70: Same here

hexa/metadata/models.py
```diff
@@ -0,0 +1,32 @@
...
+class MetadataObject(Base):
+    content_type = models.ForeignKey(
+        ContentType, null=True, blank=True, on_delete=models.CASCADE
+    )
+    object_id = models.UUIDField(default=uuid.uuid4, editable=False)
+    data_object = GenericForeignKey("content_type", "object_id")
+
+
+class MetadataAttribute(Base):
+    key = models.CharField(max_length=255, unique=True)
```
qgerome, line 19: It means that if you have multiple files that have a column named "n_people" the profiling keys will throw an error since they won't be unique. The _uniqueness_ of the key is for (MetadataObject, key)

qgerome (changes requested):
Good first iteration ! I'd avoid having datasets references in the metadata module and make the metadata module the more generic possible in terms of features.

The metadata module should be tested.
The tests in the datasets are too complex and not sufficient. Try to limit the scope of what you're trying to test to make it easier for other devs to understand what you are testing.

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
...
+}
+
+type EditAttributeResult {
+  success: Boolean!
+  errors: [MetadataEditAttributeError!]!
+}
+
+extend type Mutation {
+  "Add a custom attribute to dataset file version metadata attribute"
+  addDatasetFileMetadata(input: MetadataAttributeInput!): CreateAttributeResult! @loginRequired
```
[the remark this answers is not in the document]
nazarfil: Ok I can move all the Object related methods to object modules.
Valid remark, I will look into how to make it more generic!

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,109 @@
+"""
+Generic metadata attribute
+"""
+type MetadataAttribute {
+  id: ID!
+  key: String!
+  value: String!
```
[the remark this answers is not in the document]
qgerome: In the past I used the `value<Type>` approach and it added a lot of code on the client-side. So I clearly prefer the JSON approach (and yes it's a shame to lose the type safety mechanism :()

hexa/metadata/models.py
```diff
@@ -0,0 +1,143 @@
...
+        content_type = ContentType.objects.get_for_model(self)
+        return MetadataAttribute.objects.filter(
+            content_type=content_type, object_id=self.pk, key=key
+        ).delete()
+
+    def delete_attribute_if_has_permission(self, user, key):
+        self.can_delete_metadata(user)
+        return self.delete_attribute(key)
+
+    def get_attributes(self, **kwargs):
```
qgerome, line 130: can't you just use a `GenericRelation` ?

hexa/metadata/models.py
```diff
@@ -0,0 +1,143 @@
...
+        content_type = ContentType.objects.get_for_model(self)
+        metadata_attr = MetadataAttribute.objects.filter(
+            content_type=content_type,
+            object_id=self.pk,
+            **kwargs,
+        )
+        return metadata_attr
+
+    def get_attributes_if_has_permission(
```
qgerome, line 139: I would not add a permission system at this level to get attributes. This can be the responsibility of the caller

hexa/metadata/schema/mutations.py
```diff
@@ -0,0 +1,88 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import PermissionDenied
+from django.db import IntegrityError
+
+from hexa.metadata.schema.utils import get_model_instance
+
+mutations = MutationType()
+
+
+@mutations.field("addMetadataToObject")
```
qgerome, line 11: I'd cut the `ToObject` of all mutations

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
+"""
+Interface for type implementing metadata
+"""
+interface HasMetadata {
```
qgerome, line 4: Having a variable name that represents a boolean looks like a bad idea. You can just call it `MetadataObject`

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+
+"""
+Union of all linked to metadata object types
+"""
+union LinkedObject = DatasetVersionFile | DatasetVersion | Dataset
+
+"""
+Generic metadata object
+"""
+type MetadataObject {
```
qgerome, line 26: This represents the first version of your PR I guess

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+  id: UUID!
+  key: String!
+  value: JSON
+  system: Boolean!
+}
+
+"""
+Union of all linked to metadata object types
+"""
+union LinkedObject = DatasetVersionFile | DatasetVersion | Dataset
```
qgerome, line 21: You are just breaking the concept of making it generic.

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+
+"""
+Generic metadata object
+"""
+type MetadataObject {
+  id: String!         # Encoded concatenation of the class model type and the instance id
+  object: LinkedObject!
+  attributes : [MetadataAttribute!]!
```
qgerome, line 29: Should be on the `HasMetadata` interface

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+Union of all linked to metadata object types
+"""
+union LinkedObject = DatasetVersionFile | DatasetVersion | Dataset
+
+"""
+Generic metadata object
+"""
+type MetadataObject {
+  id: String!         # Encoded concatenation of the class model type and the instance id
+  object: LinkedObject!
```
qgerome, line 28: to delete.

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+Interface for type implementing metadata
+"""
+interface HasMetadata {
+  extendedId: String!   # Encoded concatenation of the class model type and the instance id
+  metadata: MetadataObject!
+}
+"""
+Generic metadata attribute
+"""
+type MetadataAttribute {
```
qgerome, line 11: Since we have deleted the object between the business object and the attributes, you can add a 
`target/object: MetadataObject` here

hexa/metadata/graphql/schema.graphql
```diff
@@ -0,0 +1,107 @@
...
+  errors: [MetadataDeleteAttributeError!]!
+}
+
+type EditAttributeResult {
+  success: Boolean!
+  errors: [MetadataEditAttributeError!]!
+}
+
+extend type Query {
+  metadata(extendedId: String!): MetadataObject!
```
qgerome, line 97:
```suggestion
  metadataAttibutes(extendedId: String!): [MetadataAttribute!]!
```

hexa/metadata/schema/types.py
```diff
@@ -0,0 +1,51 @@
...
+from django.core.exceptions import PermissionDenied
+
+from hexa.datasets.models import Dataset, DatasetVersion, DatasetVersionFile
+
+has_metadata = InterfaceType("HasMetadata")
+linked_object = UnionType("LinkedObject")
+
+
+@linked_object.type_resolver
+def resolve_linked_object_union_type(obj, *_):
```
qgerome, line 11: What it we add another model ?

qgerome (changes requested):
`extentedId` should be a custom type and not a `String`. It feels weird to call that extendedId as well. I'd prefer to have a type `OpaqueID` that knows how to encode/decode itself.

Food for thought on how Ariadne-relay implemented the same concept of opaque id: https://github.com/g18e/ariadne-relay

hexa/metadata/graphql/schema.graphql
nazarfil: No, even though I have removed the model, I still kept the dictionary for graphql to return the desired object and extendedId, but it is repetitive and can be deleted

hexa/metadata/models.py
```diff
@@ -0,0 +1,143 @@
...
+
+    @staticmethod
+    def decode_base64_id(encoded_id):
+        decoded_bytes = base64.b64decode(encoded_id)
+        decoded_str = decoded_bytes.decode("utf-8")
+        id, model_type = decoded_str.split(":")
+        return id, model_type
+
+    def can_view_metadata(self, user: User):
+        permission = f"{self._meta.app_label}.view_{self.class_name}"
```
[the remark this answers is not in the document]
nazarfil: After implementing it several times I have moved it here, explicitly added such permission and did override it when it doesn't fit the assumption , for dataset - the user is not restricted by its dataset permissions to delete metadata

qgerome (commented):
Posted my review

qgerome (approved):

PR #802, opened 2024-09-03
fix(Datasets): increase slug size and only add suffix if needed

cheikhgwane:
See [issue](https://bluesquare.atlassian.net/jira/software/c/projects/HEXA/boards/330?assignee=712020%3A1d0055ff-af59-47b7-b48d-4d37e2a0d0b9&selectedIssue=HEXA-1030)


hexa/datasets/models.py
```diff
@@ -53,9 +56,12 @@ def create_if_has_perm(
             raise PermissionDenied
 
         created_by = principal if not isinstance(principal, PipelineRunUser) else None
+        # Before adding a random suffix, check if the couple (workspace,dataset_name) exists before
+        with_suffix = Dataset.objects.filter(workspace=workspace, slug=name).exists()
```
qgerome, line 60: This should be in the `create_dataset_slug` function

hexa/datasets/models.py
```diff
@@ -12,10 +12,13 @@
...
-def create_dataset_slug(name: str):
-    suffix = secrets.token_hex(3)
+def create_dataset_slug(name: str, with_suffix: bool = True):
+    suffix = ""
+    if with_suffix:
+        suffix = "-" + secrets.token_hex(3)
+
     prefix = slugify(name[: 64 - 3])
-    return prefix[:23].rstrip("-") + "-" + suffix
+    return prefix[:23].rstrip("-") + suffix
```
qgerome, line 21: You are still limiting the slug to 23 + suffix.

qgerome (changes requested):

hexa/datasets/models.py
qgerome: Is it really with the name of the dataset that you want to check if there is a dataset that already exists ?

qgerome (changes requested):

qgerome (changes requested):
What about the test case when a dataset already exist with the same slug ?

hexa/datasets/models.py
cheikhgwane: Oh, you're right it will not work.

PR #808, opened 2024-09-05
fix(Pipelines): convert uuid to str

cheikhgwane:


qgerome:
Already fixed and merged

PR #816, opened 2024-09-27
Fix wrong mgt of env var TRUST_FORWARDED_PROTO

toch:
When setting `TRUST_FORWARDED_PROTO` to `no`, it was anyway interpreted as a `yes` as we were only checking for its presence. That led to a misconfiguration of the cookie and csrf avoiding login.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -34,14 +34,14 @@ ENCRYPTION_KEY = os.environ.get("ENCRYPTION_KEY")
 DEBUG = os.environ.get("DEBUG", "false") == "true"
 
 # Trust the X_FORWARDED_PROTO header from the proxy or load balancer so Django is aware it is accessed by https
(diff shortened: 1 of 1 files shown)


config/settings/base.py
```diff
@@ -34,14 +34,14 @@
 DEBUG = os.environ.get("DEBUG", "false") == "true"
 
 # Trust the X_FORWARDED_PROTO header from the proxy or load balancer so Django is aware it is accessed by https
-if "TRUST_FORWARDED_PROTO" in os.environ:
+if os.environ("TRUST_FORWARDED_PROTO", "no") == "yes":
```
qgerome, line 37: Do you ming using the same string values for the bool as the other variables in this file ? ("true", "false")

qgerome (approved):

PR #817, opened 2024-09-30
feat(Core): add query to return password requirements

cheikhgwane:
Add a query to return django password requirements.


qgerome (approved):
Perfect!

PR #848, opened 2024-11-14
fix: silent email address trim

steve-lebleu:
Just apply a trim on the email input value in a reusable way.


qgerome (changes requested):
I do no think it's a good idea to expose this type to the frontend as it exposes a backend processing to the frontend part.

Let's go to a basic solution of `.strip()` in the `login()` mutation first.

qgerome:
Also, I think you do not have the pre-commit installed for the repo.

qgerome (changes requested):
In addition, could you write a test for this please ?

PR #849, opened 2024-11-14
fix: Remove and prevent saving pipeline versions with the same name

YolanFery:
Prevent user to create pipeline versions having the same name. 

## Changes

- Add an error type in GraphQL for this use case 
- Migration to add the contraint and remove existing duplicates by converting them to "........ (v1)" and "........ (v2)" for older versions and keeping the same name for the most recent version
- Add constraint to the model
- Catching and wrapping the exception

## How/what to test

- Migration of the BE succeeds and no more duplicates exist
- Trying to push a pipeline with a duplicate name fails gracefully

## Screenshots / screencast

Screenshot 2024-11-14 at 14 00 17


hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
```diff
@@ -0,0 +1,62 @@
...
+    for duplicate in duplicates:
+        versions = PipelineVersion.objects.filter(
+            pipeline=duplicate["pipeline"], name__contains=duplicate["name"]
+        ).order_by("created_at")
+
+        for index, version in enumerate(
+            list(versions)[:-1]
+        ):  # Skip the most recent version, keep it unchanged
+            old_name = version.name
+            version.name = f"{old_name} (v{index+1})"
```
YolanFery, line 26: Is this acceptable for our users ?

hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
```diff
@@ -0,0 +1,62 @@
...
+            logger.info(
+                f"Updated version name from {old_name} to {version.name} for {version.pipeline}"
+            )
+
+
+def revert_timestamp_from_duplicates(apps, schema_editor):
+    PipelineVersion = apps.get_model("pipelines", "PipelineVersion")
+    versions = PipelineVersion.objects.all()
+
+    version_pattern = re.compile(r" \(v\d+\)$")
```
YolanFery, line 37: Might lead to false positive cases but is it really a problem ?

hexa/pipelines/schema/mutations.py
```diff
@@ -328,6 +328,10 @@ def resolve_upload_pipeline(_, info, **kwargs):
         return {"success": False, "errors": ["INVALID_TIMEOUT_VALUE"]}
     except PermissionDenied:
         return {"success": False, "errors": ["PERMISSION_DENIED"]}
+    except IntegrityError as e:
+        if "unique_pipeline_version_name" in str(e):
```
YolanFery, line 332: Not fan of hardcoding the constraint name here, do we want to extract the name in a config file and avoid duplicate definition ?

qgerome: I agree. You could replace this by checking the presence of a version with the same name and returning the specific error code if it `.exists()`. I'd still catch the IntegrityError to return a broader error type. 🤷🏻

hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
qgerome: IMO I would not try to implement a reverse migration. We tend to use those when it's mandatory to transform values to make the application working after a downgrade (value types, specific set of values, ...).

hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
qgerome: Yes. I'm not sure about the transformation though. You leave the latest one untouched ("My Version") and then you take all the older version ordered by creation_date to have in the end:
"My Version (1)"; "My Version (2)"; "My Version (3)"; "My Version". 

It's a bit counterintuitive since it does not follow what users are used to with their FS. I would apply the same change to the latest one as well. WDYT?

qgerome (changes requested):
Having a few tests for this would be a good idea

hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
YolanFery: My reasoning is that the last one is the most important because it's the one used. So I wanted to keep it untouched so that it's not a breaking change
WDYT ?

hexa/pipelines/migrations/0046_pipelineversion_unique_pipeline_version_name.py
YolanFery: OK I will remove, thanks

hexa/pipelines/schema/mutations.py
YolanFery: Checking with `.exists()` is not concurrent-safe : if 2 requests check for existence at the same time they will both succeed the check and then one of them will fail with `IntegrityError`, so we have inconsistency in error management (same error can be caught in the `exists()` check and sometimes in IntegrityError). That's why I have not followed that route. 

If we think that we are OK with that in OH, in can implement it that way, if not we need to rely on the DB or any other transactional mechanism

qgerome: As of now we do not put lot of focus on this since we do not have the throughput that would make it plausible to happen. The result would not result in an incoherent DB neither but more of a broader error message to the user. It's a trade-off we chose until now. 

I'm not a fan of having hardcoded strings neither. I can propose you multiple solutions:
- We use your approach with a check in the string
- We check if the cause of the exception is `psycopg2.errors.UniqueViolation` (`e.__cause__` I think)
- We use a constant on the model to pass the name of the constraint

Do the one you prefer.

YolanFery: I followed the constant approach with also a check of the instance class of `__cause__`

(Interestingly [at least for my learning], I tried to override `save()` to call for `full_clean()` which is checking out-of-the-box the unique constraint before inserting in the DB. But it also checks for other attributes that should be non-null but for now are passed null from the SDK . Also this setup is  very similar to checking the IntegrityError so that's why I decided to keep it)

YolanFery:
Added tests

hexa/pipelines/migrations/0048_pipelineversion_unique_pipeline_version_name.py
```diff
@@ -0,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("pipelines", "0047_pipelineversion_rename_duplicate_pipeline_version_name")
```
YolanFery, line 6: It is recommended to not mix DDL and DML operations in one migration module 
- https://plotnikova1111.medium.com/django-migrations-tips-89f4e588a005
- https://docs.djangoproject.com/en/5.1/ref/migration-operations/#runpython

YolanFery: Reason why I splitted it in 2

hexa/pipelines/schema/mutations.py
```diff
@@ -330,6 +332,12 @@ def resolve_upload_pipeline(_, info, **kwargs):
         return {"success": False, "errors": ["INVALID_TIMEOUT_VALUE"]}
     except PermissionDenied:
         return {"success": False, "errors": ["PERMISSION_DENIED"]}
+    except IntegrityError as e:
+        if isinstance(
+            e.__cause__, UniqueViolation
+        ) and UNIQUE_PIPELINE_VERSION_NAME in str(e):
+            return {"success": False, "errors": ["DUPLICATE_PIPELINE_VERSION_NAME"]}
+        raise
```
YolanFery, line 340: I am not catching the `IntegrityError` because I think that it's unexpected and should raise the full exception

cheikhgwane:
LGTM!

qgerome:
Test @qgerome

qgerome:
Test 2 @qgerome

qgerome:
Test. 4 @YolanFery

qgerome (commented):
Test comment

qgerome (approved):

PR #850, opened 2024-11-19
feat: pgAdmin dev tool setup

steve-lebleu:
Just a small dev convenient setup to expose pgAdmin as service in local development.

## How/what to test

Just run the openhexa-app using the right setup, navigate to localhost:5050, define your connection and enjoy.


YolanFery (approved):
Tested, thanks

README.md
```diff
@@ -148,13 +147,56 @@ Create a new interpreter configuration in Pycharm with the following settings:
...
+# docker-compose.dev.yaml
+
+services:
+  pgadmin:
+    image: dpage/pgadmin4
+    environment:
+      PGADMIN_DEFAULT_EMAIL: YOUR_EMAIL_ADDRESS
+      PGADMIN_DEFAULT_PASSWORD: YOUR_PWD
+    ports:
+      - "5050:80"
```
qgerome, line 170:
```suggestion
    env_file:
      - path: ./.env
        required: true
    environment:
      PGADMIN_DEFAULT_EMAIL: ${PGADMIN_DEFAULT_EMAIL:-root@openhexa.org}
      PGADMIN_DEFAULT_PASSWORD: ${PGADMIN_DEFAULT_PASSWORD:-root}
      PGADMIN_CONFIG_SERVER_MODE: "False"
      PGADMIN_CONFIG_MASTER_PASSWORD_REQUIRED: "False"
    ports:
      - "${PGADMIN_PORT:-5050}:80"
    ```

qgerome (changes requested):
Too bad that we cannot easily have the hex-app server already configured (we could but it's not trivial)

I've adapted the definition to include load vars from the .env file, run pgadmin in user-mode and set other default credentials.

That way, we don't have to log in to access the panel.

PR #851, opened 2024-11-20
fix: adds byte field to json serialzer

nazarfil:
Adds byte field serialization for parquet files 
Fixes: 
https://bluesquareorg.sentry.io/issues/6077431049/events/37bd90ef025d4ac89daa782d7791f39a/?project=5893042&referrer=issue_details.related_trace_issue


hexa/datasets/models.py
```diff
@@ -357,6 +358,8 @@ def encode(self, obj):
         def replace_nan(item):
             if isinstance(item, float) and math.isnan(item):
                 return None
+            elif isinstance(item, bytes):
+                return base64.b64encode(item).decode("utf-8")
             elif isinstance(item, dict):
```
YolanFery, line 363: [Non-blocking] It's a bit confusing that there is actually no nan replacement for bytes as the name of the method suggests, I would maybe expect something like :

```suggestion
            elif isinstance(item, str) and item.lower() == "nan":
                return None
            elif isinstance(item, bytes):
                return replace_nan(base64.b64encode(item).decode("utf-8"))
```

Maybe I am just overcomplicating here :)

YolanFery (approved):

YolanFery (approved):

qgerome (approved):

PR #857, opened 2024-11-22
feat!: Set a default version number for each pipeline version, make version name optional

YolanFery:
Pipeline versions should auto-increment, a name for the version should be optional

## Changes

- Add the `version_number` to the `PipelineVerison` model
- Add constraints and index related to this new attribute
- Make the name optional
- Modify the admin slightly to show `version_number`
- Deprecate the `version` returned by uploadPipeline

## How/what to test

- Adding a new version does not require a name, and the version number is auto-incremented

## Screenshots / screencast

Screenshot 2024-11-22 at 11 17 56
Screenshot 2024-11-22 at 11 18 32
Screenshot 2024-11-22 at 11 18 55


hexa/pipelines/models.py
```diff
@@ -179,9 +206,13 @@ def is_schedulable(self):
     def is_latest_version(self):
         return self == self.pipeline.last_version
 
+    @property
+    def version_name(self):
+        return str(self.version_number) + (f" [{self.name}]" if self.name else "")
```
YolanFery, line 211: Not sure about this format ? Any ideas ?

hexa/pipelines/tests/migrations/test_0049_pipelineversion_set_version_numbers.py
```diff
@@ -0,0 +1,42 @@
...
+        return self.migrator.apps.get_model("pipelines", "PipelineVersion")
+
+    def test_set_version_numbers(self):
+        PipelineVersion = self.get_pipeline_version_model()
+        pipeline = self.get_pipeline_model().objects.create(code=self.pipeline_code)
+        for version_number, name, date in [
+            (3, "3", "2023-01-01"),
+            (10, "10", "2023-01-02"),
+            (20, "20", "2023-01-03"),
+        ]:
```
YolanFery, line 25: Provide dummy values to pass unique constraints in the model

hexa/pipelines/schema/mutations.py
```diff
@@ -324,7 +324,6 @@ def resolve_upload_pipeline(_, info, **kwargs):
             "success": True,
             "errors": [],
             "pipeline_version": version,
-            "version": version.name,  # FIXME: This is a temporary fix to not break the SDK before 1.0.43
```
YolanFery, line 327: I think it is a good time to remove this right ?

hexa/pipelines/migrations/0051_pipelineversion_version_number_cleanup.py
```diff
@@ -0,0 +1,27 @@
...
+            index=models.Index(
+                fields=["pipeline", "version_number"],
+                name="index_pipeline_version_number",
+            ),
+        ),
+        migrations.AlterField(
+            model_name="pipelineversion",
+            name="name",
+            field=models.CharField(max_length=250, null=True, blank=True),
+        ),
```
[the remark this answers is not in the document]
qgerome: This migration is going to break if you revert it after pipeline versions have been created without names no ?

hexa/pipelines/models.py
qgerome: I would assume that if the user wants to add a name, it has to be more important than the version number so I would invert the two to put the name in front and the number in second. WDYT?

hexa/pipelines/migrations/0051_pipelineversion_version_number_cleanup.py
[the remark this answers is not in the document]
YolanFery: Indeed, good catch thanks, it will likely break
I have not caught that during my testing, I will have a look and possibly provide a revert script filling the names

hexa/pipelines/models.py
nazarfil: I think keeping the version number in square brackets sets the right tone that it is autogenerated descriptive number

hexa/pipelines/tests/test_schema/test_pipeline_versions.py
```diff
@@ -342,3 +342,14 @@ def test_create_version_with_parameters(self, version="Filler", user=None):
         pipeline.schedule = "0 15 * * *"
         pipeline.save()
         self.assertEqual(r["data"]["uploadPipeline"]["success"], True)
+
+    def test_increment_version_number_on_save(self):
+        self.test_create_version("First version")
+        self.test_create_version("Second version")
```
nazarfil, line 348: Maybe add One more version without a name?

hexa/pipelines/models.py
qgerome: @nazarfil do you mean like this ?

- ABC [v1]
- ABCD [v2]
- v3
- ABCDE [v4]

I agree that having the generated number displayed with a slight style difference can bring its autogenerated nature.

hexa/pipelines/migrations/0051_pipelineversion_version_number_cleanup.py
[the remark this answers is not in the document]
YolanFery: Added 0052 to fill empty values

hexa/pipelines/models.py
YolanFery: Thanks all for the suggestions, I changed it in the latest version

nazarfil (approved):

PR #858, opened 2024-11-25
fix: skips geometry and processes wkb

nazarfil:
skips geometry field from porfiling and decodes wkb for sampling


hexa/datasets/models.py
```diff
@@ -363,14 +365,25 @@ def custom_encoding(item):
...
+    def _encode_bytes(self, item):
+        try:
+            return base64.b64encode(item).decode("utf-8")
+        except Exception:
+            try:
+                polygone = wkb.loads(item, hex=True)
+                geo_json = shapely.geometry.mapping(polygone)
+                return geo_json
+            except Exception:
+                return None
```
YolanFery, line 385: So we fail silently in this case ? Maybe we should fail or at least have a warning, WDYT ?

hexa/datasets/tests/test_generate_sample.py
```diff
@@ -200,6 +201,50 @@ def test_generate_sample(
...
+                None,
+            ),
+        ]
+        for (
+            fixture_name,
+            expected_status,
+            expected_sample,
+            expected_status_reason,
+        ) in CASES:
+            with self.subTest(fixture_name=fixture_name):
```
YolanFery, line 222: Today I learned about `subTest` (https://stackoverflow.com/a/29384495), thanks 😊

I have been used to decorator parametrization (https://pypi.org/project/parameterized/) instead

hexa/datasets/tests/test_generate_sample.py
```diff
@@ -200,6 +201,50 @@ def test_generate_sample(
...
+                fixture_file_path = os.path.join(
+                    os.path.dirname(__file__), f"./fixtures/{fixture_name}"
```
YolanFery, line 224: [Minor] You are using a mix of Path and string interpolation for the path creation, we might want to stick with Path

YolanFery (approved):

hexa/datasets/models.py
nazarfil: If we assume we won't have byte encoded tabular data then I am on board to discard it. So far it offers the option to preview the parquet file with wkb binary data in it. As in the example for geometry  it will display a dictionary with base64 encoded geometry field.
It does not bring much information to the data scientist working with it except knowing that it is geometry data..

qgerome: Let's assume that.

PR #861, opened 2024-11-27
feat: add existing member into a workspace and automate invitation ac…

steve-lebleu:
## Description

Implements upgrade of the workflow invitation flow:

- Allow workspace admin to search about worskpace candidates.
- If a candidate already exists in database, allow admin to bypass the invitation flow and create a membership for this user / workspace.
- If the email address is not known, an invitation to register is sent to the email address, and automaticaly accepted when the registration is done.

## Changes

- Query WorspaceCandidates + related types + resolver
- Mutation InsertWorkspaceMember + related types + resolver
- Email + template about the insertion into the workspace
- Mutation Register

## How/what to test

- Unit tests
- Functional test needs to checkout the [related feature branch on front-end side](https://github.com/BLSQ/openhexa-frontend/tree/feature/improve-workflow-workspace-invitation)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ RUN groupadd --gid 1000 $GROUP &&\
   passwd -d $USER
 
 RUN chown -R $USER:$GROUP /code/
-RUN mkdir /data && chown $USER:$GROUP /data
+RUN mkdir /data && chown -R $USER:$GROUP /data && chmod 777 -R /data
 USER $USER:$GROUP
 
 # Entry point
--- a/hexa/user_management/schema.py
+++ b/hexa/user_management/schema.py
@@ -39,7 +39,11 @@ from hexa.user_management.models import (
     Team,
     User,
 )
-from hexa.workspaces.models import WorkspaceInvitation, WorkspaceInvitationStatus
+from hexa.workspaces.models import (
+    WorkspaceInvitation,
+    WorkspaceInvitationStatus,
+    WorkspaceMembership,
+)
 
 from .utils import DEVICE_DEFAULT_NAME, default_device, has_configured_two_factor
 
@@ -375,11 +379,27 @@ def resolve_register(_, info, **kwargs):
         authenticated_user = authenticate(
             username=user.email, password=mutation_input["password1"]
         )
+
         login(request, authenticated_user)
-        return {"success": True, "errors": []}
 
     except ValidationError:
         return {"success": False, "errors": ["INVALID_PASSWORD"]}
+
+    try:
+        # Automate the invitation acceptance
+        invitation.status = WorkspaceInvitationStatus.ACCEPTED
+        invitation.save()
+
+        # Create the membership
+        WorkspaceMembership.objects.create(
+            workspace=invitation.workspace,
+            user=request.user,
+            role=invitation.role,
+        )
+
+        return {"success": True, "errors": []}
+    except Exception:
+        return {"success": False, "errors": ["INTERNAL_ERROR"]}
 
 
 @identity_mutations.field("resetPassword")
--- a/hexa/workspaces/graphql/schema.graphql
+++ b/hexa/workspaces/graphql/schema.graphql
@@ -177,7 +177,9 @@ enum WorkspaceMembershipRole {
   VIEWER  # The user can view the workspace and its contents.
 }
 
-
+type WorkspaceCandidatesResult {
+    items: [User!]!
+}
 
 extend type Query {
   # Returns a workspace by its unique identifier.
@@ -186,9 +188,10 @@ extend type Query {
   workspaces(query: String, page: Int, perPage: Int): WorkspacePage! @loginRequired
   # Returns the invitations to workspaces of the current user.
   pendingWorkspaceInvitations(page: Int! = 1, perPage: Int = 10): WorkspaceInvitationPage!
-
   # Returns a database table by its identifier.
   databaseTable(id:String!): DatabaseTable @loginRequired
+  # Returns a list of registered users not yet members of the workspace.
+  workspaceCandidates(query: String, workspace: String): WorkspaceCandidatesResult! @loginRequired
 }
 
 
@@ -233,6 +236,16 @@ enum InviteWorkspaceMembershipError {
   USER_NOT_FOUND
   WORKSPACE_NOT_FOUND
 }
+
+  """
+  Enum representing the possible errors that can occur when inviting a user to a workspace.
+  """
+  enum InsertWorkspaceMembershipError {
+    ALREADY_EXISTS
+    PERMISSION_DENIED
+    WORKSPACE_NOT_FOUND
+    SERVER_ERROR
+  }
 
 """
 Enum representing the possible errors that can occur when deleting a workspace member.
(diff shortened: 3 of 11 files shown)


hexa/workspaces/schema/queries.py
```diff
@@ -62,6 +65,44 @@ def resolve_pending_workspace_invitations(_, info, page=1, per_page=10):
...
+            queryset = (
+                queryset.filter(email__icontains=query)
+                .exclude(email__in=members.values_list("user__email", flat=True))
+                .order_by("email")[:10:1]
+            )
+
+        return {
+            "items": queryset,
+        }
+    except Exception:
```
qgerome, line 100: I'd prefer to catch the exceptions we know could happen (`Workspace.DoesNotExist`, `WorkspaceMembership.DoesNotExist`) to return an empty list and raise an error otherwise. Hiding errors this way could come back to haunt us in the future :p

hexa/workspaces/schema/queries.py
```diff
@@ -62,6 +65,44 @@ def resolve_pending_workspace_invitations(_, info, page=1, per_page=10):
     return result_page(qs, page=page, per_page=per_page)
 
 
+@workspace_queries.field("workspaceCandidates")
+def resolve_users(_, info, **kwargs):
+    request = info.context["request"]
+    query = kwargs["query"].strip()
+    workspaceSlug = kwargs["workspace"].strip()
+
+    try:
```
qgerome, line 74: I'd use the permissions to check if the user can "manage_members"

```python
if request.user.user.has_perm("workspaces.manage_members", workspace) is False:
  return []
```

hexa/workspaces/graphql/schema.graphql
```diff
@@ -177,7 +177,9 @@ enum WorkspaceMembershipRole {
   VIEWER  # The user can view the workspace and its contents.
 }
 
-
+type WorkspaceCandidatesResult {
+    items: [User!]!
```
qgerome, line 181: If you do not do a pagination, let remove this type and just return a list of User

hexa/workspaces/graphql/schema.graphql
```diff
@@ -186,9 +188,10 @@ extend type Query {
   workspaces(query: String, page: Int, perPage: Int): WorkspacePage! @loginRequired
   # Returns the invitations to workspaces of the current user.
   pendingWorkspaceInvitations(page: Int! = 1, perPage: Int = 10): WorkspaceInvitationPage!
-
   # Returns a database table by its identifier.
   databaseTable(id:String!): DatabaseTable @loginRequired
+  # Returns a list of registered users not yet members of the workspace.
+  workspaceCandidates(query: String, workspace: String): WorkspaceCandidatesResult! @loginRequired
```
qgerome, line 194: Return users directly
```suggestion
  workspaceCandidates(query: String, workspace: String): [User!]! @loginRequired
```

hexa/workspaces/schema/queries.py
qgerome: In fact, here is how I would have done it (with a few more years of Python :p) 

```python
@workspace_queries.field("workspaceCandidates")
def resolve_users(_, info, **kwargs):
    request = info.context["request"]
    query = kwargs["query"].strip()

    try:
        workspace = Workspace.objects.filter_for_user(request.user).get(slug=kwargs["workspaceSlug"])

        if not request.user.has_perm("workspaces.manage_members", workspace):
            raise PermissionDenied

        # Find users by looking in email, last name or first name for a case-insensitive match
        qs = User.objects.filter(
            Q(email__icontains=query)
            | Q(last_name__icontains=query)
            | Q(first_name__icontains=query)
        )

        # Exclude users that are already members of the workspace
        qs = qs.exclude(id__in=workspace.members.values_list("id", flat=True))

        return qs.order_by("email")[:10]
    except PermissionDenied:
        return []
    except Workspace.DoesNotExist:
        return []


```

hexa/workspaces/templates/workspaces/mails/insert_user.html
qgerome, line 1: The name of the file does not reflect its purpose. It's more a member that has been added.

hexa/workspaces/utils.py
```diff
@@ -38,3 +38,24 @@ def send_workspace_invitation_email(
             },
             recipient_list=[invitation.email],
         )
+
+
+def send_workspace_insertion_email(workspace: Workspace, owner: User, user: User):
```
qgerome, line 43: Maybe?
```suggestion
def send_workspace_member_creation_email(workspace: Workspace, owner: User, user: User):
```

qgerome (changes requested):
Very nice for someone that was not used to code in Python and Django :)

Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN groupadd --gid 1000 $GROUP &&\
   passwd -d $USER
 
 RUN chown -R $USER:$GROUP /code/
-RUN mkdir /data && chown $USER:$GROUP /data
+RUN mkdir /data && chown -R $USER:$GROUP /data && chmod 777 -R /data
```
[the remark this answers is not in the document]
steve-lebleu: Indeed

hexa/user_management/schema.py
```diff
@@ -375,12 +379,28 @@ def resolve_register(_, info, **kwargs):
...
+
         login(request, authenticated_user)
-        return {"success": True, "errors": []}
 
     except ValidationError:
         return {"success": False, "errors": ["INVALID_PASSWORD"]}
 
+    try:
+        # Automate the invitation acceptance
+        invitation.status = WorkspaceInvitationStatus.ACCEPTED
```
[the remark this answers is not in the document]
steve-lebleu: What do you mean ? The user can have many invitations for the same workspace at the same time, and you would like in this scenario to accept all pending invitations ?

hexa/workspaces/graphql/schema.graphql
```diff
@@ -613,6 +654,7 @@ extend type Mutation {
   deleteWorkspace(input: DeleteWorkspaceInput!): DeleteWorkspaceResult! @loginRequired
   archiveWorkspace(input: ArchiveWorkspaceInput!): ArchiveWorkspaceResult! @loginRequired
   inviteWorkspaceMember(input: InviteWorkspaceMemberInput!): InviteWorkspaceMemberResult! @loginRequired
+  insertWorkspaceMember(input: InsertWorkspaceMemberInput!): InsertWorkspaceMemberResult! @loginRequired
```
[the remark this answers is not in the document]
steve-lebleu: It's a point. My idea was to clearly separate all the concerns / responsabilities between endpoints, because some behaviors and validations was quite differents. The merge of the two features into one single endpoint was coming with a concern about the concern separation, and the testing of this scenario. To sum up, despite the fact the implementation can seems to be "over simplified", I think it's interesting to separate the concerns for the reason mentionned. WDYM ?

hexa/user_management/schema.py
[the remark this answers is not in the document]
qgerome: I'd accept all pending invitations for all workspaces.

hexa/workspaces/graphql/schema.graphql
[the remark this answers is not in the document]
qgerome: If we look at the GraphQL layer as the API we propose to external developers, it's less obvious why you should call one or the other since the variables are the same. It's the system that has to adapt to the two cases (where the goal is the same : Having a new user linked to the workspace) not the api user.

bramj:
Closing, duplicate of https://github.com/BLSQ/openhexa-app/pull/1123 (actually the other way around I guess, but doesn't really matter :slightly_smiling_face:)

PR #863, opened 2024-12-03
Fix remove byte profiling

nazarfil:
- Adds the number of rows without nan per column
- skips any byte column for profiling


qgerome (commented):
I do not think you have solved the problem. What about the sample that is also going to be stored in DB ?

nazarfil:
> I do not think you have solved the problem. What about the sample that is also going to be stored in DB ?

The sample part was already resolved i previous PR, it will add <SKIPPED_BYTES> entry to the skipped field

qgerome (approved):

PR #864, opened 2024-12-04
fix(datasets): Set the correct engine for xls files & do not crash when DatasetVersionFile do not exist anymore

qgerome:
I don't understand how it was possible to run the tests with xlsx files without the engine ?


nazarfil (approved):

nazarfil:
> I don't understand how it was possible to run the tests with xlsx files without the engine ?

Well i am surprised as well, as for tests it works as supposed, pandas checks for available engines and picks openpyxl..

PR #865, opened 2024-12-05
fix : Pipeline Version parameters were converted to snake_case with invalid format

YolanFery:
We witnessed some Pipeline Version parameters like `dhis2_connection` converted to `dhis_2__connection` into the  database config, resulting in mismatch errors when running the pipeline. 

The issue is coming from the fact we use `convert_kwargs_to_snake_case` https://github.com/BLSQ/openhexa-app/blob/168dd660bd8b5c77ed0ec28093833e583dad3928/hexa/pipelines/schema/mutations.py#L343 to have snake_case formatted variable in Python. But this also affects the config passed as a JSON https://github.com/BLSQ/openhexa-app/blob/168dd660bd8b5c77ed0ec28093833e583dad3928/hexa/pipelines/graphql/schema.graphql#L558

I don’t think we want to modify the format of values in the JSON config.

Also `convert_kwargs_to_snake_case` is on the deprecation path (https://ariadnegraphql.org/docs/api-reference#convert_kwargs_to_snake_case).

Since it is on the deprecation path I tried to replace it by the recommendation and it seems to solve the issue (the JSON value is not changed, just the name of the top properties). 

So this PR is about the recommended deprecation path.

Note that all graphQL methods can be affected by this upgrade.

## Changes

- Remove usage of `convert_kwargs_to_snake_case` and replace it by `convert_names_case=True`
- Rename attributes to `snake_case`
- Extend a unit test to cover the issue that need fixing

## How/what to test

GraphQL resolvers are still behaving like expected


qgerome (approved):
LGTM

nazarfil (approved):

PR #869, opened 2024-12-10
HEXA-1111 stringify env var

nazarfil:
Stringify env var


YolanFery (approved):
First bug :( 
Thanks for fixing, I guess I didn't face it during development because in local it runs docker and here it's kubernetes pod, right ?
https://github.com/BLSQ/openhexa-app/blob/5120d509accce9f38a6b91cce7f87726bcc6fd79/hexa/pipelines/management/commands/pipelines_runner.py#L334-L337

cheikhgwane (approved):

PR #873, opened 2024-12-12
fix(Datasets): snake case for datasetlink id

cheikhgwane:
Small fix : forgot to migrate to snake case the linkId used in pinDataset mutation.

## Changes

- Use snake case for link id.

## How/what to test

Call the pinDataset mutation with correct input, it shouldn't break.


YolanFery (approved):
Argh, I missed at least one, thanks for fixing !

qgerome (approved):

PR #878, opened 2024-12-16
fix(Pipelines): Put back 'number' on PipelineVersion to not break backward compatibility with SDK < 1.0.0

qgerome:
Fixes OPENHEXA-1BT

## Changes

I added the `number` field again on `PipelineVersion`

## How/what to test

Try uploading & downloading pipelines using a version of openhexa.sdk < 1.0.0


YolanFery (approved):
Tested to push and download pipelines using openhexa, version 0.3.1

PR #883, opened 2024-12-20
feat: HEXA-1146 Propagate pipeline default parameters in new versions

YolanFery:
Uploading a new pipeline version was overriding config values. This PR is about keeping the previous config values while taking into account parameters that can change.

## Changes

- Add a helper method to get the previous config while considering changes in parameters 
- Unit test covering 3 scenarios

## How/what to test

- Adding new versions keeps the config
- Adding new versions with different or partially different parameter keeps the config for the previously existing ones


hexa/pipelines/models.py
```diff
@@ -340,6 +340,37 @@ def is_schedulable(self):
...
             return self.last_version and self.last_version.is_schedulable
 
+    def get_config_from_previous_version(self, new_parameters: dict):
+        """
+        Get the config from the previous version of the pipeline considering only common parameters between the new and the previous version.
+        """
+        previous_config_from_common_parameter = {}
+        if self.last_version:
+            previous_parameters = self.last_version.parameters
+            common_parameters = [
```
nazarfil, line 350: I have only one comment on the naming, that "common" might not reflect the best that we are talking about overlapping parameters between new and previous versions

YolanFery: Makes a lot of sense, thanks, fixed in next commits

nazarfil (approved):

PR #885, opened 2024-12-20
feat: HEXA-1122 list of templates in create pipeline dialog

YolanFery:
GraphQL endpoint for showing the list of templates

## Changes

- Schema change
- Add a current version to simplify the frontend
- GraphQL endpoint with paging
- GraphQL test to test the paging
- Add the search functionality

## How/what to test

A paged list of templates is correctly returned

PR #887, opened 2024-12-23
feat : HEXA-1124 button to create a new pipeline from a template

YolanFery:
GraphQL endpoint to create a new pipeline from a template

## Changes

- Endpoint and model method to create a pipeline
- Add `source_template` to pipeline to keep a reference of the template used to create this pipeline
- Unit test to cover creation of pipeline and duplicate creation of pipeline

## How/what to test

Creating a pipeline from a template using GraphQL endpoint is working


nazarfil (approved):

hexa/pipeline_templates/schema/mutations.py
```diff
@@ -74,4 +75,40 @@ def resolve_create_pipeline_template_version(_, info, **kwargs):
...
+        return {"success": False, "errors": ["PERMISSION_DENIED"]}
+
+    try:
+        template_version = PipelineTemplateVersion.objects.get(
+            id=input["pipeline_template_version_id"]
+        )
+    except PipelineTemplateVersion.DoesNotExist:
+        return {"success": False, "errors": ["PIPELINE_TEMPLATE_VERSION_NOT_FOUND"]}
+
+    pipeline_code = f"{template_version.template.source_pipeline.code} (from Template)"
```
nazarfil, line 97: Can you give more context about this comment? The (from Template) would be appended to the pipeline code and would break the parsing if we would reuse it as it.
So I would like to know if the expected flow is for the user to open the pipeline code and edit it, and the comment line will indicate that code is copied form template and has to be edited? 
If the goal of it is to mark it as code form tempalte then i would append it as a python comment "/n # (from Template)"

YolanFery: Thanks for raising this

The intend was to create a different name for the pipeline that is created but you are correct that it makes the code unparseable. In addition to mis-align the python code and the Database object (the `@pipeline(code= , name= )` would need to be modified )  

So, as part of https://github.com/BLSQ/openhexa-app/pull/885, I am using the same name as the original Pipeline to avoid such complexity at this point

PR #888, opened 2024-12-26
feat: Hexa 1117 dhis2 graphql interface

nazarfil:
## Changes
Initialization of toolbox client from workspace slug
Interface to interact with **DHIS2** using toolbox client

## How/what to test
GraphQL calls to the interface


hexa/connections/dhis2_client_helper.py
```diff
@@ -0,0 +1,52 @@
...
+        "organisation_units": dhis2.meta.organisation_units,
+        "organisation_unit_groups": dhis2.meta.organisation_unit_groups,
+        "organisation_unit_levels": dhis2.meta.organisation_unit_levels,
+        "datasets": dhis2.meta.datasets,
+        "data_elements": dhis2.meta.data_elements,
+        "data_element_groups": dhis2.meta.data_element_groups,
+        "indicator": dhis2.meta.indicators,
+        "indicator_groups": dhis2.meta.indicator_groups,
+    }
+    return metadata_methods[type](**kwargs)
```
YolanFery, line 52: What's the benefit of this method ? 

It seems like an unnecessary layer (adding a metadata requires adding it here as well, string values need to be kept aligned with function names,...)

What about simply using : `dhis2.meta.data_elements(fields="id,name")` ?

hexa/connections/graphql/schema.graphql
```diff
@@ -0,0 +1,29 @@
...
+    CONNECTION_ERROR
+    UNKNOWN_ERROR
+}
+"""
+DHIS2 metadata query result
+"""
+type DHIS2QueryResult {
+    data: [DHIS2MetadataItem!]  # List of metadata items
+    errors: [String!]!      # List of errors
+}
```
YolanFery, line 18: I have seen that usually, we use a `success` boolean as well (https://github.com/BLSQ/openhexa-app/blob/d9d7159d6fc6d1982ca6f4067fb04c878dda3ceb/hexa/files/graphql/schema.graphql#L117 and https://github.com/BLSQ/openhexa-app/blob/d9d7159d6fc6d1982ca6f4067fb04c878dda3ceb/hexa/files/graphql/schema.graphql#L142)

hexa/connections/schema/types.py
```diff
@@ -0,0 +1,31 @@
...
+def resolve_query(dhis2_client, info, **kwargs):
+    try:
+        metadata = get_dhis2_metadata(
+            dhis2_client,
+            type=kwargs.get("type"),
+            fields=kwargs.get("fields", "id,name"),
+            filter=kwargs.get("filter"),
+        )
+
+        result = [{"id": item.get("id"), "name": item.get("name")} for item in metadata]
```
YolanFery, line 21: The result is always ids and names even though fields can be any attributes ?

hexa/connections/dhis2_client_helper.py
YolanFery: If we want to keep some dynamism ( for instance to use in `@dhis2_connection.field("query")
def resolve_query(dhis2_client, info, **kwargs):`) maybe we can explore the following approach :

```

def get_dhis2_metadata(dhis2: DHIS2, type: str, **kwargs) -> dict:
    """
    Gets metadata from DHIS2
    """
    try:
        metadata_method = getattr(dhis2.meta, type)
        return metadata_method(**kwargs)
    except AttributeError:
        raise ValueError(f"Invalid metadata type: {type}")
```

It keeps maintenance low while conserving the dynamic aspect

hexa/connections/tests/test_dhis2connection.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+        responses.add(
+            responses.GET,
+            "http://127.0.0.1:8080/api/dataSets?fields=id%2Cname&pageSize=1000",
+            json={"dataSets": datasets},
+            status=200,
+        )
+        dhis2 = get_client_by_slug("dhis2-connection-1", self.USER_JIM)
+        metadata = get_dhis2_metadata(dhis2, "datasets", fields="id,name")
+        self.assertIsNotNone(metadata)
```
YolanFery, line 171: Maybe we could get some value, at least for some tests, of asserting the state of metadata and not only checking that it's not `None` ?

hexa/connections/dhis2_client_helper.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    dhis_connection_dict = {field.code: field.value for field in fields}
+    try:
+        return DHIS2(
+            url=dhis_connection_dict.get("url"),
+            username=dhis_connection_dict.get("username"),
+            password=dhis_connection_dict.get("password"),
+        )
+    except DHIS2Error:
+        return None
```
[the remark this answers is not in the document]
nazarfil: I agree, glad to have your opinion on it, we can leave the exception raised

hexa/connections/dhis2_client_helper.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if not dhis2_connection:
+        return None
+
+    fields = (
+        ConnectionField.objects.filter(connection=dhis2_connection)
+        .select_related()
+        .all()
+    )
```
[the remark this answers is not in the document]
nazarfil: Sure, it is better to use fields attribute!

hexa/connections/dhis2_client_helper.py
nazarfil: I would still keep it with the method mapping. This way we could for example replace type from snake_case to camel case as it  would follow the GraphQL naming convention

hexa/connections/schema/types.py
nazarfil: Yeah, we only need id and name for the feature, so i did limit the result with id name as metadataResult. To make it explicit I could name it metadataQuery, do you think it will be more clear?

YolanFery: I don't tink the name is a problem.

But it means that `fields` parameter has no effect at all on the result of the query since we always return id and name ? 

Also if id and name are not in the `fields`, the result will be `None` for all items. 

I think that either we make it generic and support `fields` or we remove it to avoid issues like the one described above

nazarfil: i agree that in scope of the current feature it seems contradictory to offer the fields parameter and in the same time to restrict the result to fields and name. I will lock it to id,name to fetch metadata

hexa/connections/dhis2_client_helper.py
```diff
@@ -0,0 +1,46 @@
...
+        username=dhis_connection_dict.get("username"),
+        password=dhis_connection_dict.get("password"),
+    )
+
+
+def query_dhis2_metadata(dhis2: DHIS2, type: str, **kwargs) -> dict:
+    """
+    Gets metadata from DHIS2
+    """
+    kwargs.setdefault("fields", "id,name")
```
YolanFery, line 35: [Minor] since we enforce `id` and `name` for the result in 
```
@dhis2_connection.field("query")
def resolve_query(dhis2_client, info, **kwargs):
```
I would enforce the `fields` to `id,name` at the same place as well and keep the `query_dhis2_metadata` method generic as it was

Something like :

```
@dhis2_connection.field("query")
def resolve_query(dhis2_client, info, **kwargs):
    fields = ["id", "name"]
    try:
        metadata = query_dhis2_metadata(
            dhis2_client,
            type=kwargs.get("type"),
            fields = ",".join(fields),
            filter=kwargs.get("filter"),
        )

        result = [{field: item.get(field) for field in fields} for item in metadata]
```

This avoids repetition of restrained fields, collocates the restrictions and keeps the method `query_dhis2_metadata` generic for usage in other parts of the code

YolanFery (approved):

hexa/connections/dhis2_client_helper.py
nazarfil: I agree that coding wise it is better to do it this way and avoid repetition, but having the dictionary with explicitly set keys that match the GraphQL type does help to keep both consistent.

YolanFery: Not sure I totally get your point, maybe you can explain me during the stand-up
But it's a Minor comment so I am not blocking for that

PR #890, opened 2024-12-31
feat: HEXA-1125 endpoint for source template and check if a new template version is available for upgrade

YolanFery:
The goal is to enable the Frontend to show the source template of a pipeline and if a new version of the template is available for upgrade.

## Changes

- Add a `source_template_version` attribute to the PipelineVersions to know which one was used to create the version.
- Add a property `is_new_template_version_available` to Pipelines to check if it can be upgraded
- Integration test to check if the available upgrade is correctly detected

## How/what to test

Pipelines created using a template have a True `newTemplateVersionAvailable` if the template has a new version

## Screenshots / screencast

This logic enables the Frontend to show such type of data/feature : 

Screenshot 2024-12-31 at 14 37 15


hexa/pipelines/tests/test_schema/test_pipelines.py
```diff
@@ -2574,3 +2574,73 @@ def test_delete_pipeline_recipient(self):
...
+                }
+            }
+            """,
+            {
+                "id": str(
+                    r["data"]["createPipelineFromTemplateVersion"]["pipeline"]["id"]
+                )
+            },
+        )
+        self.assertEqual(True, r["data"]["pipeline"]["newTemplateVersionAvailable"])
```
nazarfil, line 2646: With  additional check on null if there is no source_template it is all good

YolanFery: Added, thanks

nazarfil (approved):

PR #891, opened 2025-01-02
fix: HEXA-1152 Install polars for Apple Chips

YolanFery:
When running the project on Apple Chip (M4), I get the following warning :

```
/usr/local/lib/python3.12/site-packages/polars/_cpu_check.py:258: RuntimeWarning: Missing required CPU features.

The following required CPU features were not detected:
    avx, avx2, fma, bmi1, bmi2, lzcnt, movbe
Continuing to use this version of Polars on this processor will likely result in a crash.
Install the `polars-lts-cpu` package instead of `polars` to run Polars with better compatibility.

Hint: If you are on an Apple ARM machine (e.g. M1) this is likely due to running Python under Rosetta.
It is recommended to install a native version of Python that does not run under Rosetta x86-64 emulation.

If you believe this warning to be a false positive, you can set the `POLARS_SKIP_CPU_CHECK` environment variable to bypass this check.

  warnings.warn(
/code/docker-entrypoint.sh: line 33:    19 Illegal instruction     python manage.py migrate
```

## Changes

- Conditionally install `polars-lts-cpu`

## How/what to test

The project can be run on Apple Chips

PR #894, opened 2025-01-06
chore: Build images for the dev's arch

qgerome:
Instead of trying to compile polars to run in a emulated amd64 on arm64, let's try first to run the app in a non emulated env.


nazarfil (approved):

YolanFery:
It works and is better than https://github.com/BLSQ/openhexa-app/pull/891, thanks

PR #897, opened 2025-01-12
Mixpanel: Prevent our integration to block our backend

qgerome:
By default it waited a timeout on the mixpanel side and retried all requests 4 times.

Fixes OPENHEXA-1CH & OPENHEXA-1BM

### DIFF
--- a/hexa/analytics/api.py
+++ b/hexa/analytics/api.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.http import HttpRequest
-from mixpanel import Mixpanel
+from mixpanel import Consumer, Mixpanel
 from sentry_sdk import capture_exception
 from ua_parser import user_agent_parser
 
@@ -8,7 +8,15 @@ from hexa.user_management.models import AnonymousUser, User
 
 mixpanel = None
 if settings.MIXPANEL_TOKEN:
-    mixpanel = Mixpanel(token=settings.MIXPANEL_TOKEN)
+    mixpanel = Mixpanel(
+        token=settings.MIXPANEL_TOKEN,
+        consumer=Consumer(
+            retry_limit=0,
+            # It’s a good practice to set connect timeouts to slightly larger than a multiple of 3, which is the default TCP packet retransmission window.
+            # https://requests.readthedocs.io/en/latest/user/advanced/#timeouts
+            request_timeout=3.05,
+        ),
+    )
 
 
 def track(


nazarfil (approved):

hexa/analytics/api.py
```diff
@@ -1,14 +1,22 @@
...
-    mixpanel = Mixpanel(token=settings.MIXPANEL_TOKEN)
+    mixpanel = Mixpanel(
+        token=settings.MIXPANEL_TOKEN,
+        consumer=Consumer(
+            retry_limit=0,
+            # It’s a good practice to set connect timeouts to slightly larger than a multiple of 3, which is the default TCP packet retransmission window.
+            # https://requests.readthedocs.io/en/latest/user/advanced/#timeouts
+            request_timeout=3.05,
+        ),
+    )
```
YolanFery, line 19: [Questions] 

Can you please explain what will happen in case it times out ? Does it fail silently or we still get a warning somehow ? 

If it fails silently, how will we catch configuration errors ? 

If failed events are not stored anywhere it also means that the analytics are not consistent (any downtime from Mixpanel will result in missing analytics). 

Just trying to understand what are the limitations with this. A common pattern (implemented by AWS SQS) is using dead-letter queue. It is also mentioned in Mixpanel Error handling documentation (https://docs.mixpanel.com/docs/tracking-methods/integrations/google-pubsub#step-4-connecting-your-production-pipeline)

YolanFery (approved):

hexa/analytics/api.py
qgerome: > Can you please explain what will happen in case it times out ? Does it fail silently or we still get a warning somehow ?
In case of timeout, an error is raised and catch inside the two functions that use the Mixpanel client (see `api.py`). Those errors are pushed in Sentry as well.

> If failed events are not stored anywhere it also means that the analytics are not consistent (any downtime from Mixpanel will result in missing analytics).

Indeed it means that we could have inconsistent analytics. Since our analytics are currently used to help us take decisions on features and check their adoption and not for billing purposes we can accept a lower reliability in case of downtimes from Mixpanel. 

> Just trying to understand what are the limitations with this. A common pattern (implemented by AWS SQS) is using dead-letter queue. It is also mentioned in Mixpanel Error handling documentation (https://docs.mixpanel.com/docs/tracking-methods/integrations/google-pubsub#step-4-connecting-your-production-pipeline)

We could use this pattern indeed. In the future, I'd like to move the analytics in a separate queue and not take request time to send them.

YolanFery: Clear and aligned, thanks

PR #898, opened 2025-01-14
chore(Templates): Add missing fixture to create the feature flag

qgerome:
It's better to use the fixture mechanism of Django to manage feature flags.


YolanFery:
Nice, didn't know about this, thanks

YolanFery (approved):

PR #901, opened 2025-01-15
feat: HEXA-1127 button to create a new version of the pipeline using the new template

YolanFery:
GraphQL mutation for upgrading a version of pipeline created using a template

## Changes

- Split the `create_pipeline` model to separate creation of pipeline and pipelineVersikon to reuse it
- New `upgradePipelineVersionFromTemplate` mutation with associated test


hexa/pipeline_templates/models.py
```diff
@@ -100,18 +100,27 @@ class Meta:
 
     objects = PipelineTemplateVersionQuerySet.as_manager()
 
-    def create_pipeline(self, code, workspace, user):
+    def _create_pipeline(self, workspace):
         source_pipeline = self.template.source_pipeline
-        source_version = self.source_pipeline_version
-        pipeline = Pipeline.objects.create(
+        if Pipeline.objects.filter(
+            workspace=workspace, code=source_pipeline.code
```
qgerome, line 106: I don't think we should restrict the use of the same template to have multiple pipelines in a workspace. You could want to duplicate it to handle multiple input params for example

hexa/pipeline_templates/models.py
```diff
@@ -100,18 +100,27 @@ class Meta:
...
-            code=code,
+            code=source_pipeline.code,
             name=source_pipeline.name,
             description=self.template.description,
             config=source_pipeline.config,
             workspace=workspace,
         )
-        PipelineVersion.objects.create(
+
+    def create_pipeline_version(self, workspace, user, pipeline=None):
```
qgerome, line 120: WDYT of putting the user first and name if as we do usually `principal`?

qgerome: Could you type the signature of your methods ?

hexa/pipelines/graphql/schema.graphql
```diff
@@ -329,6 +329,31 @@ input DeletePipelineRecipientInput {
   recipientId: UUID!
 }
 
+"""
+Represents the input for upgrading a pipeline version from the latest template version.
+"""
+input UpgradePipelineVersionFromTemplateInput {
+  id: UUID!
```
qgerome, line 336: As it is, I have no clue of what is `id`. Is it the pipeline ? Do not hesitate to change it to a more precise name.

hexa/pipelines/schema/mutations.py
```diff
@@ -587,4 +587,30 @@ def resolve_add_pipeline_output(_, info, **kwargs):
...
+        return {
+            "success": False,
+            "errors": ["PIPELINE_NOT_FOUND"],
+        }
+    if not pipeline.source_template:
+        return {"success": False, "errors": ["PIPELINE_NOT_FROM_TEMPLATE"]}
+    if not pipeline.is_new_template_version_available:
+        return {"success": False, "errors": ["NO_NEW_TEMPLATE_VERSION_AVAILABLE"]}
+
+    pipeline_version = pipeline.source_template.last_version.create_pipeline_version(
```
qgerome, line 608: So, every user (viewer, editor, admin) could call this and upgrade the pipeline ?

qgerome: IMO, it would make sense to have a simple api such as : 
```python
template.upgrade_pipeline(pipeline, version=None | template.last_version)
```

hexa/pipeline_templates/models.py
YolanFery: We have a constraint of unique code per workspace for pipelines : https://github.com/BLSQ/openhexa-app/blob/b7c7b7775c9be0de8b901052959b8b3dde7e8991/hexa/pipelines/models.py#L277

And changing the code of the pipeline requires parsing of the Python code for changing the `@pipeline` decorator arguments. We don't to take this route now, do we ?

hexa/pipeline_templates/models.py
YolanFery: Done

I tried to enable globally the ruff rule but there over 3000errors to fix and no clear way to set it as a warning. So I left this endeavour for another time :)

hexa/pipelines/graphql/schema.graphql
YolanFery: Definitely, thanks !

hexa/pipelines/schema/mutations.py
YolanFery: Added a permission check and the abstraction layer proposed, good idea, thanks

hexa/pipeline_templates/models.py
qgerome: You're right. I think we'll have to address this limitation soon.

qgerome:
Nice. Feel free to merge when you'd like

hexa/pipeline_templates/models.py
YolanFery: I have created a ticket for that https://bluesquare.atlassian.net/browse/HEXA-1170 
It will require some brainstorming

nazarfil (approved):

PR #902, opened 2025-01-16
chore(deps): Upgrade Django from v4 to v5

qgerome:
I upgraded Django to the latest version available.

https://docs.djangoproject.com/en/5.1/releases/5.0/

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -2,7 +2,7 @@ ariadne
 ariadne_django
 colorhash
 cryptography
-Django
+Django>=5
 django-cors-headers
 django-countries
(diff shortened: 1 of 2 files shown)


YolanFery (approved):
Did a local sanity check and it looks good, didn't expect to be such an easy upgrade :D

cheikhgwane (approved):
Build and run tests in local, everything looks good

qgerome:
> Did a local sanity check and it looks good, didn't expect to be such an easy upgrade :D

Most of the time it's the case with Django :p

PR #906, opened 2025-01-20
fix: HEXA-1163 - refactor dhis2 client, move connection to workspace

nazarfil:
- refactor **DHIS2Connection** to implement **Connection** interface
- move query type to enum 
- move connections into workspace


hexa/workspaces/utils.py
```diff
@@ -38,3 +40,45 @@ def send_workspace_invitation_email(
...
+        DHIS2MetadataQueryTypes.ORGANISATION_UNITS.value: dhis2.meta.organisation_units,
+        DHIS2MetadataQueryTypes.ORGANISATION_UNIT_GROUPS.value: dhis2.meta.organisation_unit_groups,
+        DHIS2MetadataQueryTypes.ORGANISATION_UNIT_LEVELS.value: dhis2.meta.organisation_unit_levels,
+        DHIS2MetadataQueryTypes.DATASETS.value: dhis2.meta.datasets,
+        DHIS2MetadataQueryTypes.DATA_ELEMENTS.value: dhis2.meta.data_elements,
+        DHIS2MetadataQueryTypes.DATA_ELEMENT_GROUPS.value: dhis2.meta.data_element_groups,
+        DHIS2MetadataQueryTypes.INDICATORS.value: dhis2.meta.indicators,
+        DHIS2MetadataQueryTypes.INDICATOR_GROUPS.value: dhis2.meta.indicator_groups,
+    }
+    return metadata_methods[type](**kwargs)
```
YolanFery, line 70: [Minor] 

Maybe using something more dynamic to reduce manual mapping error, difficult reading and maintenance :

```
def query_dhis2_metadata(dhis2, type: DHIS2MetadataQueryTypes, **kwargs) -> dict:
    """
    Gets metadata from DHIS2
    """
    return getattr(dhis2.meta, type.value.lower())(**kwargs)   
```
    
    
  Also using the enum type for type instead of str makes it easier to use

hexa/workspaces/utils.py
```diff
@@ -38,3 +40,45 @@ def send_workspace_invitation_email(
             },
             recipient_list=[invitation.email],
         )
+
+
+class DHIS2MetadataQueryTypes(Enum):
```
YolanFery, line 45: [Very-Minor] Maybe a StrEnum (or LowercaseStrEnum or UppercaseStrEnum) would be more appropriate

YolanFery (approved):
Great, seems cleaner indeed !

hexa/workspaces/utils.py
nazarfil: Good suggestion!

hexa/workspaces/utils.py
nazarfil: changes the type from str to enum. for the getattr i would avoid linking so tightly the methods to type names to leave the flexibility to assign any key

YolanFery (approved):

hexa/workspaces/graphql/schema.graphql
```diff
@@ -854,4 +854,122 @@ extend type Mutation {
...
+
+"""
+DHIS2 metadata item
+"""
+type DHIS2MetadataItem {
+    id: String!     # Unique identifier
+    name: String!   # Name of the metadata item
+}
+enum DHIS2ConnectionError {
+    CONNECTION_ERROR
```
qgerome, line 867: I see what you mean by "CONNECTION_ERROR" but maybe we could use another wording like "REQUEST_ERROR" no?

PR #907, opened 2025-01-20
feat: Changelog support  for pipeline templates (HEXA-1166)

YolanFery:
Allow to specify a changelog when upgrading the pipeline

## Changes

- Schema change
- `availableUpgradePipelineTemplateVersions` to list the new versions

Related to https://github.com/BLSQ/openhexa-frontend/pull/1023


hexa/pipelines/models.py
```diff
@@ -464,18 +464,20 @@ def delete_if_has_perm(self, *, principal: User):
     def last_version(self) -> "PipelineVersion":
         return self.versions.first()
 
+    @property
+    def new_template_versions(self):
+        if not self.source_template:
+            return []
+        last_version_from_template = self.versions.filter(
+            source_template_version__isnull=False
+        ).first()
```
nazarfil, line 473: In case the last version from template is null, should we safeguard and return null array?

nazarfil (approved):

hexa/pipelines/models.py
YolanFery: Definitely, thanks !

I wrongly assumed pipeline always have at least one version but we allow to delete all the version

qgerome: I'd put this method on the template directly in order to keep the pipeline code "clean" of the notion of templates (as much as we can) WDYT?

nazarfil (approved):

hexa/pipelines/models.py
YolanFery: Might be better, done !

hexa/pipelines/schema/queries.py
```diff
@@ -92,6 +92,18 @@ def resolve_pipeline_version(_, info, **kwargs):
...
+@pipelines_query.field("availableUpgradePipelineTemplateVersions")
+def resolve_available_pipeline_versions(_, info, **kwargs):
+    request: HttpRequest = info.context["request"]
+    try:
+        pipeline = Pipeline.objects.filter_for_user(request.user).get(
+            id=kwargs["pipeline_id"]
+        )
+    except Pipeline.DoesNotExist:
+        return []
+    return pipeline.new_template_versions
```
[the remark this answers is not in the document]
YolanFery: If you prefer, I don't mind

qgerome (commented):
LGTM!

PR #908, opened 2025-01-22
feat: handle default parameters values for pipeline template upgrades (HEXA-1165)

YolanFery:
When upgrading to a new version from the template, the default parameter values should be kept.

## Changes

- Remove unused config from template
- Only keep parameters config of native types
- During an upgrade keep the config from the last version of the pipeline by taking into account the change in parameters
- Extend a unit test to cover the case of non-native types and the case of filtering the config


hexa/pipelines/models.py
```diff
@@ -199,7 +199,7 @@ def validate_new_config(self, new_config: dict):
         for parameter in self.parameters:
             if (
                 parameter.get("required")
-                and parameter.get("defaul") is None
+                and parameter.get("default") is None
```
YolanFery, line 202: 6 month's old type

nazarfil: 🤦

hexa/pipeline_templates/schema/mutations.py
```diff
@@ -58,10 +58,9 @@ def resolve_create_pipeline_template_version(_, info, **kwargs):
         name=input.get("name"),
         code=input.get("code"),
         description=input.get("description"),
-        config=input.get("config"),
```
YolanFery, line 61: Note to myself : once merged, let's change the graphql schema and remove the config from the input parameters in both FE and BE (will require 2 separate PRs)

hexa/pipelines/models.py
```diff
@@ -360,17 +360,24 @@ def is_schedulable(self):
...
+        def remove_default(parameter):
+            parameter_without_default = parameter.copy()
+            if "default" in parameter:
+                del parameter_without_default["default"]
+            return parameter_without_default
+
         previous_config_from_overlapping_parameters = {}
         if self.last_version:
-            previous_parameters = self.last_version.parameters
+            previous_parameters = map(remove_default, self.last_version.parameters)
```
YolanFery, line 376: As shared by @nazarfil , this returns an iterator that will yield values when used with `in`, this is an issue, the expected behaviour would be to cast it

nazarfil (approved):

hexa/pipeline_templates/models.py
```diff
@@ -134,13 +136,29 @@ def create_pipeline_version(
     ) -> PipelineVersion:
         pipeline = pipeline or self._create_pipeline(workspace)
         source_version = self.source_pipeline_version
+        kept_source_parameters = list(
```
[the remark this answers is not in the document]
YolanFery: Done

hexa/pipeline_templates/models.py
```diff
@@ -77,6 +78,20 @@ def upgrade_pipeline(
             principal, pipeline.workspace, pipeline
         )
 
+    def new_versions(self, pipeline: Pipeline) -> list["PipelineTemplateVersion"]:
+        """Return the versions of the template that are newer than the last version of the pipeline"""
+        if pipeline.source_template != self:
+            raise ValueError("The specified pipeline is not from this template")
+
+        last_version_from_template = pipeline.versions.filter(
```
qgerome, line 86: nitpicking: to ease my understanding of the clause, I would have use this wording. I do not think it's worth it to change it. 
```suggestion
        last_version_created_from_template = pipeline.versions.filter(
```

hexa/pipelines/graphql/schema.graphql
```diff
@@ -248,7 +248,7 @@ type Pipeline {
   notebookPath: String  # The path to the notebook file for the pipeline (only filled if the pipeline is a notebook).
   template: PipelineTemplate  # The template associated with the pipeline.
   sourceTemplate: PipelineTemplate  # The source template associated with the pipeline.
-  newTemplateVersionAvailable: Boolean!  # Indicates if a new template version is available for the pipeline.
+  newTemplateVersions: [PipelineTemplateVersion!]  # The new template versions available for the pipeline.
```
qgerome, line 251:
```suggestion
  newTemplateVersions: [PipelineTemplateVersion!]!  # The new template versions available for the pipeline.
```

hexa/pipeline_templates/models.py
```diff
@@ -77,6 +78,20 @@ def upgrade_pipeline(
             principal, pipeline.workspace, pipeline
         )
 
+    def new_versions(self, pipeline: Pipeline) -> list["PipelineTemplateVersion"]:
```
qgerome, line 81: I have the impression that adding a verb in front of `new_versions` would clarify how to use this

```suggestion
    def get_new_versions(self, pipeline: Pipeline) -> list["PipelineTemplateVersion"]:
```
?

hexa/pipelines/models.py
```diff
@@ -465,17 +474,8 @@ def last_version(self) -> "PipelineVersion":
...
-            and hasattr(self.last_version, "source_template_version")
-            and self.last_version.source_template_version
-        ):
-            return (
-                self.source_template.last_version.created_at
-                > self.last_version.source_template_version.created_at
-            )
-        return False
+    def new_template_versions(self):
+        return self.source_template.new_versions(self) if self.source_template else []
```
qgerome, line 478: Looking at this sole line, My first idea would be to implement this on the query set of `PipelineTemplateVersion` as `PipelineTemplateVersion.objects.get_updates_for(pipeline:Pipeline)`

qgerome: And then I would maybe re-evaluate the need for this property.

qgerome: I'm ok with this anyway.

hexa/pipelines/models.py
```diff
@@ -518,14 +518,13 @@ def merge_pipeline_config(
         merged_config = {**cleaned_pipeline_version_config, **cleaned_provided_config}
         return merged_config
 
-    def get_or_create_template(self, name, code, description, config):
+    def get_or_create_template(self, name, code, description):
```
qgerome, line 521: Could you type it please ?

hexa/pipelines/schema/mutations.py
```diff
@@ -604,7 +604,7 @@ def resolve_upgrade_pipeline_version_from_template(_, info, **kwargs):
         return {"success": False, "errors": ["PERMISSION_DENIED"]}
     if not pipeline.source_template:
         return {"success": False, "errors": ["PIPELINE_NOT_FROM_TEMPLATE"]}
-    if not pipeline.is_new_template_version_available:
+    if len(pipeline.new_template_versions) == 0:
```
qgerome, line 607: That's also a reason why I'd tend to use a query set. If the list is more than a few hundreds, it can start to be very slow. I agree that's VERY unlikely :p

hexa/pipelines/models.py
YolanFery: Done but the return type can not be typed because of circular imports

hexa/pipelines/schema/mutations.py
YolanFery: Good point, thanks for the suggestion again

hexa/pipeline_templates/models.py
```diff
@@ -86,7 +87,22 @@ def __str__(self):
...
+        """Return the versions of the template that are newer than the last version of the pipeline"""
+        if pipeline.source_template is None:
+            return self.none()
+
+        last_version_created_from_template = pipeline.versions.filter(
+            source_template_version__isnull=False
+        ).first()
+
+        if not last_version_created_from_template:
+            return self.none()
```
qgerome, line 100: nice

qgerome (approved):

qgerome:
LGTM!

PR #909, opened 2025-01-22
fix: Support deleted source pipeline (HEXA-1169)

YolanFery:
Support deleted source pipeline

This change should have no effect now because this method is only used when creating a new template or template version, but this is only proposed when the source pipeline is not deleted. So at this point it is for correctness or avoiding future issues.


qgerome (approved):

qgerome (commented):
Could you add a comment above to explain the use case ?

PR #913, opened 2025-01-23
chore: Change upgrade() to upgrade_pipeline() to be more explicit and add docstrings on a few PipelineTemplate methods

qgerome:
Title is self-explanatory


YolanFery (approved):
Thanks

PR #915, opened 2025-01-27
chore(Datasets): Rename DatasetVersion.description to DatasetVersion.changelog

qgerome:
PATHWAYS-308

## Changes

- I renamed `DatasetVersion.description`to `DatasetVersion.changelog` for better discoverability
- I kept the description in the schema for now even though I was not able to find usage of the field.

## How/what to test

Both changelog & description can be get on the object. User can still create dataset versions.


nazarfil:
Tested, works in  FE and could query both changelog and description in GraphQL api

nazarfil (approved):

PR #918, opened 2025-01-29
feat: Page showing templates and allow to delete them (HEXA-1129)

YolanFery:
Enable template deletion and improve searching and filtering the list of templates.

## Changes

- GraphQL for deleting template and filtering template by workspace
- Allow the `source_pipeline` and `source_pipeline_version` to be `null` to enable a soft delete of template and remove references to existing pipelines through a custom `delete` method
- Ensure deleting through admin panel is also deleting references
- Searching for `pipelineTemplates` should also look for the search query in the `description`
- Unit test to cover the deleting through the GraphQL layer

## How/what to test

Adding and deleting templates. Deleting a template should allow to recreate a new template from the same pipeline.


hexa/pipeline_templates/graphql/schema.graphql
```diff
@@ -90,6 +115,7 @@ type PipelineTemplate {
   versions: [PipelineTemplateVersion!]  # The list of versions of the pipeline template.
   currentVersion: PipelineTemplateVersion  # The current version of the pipeline template.
   sourcePipeline: Pipeline  # The source pipeline of the pipeline template.
+  canDelete: Boolean # Indicates if the pipeline template can be deleted bu the requesting user.
```
qgerome, line 118: Could you please add the boilerplate code to handle `permissions` type ?

hexa/pipeline_templates/models.py
```diff
@@ -62,6 +71,17 @@ def create_version(self, source_pipeline_version: "PipelineVersion"):
         )
         return template_version
 
+    def delete(self):
+        # Delete references to the source pipeline and version
+        if self.source_pipeline:
```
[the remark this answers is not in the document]
YolanFery: Yes it is because of the soft delete, referenced are still existing

hexa/pipeline_templates/graphql/schema.graphql
YolanFery: Indeed, it's better and consistent with this pattern

Done

nazarfil (approved):

PR #921, opened 2025-02-07
feat: Management pages for templates (HEXA-1129, HEXA-1130)

YolanFery:
Allow edit and deleting of pipeline templates and versions.
Rework the deletion model to ensure no loss of data and smooth objects management when deletion occur.

## Changes

- GraphQL schemas
- Page template versions
- Handle deleted template and allow to restore it
- Updated unit test to cover few cases

## How/what to test

Might be good to test with the front-end PR (https://github.com/BLSQ/openhexa-frontend/pull/1070) when ready. To test objects updates and deletions


hexa/pipeline_templates/models.py
```diff
@@ -139,11 +166,14 @@ class Meta:
...
     changelog = models.TextField(blank=True, null=True)
+    user = models.ForeignKey(
+        "user_management.User", null=True, on_delete=models.SET_NULL
+    )
     template = models.ForeignKey(
         PipelineTemplate, on_delete=models.CASCADE, related_name="versions"
     )
     source_pipeline_version = models.OneToOneField(
-        PipelineVersion, on_delete=models.RESTRICT, related_name="template_version"
+        PipelineVersion, on_delete=models.CASCADE, related_name="template_version"
```
YolanFery, line 176: Each time a `PipelineVersion` will be deleted we will just delete the `PipelineTemplateVersion`, so that we ensure that `PipelineVersion` can be deleted freely and `PipelineTemplateVersion` always have a source `PipelineVersion`

qgerome (approved):

nazarfil (approved):

PR #930, opened 2025-02-17
feat!: Allow creation of a pipeline with a duplicate name  (HEXA-1170)

YolanFery:
Deprecate the pipeline code input and auto generate the pipeline code to allow duplicate pipelines

## Changes

- Introduce `create_with_unique_code` to generate a unique code
- Use the new method in the relevant places
- Add a descriptive error message to duplicate creation to warn of the existence of `create_with_unique_code`
- Add a test to cover the creation of duplicate pipelines
- Update tests

## How/what to test

- It is possible to create multiple pipelines form the same template, the pipeline codes are auto generated


hexa/pipelines/models.py
```diff
@@ -267,6 +269,20 @@ def parse_log_level(cls, value):
...
+class PipelineManager(DefaultSoftDeletedManager.from_queryset(PipelineQuerySet)):
+    def _create_unique_code(self, name: str, workspace: Workspace):
+        suffix = ""
+        while True:
+            code = slugify(name[: 255 - len(suffix)] + suffix)
+            if not super().filter(workspace=workspace, code=code).exists():
+                return code
+            suffix = "-" + secrets.token_hex(3)
+
+    def create_with_unique_code(self, name: str, workspace: Workspace, **kwargs):
```
qgerome, line 281: I'd prefer have the same api that in the other models: `create_if_has_perm`
The fact that it's going to be created with a unique code is a implementation detail of the lower level.

hexa/pipelines/schema/mutations.py
```diff
@@ -64,7 +60,9 @@ def resolve_create_pipeline(_, info, **kwargs):
         else:
             data["type"] = PipelineType.ZIPFILE
 
-        pipeline = Pipeline.objects.create(**data)
+        pipeline = Pipeline.objects.create_with_unique_code(
+            input.get("name"), workspace, **data
```
qgerome, line 64: Nitpick: I'd either put the `"name": input["name"]` in the data dict or I'd pass `name=input["name"]` here.

hexa/pipelines/tests/test_models.py
```diff
@@ -406,3 +406,22 @@ def test_new_template_version(self):
...
+            workspace=workspace,
+        )
+        pipeline2 = Pipeline.objects.create_with_unique_code(
+            name="Test Pipeline",
+            workspace=workspace,
+        )
+
+        self.assertNotEqual(pipeline1.code, pipeline2.code)
+        self.assertEqual(pipeline1.code, "test-pipeline")
+        self.assertTrue(pipeline2.code.startswith("test-pipeline-"))
```
qgerome, line 427: Would be nice to mock the helper that returns the suffix in order to check the complete code.

qgerome (commented):
Looks nice !

qgerome (approved):

hexa/pipelines/models.py
YolanFery: Not totally agreeing with that, I actually want to surface and expose that the code used is not the one passed in the `kwargs` but is auto generated. I found that `create_if_has_perm ` could mislead in thinking that nothing is changed.

hexa/pipelines/schema/mutations.py
YolanFery: Changed

hexa/pipelines/tests/test_models.py
YolanFery: A bit of implementation detail but we can. I would have preferred seeding but I have not found a straightforward way to do it

hexa/pipelines/schema/queries.py
```diff
@@ -32,6 +32,12 @@ def resolve_pipelines(_, info, **kwargs):
     else:
         qs = Pipeline.objects.filter_for_user(request.user).order_by("name", "id")
 
+    name_filter = kwargs.get("name", None)
+    if name_filter:
+        matching_qs = list(qs.filter(name__icontains=name_filter))
+        non_matching_qs = list(qs.exclude(name__icontains=name_filter))
+        qs = matching_qs + non_matching_qs
```
YolanFery, line 39: Used by the CLI to handle the paging gracefully

hexa/pipelines/tests/test_models.py
qgerome: Indeed it would have been better.

hexa/pipelines/models.py
qgerome: As far as I can tell the `code` is not passed anymore. It's auto generated and has the same implementation as the other slug|code of the models. no ?

hexa/pipelines/schema/queries.py
qgerome: I do not understand the purpose of this operation. It cancels the filtering on the name (that was added in this PR). Could you give me more details on what you want to achieve ?

YolanFery: In the CLI, we want to show the list of pipelines (https://github.com/BLSQ/openhexa-sdk-python/blob/HEXA-1170-allow-creation-of-a-pipeline-from-a-template-with-a-different-name/openhexa/cli/cli.py#L375 from https://github.com/BLSQ/openhexa-sdk-python/pull/227/files). Since the `code` is optional and will be deprecated, we want to list the pipelines by name. This optional filter is used for that.

hexa/pipelines/models.py
YolanFery: Yes that's exact. 

But to me, if I see a model that has a code and pass `PipelineTemplate.objects.create_if_has_perm(**{code:someCode, ....})` I expect the code to be `someCode` and not be replaced under the hood. 

BUT actually, since we use `super().create(name=name, workspace=workspace, code=code, **kwargs)` the `code` from `kwargs` will override the explicit `code=`, so that could be acceptable to me even thought the intent and logic is hidden

hexa/pipelines/schema/queries.py
nazarfil: interesting, so both filtered and unfiltered are returned. And filtered are positioned first in the array?

nazarfil (approved):

hexa/pipelines/schema/queries.py
YolanFery: Indeed, relevant matches are returned first 
Maybe we can find better naming to make it more explicit

qgerome (approved):

PR #931, opened 2025-02-17
Embedded dashboards from Superset made right!

qgerome:
Adds a way to embed dashboards in a more serious way than the POC made on Vercel...


YolanFery (approved):
Clean model, nice !

PR #933, opened 2025-02-21
feat:  Queue and send analytics asynchronously to MixPanel (HEXA-1191)

YolanFery:
We are facing unreliable MixPanel endpoint, causing delays in requests because the analytics tracking is happening synchronously. 
Solve the problem by treating analytics asynchronously. 

## Changes

- Install and use `AsyncBufferedConsumer`
- Sanity unit tests check 

## How/what to test

Events are stacked and sent eventually 
⚠️ the flush is only happening when a new event is created


hexa/analytics/api.py
```diff
@@ -10,11 +11,11 @@
 if settings.MIXPANEL_TOKEN:
     mixpanel = Mixpanel(
         token=settings.MIXPANEL_TOKEN,
-        consumer=Consumer(
-            retry_limit=0,
+        consumer=AsyncBufferedConsumer(  # See AsyncMixpanelTest for expected behavior
```
nazarfil, line 14: Thanks, clear

PR #943, opened 2025-03-14
fix: Datasets filesnot profiled (HEXA-1219)

YolanFery:
The profiling fails because there is a number in one of the column headers. Changing it to a string is working.I will make a fix to convert any header number into a string to avoid the issue

## Changes

- Cover with test
- Format to str before encoding


hexa/analytics/api.py
```diff
@@ -1,14 +1,15 @@
...
 from mixpanel import Mixpanel
-from mixpanel_async import AsyncBufferedConsumer
 from sentry_sdk import capture_exception
 from ua_parser import user_agent_parser
 
 from hexa.user_management.models import AnonymousUser, User
 
 mixpanel = None
 if settings.MIXPANEL_TOKEN:
+    from mixpanel_async import AsyncBufferedConsumer
```
YolanFery, line 11: Allows me to test locally and reduce import to only strictly needed

PR #945, opened 2025-03-17
feat(templates): Ask for deletion confirmation, more details when unable to create template (HEXA-1209)

YolanFery:
Address feedback from Mariano in https://whimsical.com/testing-notes-march-2025-Dj7QepTsa731LwFwjTqSNH


hexa/pipelines/graphql/schema.graphql
```diff
@@ -198,7 +198,25 @@ type PipelinePermissions {
...
   createVersion: Boolean!  # Indicates if a new version of the pipeline can be created.
   stopPipeline: Boolean!  # Indicates if the pipeline can be stopped.
-  createTemplateVersion: Boolean!  # Indicates if a new template version can be created.
+  createTemplateVersion: CreateTemplateVersionPermission!  # Indicates if a new template version can be created.
+}
+
+"""
+Represents the permission details for creating a template version.
+"""
+type CreateTemplateVersionPermission {
```
qgerome, line 207: This creates a precedent and need to be discussed first. I'd personally prefer to keep the permissions of all objects as a simple map of booleans representing a simple state.

qgerome: I had a second thought on this. I think it adds value and allow us to keep the context in the API layer (no need to guess the reason in the frontend). I used to give that piece of info from a field in the base object but it makes more sense like this IMO.
I'd expose that as a reason|explanation instead of an error

YolanFery: Yes the intent is to not duplicate logic. I agree that it complexifies but it a needed complexity.

I agree with the naming. I will change it

qgerome (approved):

PR #948, opened 2025-03-18
Superset improvements

qgerome:
## Changes

- "View on site" link in the Django admin redirects to the embedded dashboard on app.openhexa.org and not api.openhexa.org (it's now using `NEW_FRONTEND_DOMAIN`)
- Add inline panel to add dashboards directly from the admin page of a superset instance
- Add temporary endpoint to support getting dashboard by their external id (this is what's used by the former system)

## How/what to test

- Click on "view on site", you should arrive on http://localhost:3000
- Test the endpoint `/superset/dashboard/external/<external_id>`

### DIFF
--- a/hexa/superset/admin.py
+++ b/hexa/superset/admin.py
@@ -3,11 +3,21 @@ from django.contrib import admin
 from .models import SupersetDashboard, SupersetInstance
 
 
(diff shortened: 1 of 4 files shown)


YolanFery (approved):
Tested the admin actions and clicking on the View site button directing to the correct url

PR #952, opened 2025-03-24
Start to work on MJML

qgerome:
Changes email templating to mjml for both
- Invitation email
- Pipeline result email
Adds 
- images to display in the template 

Screenshot 2025-03-26 at 16 46 48
Screenshot 2025-03-26 at 16 47 26


nazarfil:
The result looks fine both in txt and email, only need to store the images in a CDN and fix tests, gonna do tomorrow

nazarfil (approved):

PR #971, opened 2025-04-03
fix: align markdown configuration with live demo of MDXEditor

YolanFery:
Align our configuration of markdown with https://mdxeditor.dev/editor/demo
-> Adding inline or multilines code block is only possible through the toolbar or source mode view

## Changes

- Remove filtering of backticks
- Add the diff/source views with toolbar switch

## How/what to test

The editor is working espacially with code blocks and backticks

### DIFF
--- a/frontend/src/core/components/MarkdownEditor/MarkdownEditor.tsx
+++ b/frontend/src/core/components/MarkdownEditor/MarkdownEditor.tsx
@@ -7,6 +7,8 @@ import {
   codeMirrorPlugin,
   CodeToggle,
   ConditionalContents,
+  diffSourcePlugin,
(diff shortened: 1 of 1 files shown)


qgerome (commented):
I fail to see why having the diff plugin enabled should resolve our problem wit backticks 😜

YolanFery:
> I fail to see why having the diff plugin enabled should resolve our problem wit backticks 😜

The diff plugin includes a source mode where backtick can be introduced unescaped :

First backtick is entered in the editing mode, the second is in source mode : 
Screenshot 2025-04-07 at 09 20 22

qgerome (approved):

PR #972, opened 2025-04-04
Hexa 1205 feat add email tracking to mixpanel

nazarfil:
Adds events to track email invitation status
Modifies pre-commit to work for backend
Moves from deprecated imghdr library to filetype


.github/workflows/backend-ci.yml
```diff
@@ -10,16 +10,16 @@ on:
 
 jobs:
   pre-commit:
```
YolanFery, line 12: As said we can rename this to `lint` for instance

.pre-commit-config.yaml
```diff
@@ -1,13 +1,20 @@
 repos:
 - repo: local
   hooks:
-    - id: run-backend-pre-commit
-      name: Run Backend Pre-Commit
-      entry: ./backend/script/pre-commit.sh
```
YolanFery, line 6: Let's delete the file `./backend/script/pre-commit.sh` then

backend/hexa/analytics/api.py
```diff
@@ -72,6 +75,41 @@ def track(
...
+def track_invitation(
+    invitation: WorkspaceInvitation,
+    event: str,
+    properties: dict = {},
+):
+    """Track event to mixpanel.
+
+    Parameters
+    ----------
+    event : str
```
YolanFery, line 87: [Minor] You are missing the invitation parameter in the docstring

backend/hexa/workspaces/utils.py
```diff
@@ -30,6 +31,8 @@ def send_workspace_invitation_email(
             )
             action_url = f"{settings.NEW_FRONTEND_DOMAIN}/register?{urlencode({'email': invitation.email, 'token': token})}"
 
+        track_invitation(invitation, "emails.invite_sent")
```
YolanFery, line 34: I would put it after the `send_mail` so that if the `send_mail` fails we don't have analytics for it

backend/hexa/user_management/schema.py
```diff
@@ -347,6 +347,7 @@ def resolve_register(_, info, **kwargs):
         invitation = WorkspaceInvitation.objects.get_by_token(
             token=mutation_input["invitation_token"]
         )
+        track_invitation(invitation, "emails.registration_landed")
```
YolanFery, line 350: WDYT of putting it after
```
if invitation.status != WorkspaceInvitationStatus.PENDING:
            return {"success": False, "errors": ["INVALID_TOKEN"]}
```
to only track valid invitations ?

backend/hexa/analytics/api.py
```diff
@@ -72,6 +75,41 @@ def track(
...
+        {
+            "$timestamp": datetime.utcnow().isoformat(),
+            "$workspace_id": invitation.workspace.id,
+            "$invitee_email": invitation.email,
+            "$invitee_role": invitation.role,
+        }
+    )
+    try:
+        mixpanel.track(
+            distinct_id=invitation,
```
YolanFery, line 105: Not totally sure about this line, something like `distinct_id=str(invitation.id)` would like more correct to me but that's just a guess at this point

YolanFery (commented):
Thanks for working out the pre-commit situation

backend/hexa/user_management/schema.py
nazarfil: In the desired events doc, we aim to make the difference between event that landed to users mailbox and they clicked on it, and events that succeeded. For example if an invitation has a token that expired, it would reach this even but not the success, and we could track it. 
https://docs.google.com/spreadsheets/d/1pDBXPnTL7VlQEfcCwVifb67gBZsDeVfbz1PN8JmFt70/edit?gid=1966580448#gid=1966580448

backend/hexa/analytics/api.py
```diff
@@ -72,6 +75,43 @@ def track(
         capture_exception(e)
 
 
+def track_invitation(
```
qgerome, line 78: This function should not be in the hexa/analytics/api. You can leave it where it's used.

backend/hexa/analytics/api.py
```diff
@@ -72,6 +75,43 @@ def track(
...
+    properties : dict
+       A dictionary holding the event properties
+    """
+    if mixpanel is None:
+        return
+
+    properties.update(
+        {
+            "$timestamp": datetime.utcnow().isoformat(),
+            "$workspace_id": invitation.workspace.id,
```
qgerome, line 100: Why do you prefix the properties by $ ? It usually used to fill in reserved properties no ?

backend/hexa/user_management/schema.py
qgerome: WDYT of adding the state of the invitation on the event sent to Mixpanel ? It allows us to make the distinction between the initial flow and the one in error.

qgerome (changes requested):

backend/hexa/analytics/api.py
nazarfil: You are right these, dollar sign is only for default properties and events

backend/hexa/analytics/api.py
nazarfil: It is used in user_management/schema and workspace/utls, doesn't it make sense to gather tracking in the analytics folder?

backend/hexa/user_management/schema.py
nazarfil: Yes, I can add status. I assumed that the event name is enough to track the state. But it is easy to add the status, and it will for sure help with tracing!

backend/hexa/analytics/api.py
qgerome: Since you are handling 3 different events from the same function, I'm wondering if it's even useful to put that inside a function. Each event call is a slight wrapped track_event call. I'd just use the track_event function then.

nazarfil: Ok I see, I've changed to use track method instead

nazarfil:
I have move the imghdr library to use filetype, as it is deprecated from 3.13 and breaks the build

qgerome (approved):

PR #994, opened 2025-04-08
fix: global docker compose for BE and FE

YolanFery:


docker-compose.yaml
```diff
@@ -115,6 +113,9 @@ services:
     restart: unless-stopped
     depends_on:
       - app
+    volumes:
+      - ./.env:/.env
+      - ./backend:/backend
```
qgerome, line 118: Nice

qgerome (approved):

PR #1027, opened 2025-04-15
HEXA-1243 fix(dhis2): keep selected ids and labels from previous query

nazarfil:
Adds cache to fix selected items disappearing

(Dont take opaque dialog into account, thats another lib update) 

https://github.com/user-attachments/assets/1222dc2a-2f1c-4156-a14f-f410a74ca1d4

### DIFF
--- a/frontend/src/workspaces/features/RunPipelineDialog/DHIS2Widget.tsx
+++ b/frontend/src/workspaces/features/RunPipelineDialog/DHIS2Widget.tsx
@@ -203,23 +203,27 @@ const DHIS2Widget = ({
   );
 
 
-  const selectedObjects = useMemo(() => {
-    const ids = ensureArray(form.formData[parameter.code]);
+const selectedObjects = useMemo(() => {
+  const ids = ensureArray(form.formData[parameter.code]);
 
-    return parameter.multiple
-      ? ids.map((id: string) =>
-          options.find((item) => item.id === id) ||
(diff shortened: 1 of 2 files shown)


frontend/src/workspaces/features/RunPipelineDialog/DHIS2Widget.tsx
```diff
@@ -168,40 +185,44 @@ const DHIS2Widget = ({
...
-        id: form.formData[parameter.code],
-        label: t("Unknown ID: {{id}}", { id: form.formData[parameter.code] }),
-      }
-    );
+        )
+      : options.find((item) => item.id === ids[0]) ||
+        cachedSelectionsRef.current.get(ids[0]) || {
+          id: ids[0],
+          label: t("Unknown ID: {{id}}", { id: ids[0] }),
+        };
```
YolanFery, line 221: [Minor] 

The logic for the multiple or not is the same, it can be extracted into a function, like for instance :

```
const ids = ensureArray(form.formData[parameter.code]);
const selectObject = (id) => 
         options.find((item) => item.id === id) ||
          cachedSelectionsRef.current.get(id) || {
            id,
            label: t("Unknown ID: {{id}}", { id }),
          }; 
return parameter.multiple
      ? ids.map(selectObject)
      : selectObject(ids[0])
```

YolanFery (approved):
The result looks nice !

PR #1030, opened 2025-04-16
feat: search bar (HEXA-1232, HEXA-1246)

YolanFery:
Frontend and Backend for enabling a Search bar allowing to search artefacts in all the workspaces that the user is member of.

## Changes

- Implement filters in the model to search by workspace
- Controller for each resource type ( -> this allow parallel loading and showing granular loading status)
- Frontend button and shortcuts to open/close overlay and navigate results
- Show a filter by workspace section to allow more granular search
- Show combined results and results by type
- Handle paging

## How/what to test

Searching for artefacts using keyboard shortcuts.

## Screenshots / screencast
Button : 
Screenshot 2025-04-28 at 09 50 54
Overlay : 
Screenshot 2025-04-28 at 09 51 10
Results and filters :
Screenshot 2025-04-28 at 09 51 20


YolanFery:
@qgerome @nazarfil let me know if you want me to split this PR by meaningful commits (and define the meaning) or any other split (multiple PRs)
Happy to help simplifying the review

qgerome:
Screenshot 2025-04-28 at 13 51 14
Could you add more Y padding and maybe an explanation ?

Screenshot 2025-04-28 at 14 44 21
I would keep the search input on top of everything, it feels weird to have more options appearing above it (plus with that extra padding space).

The workspaces filter could be a simple combobox IMO.
I'm not a fan of the "Select all" & "Clear all" buttons on top. It's not clear that they are linked to the workspaces list.

I'd either remove the loading state per type or always take the space to avoid the flickering.

qgerome:
Regarding the tab bar, I'd prefer a simple approach such as the one in Whimsical. I'm not sure it makes sense to display the count of each type.

qgerome:
When I do a search in dev on "covid" I do not have the demo workspace's files in the result :/ (tested with "covid", "*covid*)

backend/hexa/pipeline_templates/graphql/schema.graphql
```diff
@@ -164,6 +164,8 @@ type PipelineTemplate {
   currentVersion: PipelineTemplateVersion  # The current version of the pipeline template.
   sourcePipeline: Pipeline  # The source pipeline of the pipeline template.
   permissions: PipelineTemplatePermissions!  # The permissions for the pipeline template.
+  workspace: Workspace!  # The workspace associated with the pipeline template.
```
qgerome, line 167: The `workspace` can be null on a template

backend/hexa/pipeline_templates/graphql/schema.graphql
```diff
@@ -164,6 +164,8 @@ type PipelineTemplate {
...
+  updatedAt: DateTime  # The last updated date of the pipeline template.
```
qgerome, line 168: It's never `null`. At the creation it is set to `now()`

backend/hexa/search/graphql/schema.graphql
```diff
@@ -0,0 +1,87 @@
...
+}
+
+type FileResultPage {
+  pageNumber: Int!
+  totalPages: Int!
+  totalItems: Int!
+  items: [FileResult!]!
+}
+
+extend type Query {
```
qgerome, line 81: IMO having a `search(query: String, types: [], workspaces:[String!])` is easier. I think it should be the responsability of the user to merge the results.

frontend/src/core/components/forms/Input/Input.tsx
```diff
@@ -15,6 +16,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>((props, ref) => {
     error,
     defaultValue,
     className,
+    classNameOverrides,
```
qgerome, line 19: What's different from `className` that is also on the `input` ?

qgerome (changes requested):

frontend/src/core/components/forms/Input/Input.tsx
YolanFery: This ensures that it overrides the opinionated classes in the input component. Do we want the className to also override ?

backend/hexa/search/graphql/schema.graphql
YolanFery: It is currently the case, the client has the responsibility to merge the results. I didn't take the unified endpoint route to have precise output types and allow for parallel loading/rendering (files search might take more time than the rest for instance)

backend/hexa/pipeline_templates/graphql/schema.graphql
YolanFery: Good catch, thanks. Indeed, when a workspace is deleted, associated templates will have a null reference

YolanFery: Should be fixed in next commits

backend/hexa/pipeline_templates/graphql/schema.graphql
YolanFery: Good catch, thanks. Should be fixed in next commits

frontend/src/workspaces/layouts/WorkspaceLayout/Sidebar.tsx
```diff
@@ -161,6 +193,35 @@ const Sidebar = (props: SidebarProps) => {
   return (
     <div className={clsx("relative z-20 flex h-full flex-col", className)}>
       <div className="flex h-full grow flex-col border-r border-gray-200 bg-gray-800">
+        <SpotlightSearch
+          key={isSearchOpen ? Date.now() : null}
```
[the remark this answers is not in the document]
YolanFery: Good idea, thanks.
Done

backend/hexa/datasets/schema/types.py
```diff
@@ -120,6 +120,11 @@ def resolve_dataset_updated_at(obj: Dataset, info, **kwargs):
     )
 
 
+@dataset_object.field("workspace")
```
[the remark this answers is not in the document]
YolanFery: You are right. I removed it in next commits.
I probably got confused by another issue

YolanFery:
> Regarding the tab bar, I'd prefer a simple approach such as the one in Whimsical. I'm not sure it makes sense to display the count of each type.

I have a different opinion here. I think that the count of each type provides more context to the user (the query is too restrictive, the query is matching on some types with having to switch tabs,...). It also makes the feature appear more polished IMO. The search feature in Github gives the count as well

YolanFery:
> When I do a search in dev on "covid" I do not have the demo workspace's files in the result :/ (tested with "covid", "_covid_)
> 
> EDIT: I do have it if I only select one workspace and not all.

Thanks for the details, I will work on fixing this bug

YolanFery:
> Screenshot 2025-04-28 at 13 51 14 Could you add more Y padding and maybe an explanation ? Screenshot 2025-04-28 at 14 44 21 I would keep the search input on top of everything, it feels weird to have more options appearing above it (plus with that extra padding space).
> The workspaces filter could be a simple combobox IMO. I'm not a fan of the "Select all" & "Clear all" buttons on top. It's not clear that they are linked to the workspaces list.
> 
> I'd either remove the loading state per type or always take the space to avoid the flickering.

Thanks for the feedback.

- I will add some Y padding.
- I can try to have the workspace filter below.
- The workspace filter is a multicombobox when the number of workspaces is > 10.
- I will remove "Select all" & "Clear all" buttons.
- I will always take the space to avoid the flickering.

YolanFery:
> > When I do a search in dev on "covid" I do not have the demo workspace's files in the result :/ (tested with "covid", "_covid_)
> > EDIT: I do have it if I only select one workspace and not all.
> 
> Thanks for the details, I will work on fixing this bug

It is actually not a bug, I think it is because you only have created one of the required buckets in GCS for the multiple workspaces, one of the query to GCS fails, failing the overall request, creating a bucket per workspace should make it work (or only selecting workspaces with an existing bucket)

But, unrelated, there was another issue with the GCP integration, I corrected it to support match_glob correctly and support directory searching

nazarfil (approved):

qgerome (approved):
Nice ! You need to rebase this PR

backend/hexa/search/schema/queries.py
```diff
@@ -0,0 +1,118 @@
...
+from hexa.datasets.models import Dataset
+from hexa.files import storage
+from hexa.pipeline_templates.models import PipelineTemplate
+from hexa.pipelines.models import Pipeline
+from hexa.workspaces.models import Workspace
+
+search_query = QueryType()
+
+
+def apply_scored_search(queryset: QuerySet, fields: list[str], query: str):
```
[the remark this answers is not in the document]
YolanFery: > So the decision to go with scoring is to keep the implementation consistent, and use the simplest out of the box solution?

Kind of, yes. I wanted a simple unified scoring mechanism that would work across all types of assets. Also I wanted to have control.
Control was lacking with `SearchVector`/`pg_trgm` : I tried and got troubles with partial matches, upper/lower case matches,...
So I decided to start simple and this part can be refined later independently based on feedback and performances. I didn't want to over-engineer this part and rather wait for more informed decision

PR #1036, opened 2025-04-17
fix(DHIS2): organisation_unit_levels did not support filters parameter OPENHEXA-1H1

qgerome:
Upgrading the openhexa-toolbox library to pass extra params as kwargs


nazarfil:
It is working in 2.0.3  toolbox, I have fixed it - we need to make a release, but first we had to revert the tailwindcss library version upgrade, as it was making each dialog translucent

nazarfil (approved):

PR #1072, opened 2025-05-07
chore: generate aggregated graphql file (HEXA-1238)

YolanFery:
We need an aggregated file for https://the-guild.dev/graphql/inspector/docs/products/action since glob are not working for this github action https://github.com/graphql-hive/graphql-inspector/issues/1978


YolanFery:
Merging it since it is need to have a first version in the repo to have a reference for https://the-guild.dev/graphql/inspector/docs/products/action

PR #1094, opened 2025-05-12
chore: Expose globally enabled feature flags to users via GraphQL & remove FeatureFlag.config from GraphQL (not used)

qgerome:
Globally enabled features were not sent to the frontend via `me.features`.
The new implementation returns all the features flags (global & user enabled).
The FeatureFlag.config has been removed from the API as it has never been used. 

## How/what to test

- Test with a feature with `force_activate = True`. It should be returned as a feature flag of the user

### DIFF
--- a/backend/hexa/user_management/schema.py
+++ b/backend/hexa/user_management/schema.py
@@ -34,7 +34,7 @@ from hexa.user_management.models import (
     AlreadyExists,
     CannotDelete,
     CannotDowngradeRole,
-    FeatureFlag,
+    Feature,
     Membership,
     Organization,
     Team,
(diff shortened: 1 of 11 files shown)


backend/hexa/user_management/schema.py
```diff
@@ -140,7 +129,13 @@ def resolve_me_features(_, info):
     principal: User = request.user
 
     if principal.is_authenticated:
-        return principal.featureflag_set.all()
+        return [
+            {
+                "code": feature.code,
+                "config": {},
+            }  # TODO: Remove the config field once the migration is done
```
nazarfil, line 136: The reason of leaving an empty config ={} and just deprecating graphQL interface, (instead of removing it completely from FeatureFlag model) is to give users some time to confirm it is not used during deprecation period?

nazarfil (approved):

backend/hexa/user_management/schema.py
qgerome: It's to not break the API directly indeed.

PR #1097, opened 2025-05-12
fix: Tailwind upgrade to > 4.1.4 is breaking the backdrop of dialog (HEXA-1258)

YolanFery:
After the upgrade to Tailwind > 4.1.4, the backdrop is overlapping the dialog content

### DIFF
--- a/frontend/src/core/components/Dialog/Dialog.tsx
+++ b/frontend/src/core/components/Dialog/Dialog.tsx
@@ -107,7 +107,7 @@ function Dialog(props: DialogProps) {
         transition
         className="fixed inset-0 bg-black/30 duration-300 backdrop-blur-xs ease-out data-[closed]:opacity-0"
       />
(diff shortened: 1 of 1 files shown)


frontend/src/core/components/Dialog/Dialog.tsx
```diff
@@ -107,7 +107,7 @@ function Dialog(props: DialogProps) {
         transition
         className="fixed inset-0 bg-black/30 duration-300 backdrop-blur-xs ease-out data-[closed]:opacity-0"
       />
-      <div className="h-screen px-4 pb-20 pt-4 text-center sm:block sm:p-0">
+      <div className="fixed inset-0 h-screen px-4 pb-20 pt-4 text-center sm:block sm:p-0">
```
YolanFery, line 110: This one is a bit mysterious to me but seems to make it work. Open to explanation/improvement

PR #1123, opened 2025-05-15
feat: Simplify flow of adding/inviting members to workspaces

bramj:
This PR simplifies the flow of adding users to a workspace, by splitting the "adding" and the "inviting".

The new behavior:

- Existing users are directly added to the workspace. They receive an email to notify they've been added to the workspace. No `WorkspaceInvitation` is created.
- For `New users`,   a `WorkspaceInvitation` is created and they receive an email to join the platform (as before). Upon registration, this invite is automatically accepted.

Also:
- A Django migration to accept all pending invitations for known users of the platform
- The user search is behind a feature flag `"users.search"`
- HEXA-1249: fixed some Mixpanel tracking

Tickets:
- https://bluesquare.atlassian.net/browse/HEXA-1092
- https://bluesquare.atlassian.net/browse/HEXA-1249

## How/what to test

- Invite a user with a new email, check the sent email, sign up using the invitation link and check if the invite is auto-accepted
- Invite (add) a user with an existing email, check the email, check if the user is now added
- Add the feature flag to your user and you can see the input field change to a "user picker"

## Screenshots / screencast
[add_invite_member-2025-05-15_11.48.58.webm](https://github.com/user-attachments/assets/46703ec1-5e6b-4114-9f5c-f786a0fdef1a)


frontend/src/workspaces/features/UserPicker/UserPicker.tsx
```diff
@@ -1,15 +1,21 @@
+import { gql } from "@apollo/client";
 import { useCallback, useState } from "react";
 import { ComboboxOption } from "@headlessui/react";
 import clsx from "clsx";
 import { useTranslation } from "next-i18next";
 
 import useDebounce from "core/hooks/useDebounce";
 import { Combobox } from "core/components/forms/Combobox";
-import { useGetUsersQuery } from "identity/graphql/queries.generated";
+import UserComponent from "core/features/User";
```
qgerome, line 9: We tend to import the component using the filename as name

frontend/src/workspaces/features/UserPicker/UserPicker.tsx
```diff
@@ -1,15 +1,21 @@
...
+
+import {
+  useGetUsersQuery,
+  UserPicker_UserFragment,
+} from "./UserPicker.generated";
 
 type UserPickerProps = {
   workspaceSlug: string;
-  form: any;
+  onChange(user: UserPicker_UserFragment | null): void;
```
qgerome, line 18: +
```
value: UserPicker_UserFragment | null
```

frontend/src/workspaces/features/InviteMemberDialog/InviteMemberDialog.tsx
```diff
@@ -26,7 +27,7 @@ type InviteMemberDialogProps = {
 };
 
 type Form = {
-  user: User | null;
+  user: UserPicker_UserFragment | null;
```
qgerome, line 30: Instead of relying on the presence of this fragment, I tend to use the type of the prop directly
```
user: React.ComponentProps<typeof UserPicker>["user"]
```

backend/hexa/user_management/schema.py
```diff
@@ -245,6 +251,35 @@ def resolve_organizations(*_):
...
+        if not request.user.has_perm("workspaces.manage_members", workspace):
+            raise PermissionDenied
+
+        users = User.objects.filter(
+            Q(email__icontains=query)
+            | Q(first_name__icontains=query)
+            | Q(last_name__icontains=query)
+        )
+
+        users = users.exclude(id__in=workspace.members.values_list("id", flat=True))
```
YolanFery, line 274: [Minor] I would exclude first before searching by text. Searching by text is costly. Also adding a small comment why we exclude can help understand quicker that we are excluding current members of the workspace

backend/hexa/user_management/schema.py
```diff
@@ -245,6 +251,35 @@ def resolve_organizations(*_):
...
+
+        return users.order_by("email")[:10]
+    except PermissionDenied:
+        return []
+    except Workspace.DoesNotExist:
+        return []
```
YolanFery, line 280: [Minor] usually we try to return some descriptive errors that we handle in the frontend as well, but that's not always the case

backend/hexa/user_management/tests/test_schema.py
```diff
@@ -1041,6 +1042,98 @@ def test_super_user_authorized_action(self):
...
+                    firstName
+                    lastName
+                }
+            }
+            """,
+            {"query": "", "workspaceSlug": workspace.slug},
+        )
+
+        users = r["data"]["users"]
+        self.assertEqual(User.objects.count(), 6)
```
YolanFery, line 1067: We only create one user in the test, why do we expect 6 ?

backend/hexa/user_management/tests/test_schema.py
```diff
@@ -1041,6 +1042,98 @@ def test_super_user_authorized_action(self):
...
+        self.assertEqual(len(users), 5)  # Finds all users except themselves
+        self.assertEqual(
+            {"email": self.USER_ADMIN.email, "firstName": "", "lastName": ""}, users[0]
```
YolanFery, line 1070: [Minor] I would test with non empty first name and last name instead of checking that they are empty which is not bringing a lot of value

backend/hexa/user_management/tests/test_schema.py
```diff
@@ -1041,6 +1042,98 @@ def test_super_user_authorized_action(self):
...
+            """,
+            {"query": "taylor", "workspaceSlug": workspace.slug},
+        )
+        self.assertEqual(len(r["data"]["users"]), 1)
+        self.assertEqual(
+            {"email": self.USER_TAYLOR.email, "firstName": "", "lastName": ""},
+            r["data"]["users"][0],
+        )
+
+    def test_search_users_invalid_workspace(self):
```
YolanFery, line 1099: [Minor] as said above, usually I would expect an error saying invalid workspace instead of empty results

backend/hexa/user_management/tests/test_schema.py
bramj: There are 5 more in the `setUp`, but I agree it's too magical, I'll add at least a comment

backend/hexa/user_management/schema.py
YolanFery: Indeed, performance might be the same

nazarfil (approved):

backend/hexa/user_management/schema.py
bramj: Double-checked and indeed since it's a simple AND query, Postgres generates the exact same query plan (`EXPLAIN ANALYZE`) regardless of the order of the clauses. I'm adding a comment to clarify.

YolanFery: Cool, thanks

backend/hexa/user_management/schema.py
YolanFery: Yes that's why I wrote that it's not always the case but didn't realize the pattern

Aligned, sounds good to me

PR #1126, opened 2025-05-21
fix: profiling and sampling issue with infinity (HEXA-1278)

YolanFery:
We have errors coming up when infinity is used in column numbers. But it's not supported well by postgresql and our statistics methods

## Changes

- Replace infinity by the equivalent string for sampling
- Replace infinity by 0 for profiling
- Update the profiling unit test

### DIFF
--- a/backend/hexa/datasets/queue.py
+++ b/backend/hexa/datasets/queue.py
@@ -116,7 +116,9 @@ def generate_profile(df: pd.DataFrame) -> list:
         constant_values = df.apply(lambda x: x.nunique() == 1).astype("bool").to_dict()
         count = df.count().to_dict()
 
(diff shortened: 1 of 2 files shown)


backend/hexa/datasets/queue.py
```diff
@@ -80,6 +80,9 @@ def generate_sample(
                 random_state=SAMPLING_SEED,
                 replace=True,
             )
+            sample = sample.replace(
+                {np.inf: "inf", -np.inf: "-inf"}
+            )  # We are not supporting Infinity as numbers
```
nazarfil, line 85: Makes sense that we don't :D, interesting how they do math with inf numbers

nazarfil (approved):

PR #1187, opened 2025-06-06
Hexa 1245 feat cleaner email invitation for existing

nazarfil:
- removes context from txt file
- removes content from existing email user
- removes attachments


bramj (approved):
Great, good to go! :+1:

PR #1210, opened 2025-06-12
fix: click outside search bar

YolanFery:
Clicking outside the search bar was not always working.
Because : 

1. The ref was not on a the proper div
2. Content with opacity-0 (used for animation) was taking the clicks


bramj (approved):
LGTM 👍 looks like you had fun 😅

PR #1237, opened 2025-07-01
fix: filter for user methods to support pipeline token for most methods (HEXA-1316 )

YolanFery:
Allow pipeline run user to filter for objects

## Reverted

For long we have used branching when overriding `filter_for_user` methods to support pipeline run user. 

It had 3 drawbacks : 
- It was not enforced, so the support for the PipelineUser was not automatic
- The code was duplicated 
- There was an import at runtime of a module in the method to perform the check. This might introduce a slight delay.

## Changes

By default if the user is a PipelineRunUser, try and compare the workspaces, if not possible, fallback to a method that can be overridden.
This ensures that we reduce code duplication and that we have some automation. 
This does NOT enforce implementation, but I think it's good enough to not perform extensive and risky refactoring

Added test to cover some models and the case were the specific for pipeline run user is not implemented

### DIFF
--- a/backend/hexa/core/models/base.py
+++ b/backend/hexa/core/models/base.py
@@ -75,14 +75,26 @@ class BaseQuerySet(models.QuerySet):
 
         1. Inactive users (including anonymous users) will get an empty queryset
         2. Superusers will get the full, unfiltered queryset (unless return_all_if_superuser is set to False)
-        3. Regular, authenticated users will get the queryset filtered using the query_object argument
+        3. Pipeline users will get the queryset filtered using the workspace of the pipeline run or any custom logic defined in _get_pipeline_run_user_workspace_query
+        4. Regular, authenticated users will get the queryset filtered using the query_object argument
         """
         if not user.is_authenticated:
             return self.none()
         elif return_all_if_superuser and user.is_superuser:
             return self.all()
+        elif user.__class__.__name__ == "PipelineRunUser":
+            workspace_query = self._get_pipeline_run_user_workspace_query(user)
+            return self.filter(workspace_query)
 
         return self.filter(query_object).distinct()
+
+    def _get_pipeline_run_user_workspace_query(self, user):
+        """Get workspace query for PipelineRunUser - override for custom logic"""
+        if hasattr(self.model, "workspace"):
+            return models.Q(workspace=user.pipeline_run.pipeline.workspace)
+        raise NotImplementedError(
+            f"Override _get_pipeline_run_user_workspace_query for {self.model} to allow pipeline run user access control"
+        )
 
     def prefetch_indexes(self):
         if not hasattr(self.model, "indexes"):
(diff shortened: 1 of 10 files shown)


backend/hexa/core/models/base.py
```diff
@@ -75,15 +75,27 @@ def _filter_for_user_and_query_object(
...
         2. Superusers will get the full, unfiltered queryset (unless return_all_if_superuser is set to False)
-        3. Regular, authenticated users will get the queryset filtered using the query_object argument
+        3. Pipeline users will get the queryset filtered using the workspace of the pipeline run or any custom logic defined in _get_pipeline_run_user_workspace_query
+        4. Regular, authenticated users will get the queryset filtered using the query_object argument
         """
         if not user.is_authenticated:
             return self.none()
         elif return_all_if_superuser and user.is_superuser:
             return self.all()
+        elif user.__class__.__name__ == "PipelineRunUser":
```
YolanFery, line 85: Use the class name instead of checking with `isinstance` which requires a runtime import of the class
I assume that it is faster ☺️

backend/hexa/core/tests/test_pipeline_user_filtering.py
```diff
@@ -0,0 +1,89 @@
...
+
+        class SomeModel:
+            objects = SomeModelQuerySet.as_manager()
+
+        with self.assertRaises(NotImplementedError) as context:
+            SomeModel.objects.filter_for_user(self.pipeline_user)
+            self.assertIn(
+                "Override _get_pipeline_run_user_workspace_query for SomeModel to allow pipeline run user access control",
+                str(context.exception),
+            )
```
YolanFery, line 89: Model what will happen when the `filter_for_user` is called but the specific method for pipeline run user is not implemented

backend/hexa/core/models/base.py
```diff
@@ -75,15 +75,27 @@ def _filter_for_user_and_query_object(
...
 
         return self.filter(query_object).distinct()
 
+    def _get_pipeline_run_user_workspace_query(self, user):
+        """Get workspace query for PipelineRunUser - override for custom logic"""
+        if hasattr(self.model, "workspace"):
+            return models.Q(workspace=user.pipeline_run.pipeline.workspace)
+        raise NotImplementedError(
+            f"Override _get_pipeline_run_user_workspace_query for {self.model} to allow pipeline run user access control"
+        )
```
bramj, line 97: Doesn't this mean that now when I try to call `filter_for_user()` with a `PipelineRunUser` on a Postgres connection (https://github.com/BLSQ/openhexa-app/blob/main/backend/hexa/plugins/connector_postgresql/models.py#L32-L36) I'll get a `NotImplementedError` because the `Database` model has no workspace field and no custom implementation?

bramj (commented):
Just one main question, the `raise NotImplementedError` seems to me like a change that potentially affects many places in the codebase, but my understanding is limited and I can be wrong.

YolanFery:
> Doesn't this mean that now when I try to call `filter_for_user()` with a `PipelineRunUser` on a Postgres connection (https://github.com/BLSQ/openhexa-app/blob/main/backend/hexa/plugins/connector_postgresql/models.py#L32-L36) I'll get a `NotImplementedError` because the `Database` model has no workspace field and no custom implementation?


I now have backtracked and implemented the naive solution (and adding additional debt) to not be blocked too long on this issue
I have not achieved to simplify while not refactoring all the `filter_for_user` calls

bramj (approved):
Pragmatic solution, looking good to me! :+1:

PR #1239, opened 2025-07-03
See code

YolanFery:

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast


frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -0,0 +1,208 @@
...
+                {selectedFile.language}
+                {" • "}
+                {selectedFile.lineCount}
+                {` ${t("lines")}`}
+              </div>
+            </div>
+            <div className="flex-1 overflow-hidden">
+              <CodeMirror
+                value={selectedFile.content!}
+                readOnly={true}
```
bramj, line 167: Really exiting :) Seems to me like making editing possible will be a relative low effort by using `CodeMirror`, great choice :+1:

frontend/src/workspaces/layouts/PipelineLayout/PipelineLayout.tsx
```diff
@@ -99,7 +99,17 @@ const PipelineLayout = (props: PipelineLayoutProps) => {
...
+        pipeline.currentVersion
+          ? [
+              {
+                label: t("Code"),
+                href: `/workspaces/${encodeURIComponent(workspace.slug)}/pipelines/${encodeURIComponent(pipeline.code)}/code`,
+                id: "code",
+              },
+            ]
+          : [],
+      )}
```
bramj, line 112: We should definitely add this to the template pipelines as well. But I get that we'll be reworking that page soon.

bramj (approved):
Great work @YolanFery :+1:  This all looks and feels really well.

Just a few minor thoughts, but nothing that should keep us from merging this.

backend/hexa/pipelines/schema/types.py
```diff
@@ -280,6 +283,18 @@ def resolve_pipeline_run_code(run: PipelineRun, info, **kwargs):
...
 
+def get_language_from_path(path: str) -> str:
+    """Get language from file path extension."""
+    extension = Path(path).suffix
+    supported_languages = {
+        ".py": "python",
+        ".json": "json",
+        ".r": "r",
+        ".md": "markdown",
+    }
```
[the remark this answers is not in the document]
YolanFery: Indeed, thanks for raising those 

> * R files can also be named `.R` (`.R` files are marked `text`, but the syntax highlighter seems to still do it correctly)

Corrected

> * The CLI currently doesn't allow `.json` files to be added. https://github.com/BLSQ/openhexa-sdk-python/blob/main/openhexa/cli/api.py#L562 I guess there's no reason why we don't allow that.

Kept it for the future 

> * As a follow up improvement, we could add `.ipynb` here and see if there's a way to nicely display those. Right now they're shown as plain text:

Improved a bit but not sure we can go further, codemirror does not seems to support this format well

frontend/src/workspaces/layouts/PipelineLayout/PipelineLayout.tsx
YolanFery: Yes we have a ticket for this but wanted to keep this PR with limited scope to simplify the review

YolanFery: https://bluesquare.atlassian.net/browse/HEXA-1309

PR #1241, opened 2025-07-08
fix: add paging to dataset files (HEXA-1329 )

YolanFery:
The paging navigation was not added to dataset files

## Screenshots / screencast

Screenshot 2025-07-08 at 11 02 11


YolanFery:
> LGTM 👍
> 
> I just get an error locally when doing a page refresh when I'm on the files list and my language is FR (not so for EN) image To be honest, not sure if this is an error in my personal config.

Good catch.. thanks, I could reproduce as well

Fixed, do you mind double-checking ? Thanks

bramj (approved):
All good now :+1:

PR #1250, opened 2025-07-08
feat: add current piepline run id

nazarfil:
exposes context of current pipeline run .
Yeah using the midleware for PipelineRunUser , but yeah this is what we get :)


YolanFery (approved):
Nice :)

nazarfil:
I have discovered we push the pipeline run id as ENV_VAR to pipeline_runner so i close this pr

PR #1283, opened 2025-08-18
Pathways 634 workspace specific env variables

nazarfil:
The goal of this PR is to extend workspace with workspace specific configuration parameters to be used across workspace

## Changes

Added `configuration` field to workspace model with CRUD methods
Added tests to test schema and models
Updated settings tab with breadcrumbs tabs as in `dataset`or `pipeline` page to have similar UI/UX
Added dialog to update or edit configuration field and display 
Test frontend features

## How/what to test

Go to settings tab, navigate to General Settings , try to add, remove, delete additional configuration. 

## Screenshots / screencast

https://github.com/user-attachments/assets/2bfd90b0-cd60-487c-bbb3-0cc523f40bad

### DIFF
--- a/backend/hexa/workspaces/migrations/0051_workspace_configuration.py
+++ b/backend/hexa/workspaces/migrations/0051_workspace_configuration.py
@@ -5,7 +5,7 @@ from django.db import migrations, models
 
 class Migration(migrations.Migration):
     dependencies = [
-        ("workspaces", "0049_organizationworkspaceinvitation"),
+        ("workspaces", "0050_merge_20250818_0957"),
     ]
 
     operations = [


frontend/src/workspaces/features/ConfigurationProperty/ConfigurationProperty.test.tsx
```diff
@@ -0,0 +1,457 @@
...
+        </button>
+        {Object.entries(configuration || {}).map(([key, value]) => (
+          <div key={key}>
+            {key}: {typeof value === "string" ? value : JSON.stringify(value)}
+          </div>
+        ))}
+      </div>
+    );
+  };
+});
```
YolanFery, line 27: So basically everything is mocked ? 


I think this test file needs to  be corrected or removed (its a small wrapper around the main component, so not much logic to test)

frontend/src/workspaces/features/ConfigurationList/ConfigurationList.tsx
```diff
@@ -0,0 +1,178 @@
...
+      return value;
+    }
+    return JSON.stringify(value);
+  };
+
+  const getValueType = (value: any): string => {
+    if (typeof value === "string") return "text";
+    if (Array.isArray(value)) return "array";
+    if (value === null) return "null";
+    if (typeof value === "object") return "object";
```
YolanFery, line 81: Maybe we should show `JSON` instead of object

YolanFery (changes requested):
Well done, impressive, especially in a few days...

Reviewed and tested : 
Screenshot 2025-08-20 at 17 23 17


Besides my 2 other comments I want to add that the UI, when the configuration is empty, is misleading, it looks like a dropzone but isn't. I think we should find a different UI or just leave it blank. We could also go way further and make it a real dropzone (we have example in the codebase) and accept json files but it's out of scope 😄

Screenshot 2025-08-20 at 17 49 56

frontend/src/workspaces/features/ConfigurationProperty/ConfigurationProperty.test.tsx
nazarfil: Agree, thanks!

YolanFery:
> > Besides my 2 other comments I want to add that the UI, when the configuration is empty, is misleading, it looks like a dropzone but isn't. I think we should find a different UI or just leave it blank. We could also go way further and make it a real dropzone (we have example in the codebase) and accept json files but it's out of scope 😄
> 
> Screenshot 2025-08-21 at 10 13 29 Here is the new suggestion, wdyt?

Sounds good to me

YolanFery (approved):

PR #1325, opened 2025-08-28
feat(organization): F4 - Internal dataset governance (PATHWAYS-809, HEXA-1349)

yolanfery:
Provide a way to share datasets across the whole organization

## Changes

- To avoid issue with conflicting dataset names, change the URL of datasets to include a `/from/`
- Adapt permissions
- added `sharedWithOrganization` to schema and model
- Filtering for users now include datasets form org sharing

## How/what to test

Datasets are correctly shared/viewable between workspaces

## Screenshots / screencast

Screenshot 2025-08-28 at 16 12 15
Screenshot 2025-08-28 at 16 12 21

### DIFF
--- a/backend/hexa/search/tests/test_schema/test_search.py
+++ b/backend/hexa/search/tests/test_schema/test_search.py
@@ -37,6 +37,11 @@ class SearchResolversTest(GraphQLTestCase):
             "standardpassword",
             is_superuser=False,
         )
+        cls.ADMIN_USER = User.objects.create_user(
+            "admin@bluesquarehub.com",
+            "standardpassword",
+            is_superuser=True,
+        )
         cls.ORGANIZATION = Organization.objects.create(
             name="Test Organization",
         )
@@ -79,21 +84,21 @@ class SearchResolversTest(GraphQLTestCase):
             role=WorkspaceMembershipRole.VIEWER,
         )
 
-        cls.DATASET1 = Dataset.objects.create(
+        cls.DATASET1 = Dataset.objects.create_if_has_perm(
+            principal=cls.ADMIN_USER,
             name="Dataset",
-            slug="dataset",
             description="First dataset",
             workspace=cls.WORKSPACE1,
         )
-        cls.DATASET2 = Dataset.objects.create(
+        cls.DATASET2 = Dataset.objects.create_if_has_perm(
+            principal=cls.ADMIN_USER,
             name="Dataset 2",
-            slug="dataset-2",
             description="Second dataset",
             workspace=cls.WORKSPACE2,
         )
-        cls.DATASET3 = Dataset.objects.create(
+        cls.DATASET3 = Dataset.objects.create_if_has_perm(
+            principal=cls.ADMIN_USER,
             name="Dataset 3",
-            slug="dataset-3",
             description="Third dataset (user not part of)",
             workspace=cls.WORKSPACE3,
         )
@@ -309,22 +314,23 @@ class SearchResolversTest(GraphQLTestCase):
             len(response["data"]["searchFiles"]["items"]),
             4,  # 2 files * 2 workspaces
         )
-        self.assertEqual(
-            response["data"]["searchFiles"]["items"][0],
+
+        items = response["data"]["searchFiles"]["items"]
+        expected_items = [
             {
                 "file": {"name": "file1", "path": "file1"},
                 "score": 1.0,
                 "workspace": {"name": "Workspace 1"},
             },
-        )
-        self.assertEqual(
-            response["data"]["searchFiles"]["items"][1],
             {
                 "file": {"name": "file2", "path": "file2"},
                 "score": 0.5,
                 "workspace": {"name": "Workspace 1"},
             },
-        )
+        ]
+
+        for expected_item in expected_items:
+            self.assertIn(expected_item, items)
 
     @patch("hexa.search.schema.queries.get_database_definition")
     def test_search_database_tables(self, mock_get_database_definition):


backend/hexa/datasets/schema/queries.py
```diff
@@ -67,7 +67,7 @@ def resolve_dataset_link_by_slug(_, info, **kwargs):
     try:
         return DatasetLink.objects.filter_for_user(request.user).get(
             dataset__slug=kwargs["dataset_slug"],
-            workspace__slug=kwargs["workspace_slug"],
+            dataset__workspace__slug=kwargs["workspace_slug"],
```
YolanFery, line 70: Consider the source workspace

backend/hexa/user_management/permissions.py
```diff
@@ -73,8 +73,8 @@ def archive_workspace(principal: User, organization: Organization):
     ).exists()
 
 
-def list_all_workspaces(principal: User, organization: Organization):
-    """Only admin and owner users can list all workspaces"""
+def has_admin_privileges(principal: User, organization: Organization):
+    """Check if user has admin or owner privileges in the organization"""
```
YolanFery, line 77: Let's use that later to bypass restrictions (like access to workspace)

backend/hexa/datasets/models.py
```diff
@@ -49,7 +55,14 @@ def filter_for_workspace_slugs(
...
         return self._filter_for_user_and_query_object(
             user,
-            Q(workspace__members=user, workspace__slug__in=workspace_slugs),
+            Q(
+                workspace__members=user, workspace__slug__in=workspace_slugs
+            )  # 1. It exists a link to a workspace where user is member and the workspace slug is in the list
+            | Q(  # 2. The dataset is shared with organization and user is member of the organization and a link exists to a workspace with slug in the list
+                shared_with_organization=True,
+                workspace__organization__organizationmembership__user=user,
+                links__workspace__slug__in=workspace_slugs,
```
bramj, line 64: I think we can extract some common behavior between `filter_for_user` and  `filter_for_workspace_slugs`, what do you think?

bramj (changes requested):
There's some strange behavior on the following edge case:
1. Create a dataset in workspace A (Org A), share it with the organization A
2. In a different workspace B (Org A), share this same dataset with yet another workspace C

Now there are two possibilities:
1. Workspace C belongs to Org A: step 2 didn't make sense, I think we shouldn't allow that
2. Workspace C belongs to Org B: I suppose that makes sense, but it doesn't seem to work

All this is rather complex. Maybe we should disable the re-sharing of datasets that don't belong to one's organization? What do you think? // @mcrimi 

https://github.com/user-attachments/assets/a16dcc3c-784c-4a28-85e8-b9d9d6309773

One more thing: Creating a new dataset appears broken for me (when I click "Create" in the dialog, the UI doesn't react), but not sure if that's just me.

backend/hexa/datasets/permissions.py
```diff
@@ -51,24 +58,37 @@ def download_dataset_version(principal: User, version: DatasetVersion):
...
-    This also includes members of workspaces that have been shared this dataset.
+    """Users can view datasets if they are workspace members, have access through dataset links,
+    or have access through organization sharing.
     """
     return (
         dataset.links.filter(workspace__in=principal.workspace_set.all()).exists()
         or dataset.workspace.workspacemembership_set.filter(
             user=principal,
             role__in=[WorkspaceMembershipRole.ADMIN, WorkspaceMembershipRole.EDITOR],
         ).exists()
```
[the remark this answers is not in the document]
YolanFery: Not sure at this point

backend/hexa/datasets/models.py
```diff
@@ -40,7 +40,13 @@ def filter_for_user(self, user: AnonymousUser | User):
         else:
             return self._filter_for_user_and_query_object(
                 user,
-                models.Q(links__workspace__members=user),
+                Q(
```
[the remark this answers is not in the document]
YolanFery: Correct, not a priority but will be tackled as part of https://bluesquare.atlassian.net/browse/PATHWAYS-869

backend/hexa/datasets/models.py
YolanFery: DRYed it, thanks

backend/hexa/datasets/models.py
```diff
@@ -437,10 +451,29 @@ def filter_for_user(self, user: AnonymousUser | User):
...
+                        models.When(
+                            models.Q(dataset__shared_with_organization=True)
+                            & ~models.Q(dataset__workspace=models.F("workspace")),
+                            then=models.Value(0),
+                        ),
+                        default=models.Value(1),
+                        output_field=models.IntegerField(),
+                    )
+                )
+                .filter(is_original_link=1)
```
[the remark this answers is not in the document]
YolanFery: I simplified, it should be better now

YolanFery:
> There's some strange behavior on the following edge case:
> 
> 1. Create a dataset in workspace A (Org A), share it with the organization A
> 2. In a different workspace B (Org A), share this same dataset with yet another workspace C
> 
> Now there are two possibilities:
> 
> 1. Workspace C belongs to Org A: step 2 didn't make sense, I think we shouldn't allow that
> 2. Workspace C belongs to Org B: I suppose that makes sense, but it doesn't seem to work
> 

It was just a bug and I fixed it

> One more thing: Creating a new dataset appears broken for me (when I click "Create" in the dialog, the UI doesn't react), but not sure if that's just me.

Fixed it

PR #1328, opened 2025-09-02
fix:  images in emails are not shown correctly, at least in gmail (HEXA-1351)

YolanFery:
Attachments were missing in the `send_email` call

## Changes

- Add the missing `.svg` attachment
- Extract into utils a helper method to include the `.png` logo, to be reused across the apps.

### DIFF
--- a/backend/hexa/user_management/utils.py
+++ b/backend/hexa/user_management/utils.py
@@ -1,10 +1,10 @@
 from datetime import datetime, timezone
 from urllib.parse import urlencode
 
+from django.conf import settings
 from django.utils.translation import gettext_lazy, override
 from django_otp import devices_for_user, user_has_device
 
-from config import settings
 from hexa.analytics.api import track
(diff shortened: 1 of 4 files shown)


nazarfil (approved):
Screenshot 2025-09-03 at 11 21 00
Tested locally with mailhog pipeline notification worked well

backend/hexa/user_management/utils.py
```diff
@@ -98,7 +78,7 @@ def send_organization_add_user_email(
     from django.utils.translation import gettext_lazy, override
 
     from hexa.analytics.api import track
-    from hexa.core.utils import send_mail
+    from hexa.core.utils import get_email_attachments, send_mail
```
bramj, line 81: [Detail] Is there a reason why the imports are at the method level in `send_organization_invite` and `send_organization_add_user_email`?

PR #1338, opened 2025-09-08
fix: permissions and redirect when revoking access to a shared dataset (HEXA-1353)

YolanFery:
Correct the permission and add a redirection when the dataset is not shared anymore with the workspace

### DIFF
--- a/backend/hexa/datasets/models.py
+++ b/backend/hexa/datasets/models.py
@@ -499,7 +499,7 @@ class DatasetLink(Base):
     objects = DatasetLinkQuerySet.as_manager()
 
     def delete_if_has_perm(self, *, principal: User):
-        if not principal.has_perm("datasets.delete_linked_dataset", self):
+        if not principal.has_perm("datasets.delete_dataset_link", self):
             raise PermissionDenied
         self.delete()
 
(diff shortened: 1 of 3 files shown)


frontend/src/core/helpers/apollo.ts
```diff
@@ -63,6 +63,9 @@ const CACHE_CONFIG: InMemoryCacheConfig = {
     DatasetPermissions: {
       merge: true,
     },
+    DatasetLinkPermissions: {
+      merge: true,
+    },
```
YolanFery, line 68: Fix Appollo cache error

backend/hexa/datasets/models.py
```diff
@@ -499,7 +499,7 @@ class DatasetLink(Base):
     objects = DatasetLinkQuerySet.as_manager()
 
     def delete_if_has_perm(self, *, principal: User):
-        if not principal.has_perm("datasets.delete_linked_dataset", self):
+        if not principal.has_perm("datasets.delete_dataset_link", self):
```
YolanFery, line 502: As far as I have seen this has been wrong for a long time for non superuser/admin users...

bramj: Happens to the best of us 🙂

bramj (approved):
LGTM 👍

backend/hexa/datasets/models.py
YolanFery: Yes of course, not blaming, just surprised that we spot it now

bramj: (fyi absolutely no blame intended :+1: )

PR #1348, opened 2025-09-09
fix : Org admins can't manage workspace settings of newly created workspaces (PATHWAYS-889)

yolanfery:
Check for organization roles when doing actions on workspaces


bramj (approved):
Tested and fixes the issue 👍

PR #1360, opened 2025-09-12
fix: dataset view is constrained to the workspaces where the org admin is explicitly assigned (PATHWAYS-890)

yolanfery:
Allow orga owners/admins to see all the datasets in the organization whatever their workspace roles

### DIFF
--- a/backend/hexa/datasets/schema/types.py
+++ b/backend/hexa/datasets/schema/types.py
@@ -61,10 +61,9 @@ def resolve_workspace_datasets(obj: Workspace, info, pinned=None, query=None, **
     org_shared = DatasetLinkQuerySet.optimize_query(
         DatasetLink.objects.filter(organization_shared_q)
     )
-    org_shared_ids = org_shared.values("dataset_id")
     direct_links = DatasetLinkQuerySet.optimize_query(
         DatasetLink.objects.filter(workspace=obj).exclude(
-            dataset_id__in=org_shared_ids.union(org_shared_ids)
+            dataset_id__in=org_shared.values("dataset_id")
         )
     )
 


backend/hexa/datasets/schema/types.py
```diff
@@ -58,24 +62,41 @@ def resolve_workspace_datasets(obj: Workspace, info, pinned=None, query=None, **
...
+            dataset_id__in=org_shared_ids.union(org_admin_or_owner_ids)
         )
     )
 
     if query is not None:
-        org_shared = org_shared.filter(name__icontains=query)
-        direct_links = direct_links.filter(name__icontains=query)
+        org_shared = org_shared.filter(dataset__name__icontains=query)
+        org_admin_or_owner = org_admin_or_owner.filter(dataset__name__icontains=query)
+        direct_links = direct_links.filter(dataset__name__icontains=query)
```
nazarfil, line 92: I remember seeing `name__icontains` in previous PR and spending some time searching how did it work  with dataset link name,  remember testing search in UI and it was working so i accepted it. Glad to see the change here

backend/hexa/datasets/schema/types.py
```diff
@@ -58,24 +62,41 @@ def resolve_workspace_datasets(obj: Workspace, info, pinned=None, query=None, **
...
+            dataset__workspace=F("workspace"),
+        )
+
     org_shared = DatasetLinkQuerySet.optimize_query(
         DatasetLink.objects.filter(organization_shared_q)
     )
+    org_shared_ids = org_shared.values("dataset_id")
+    org_admin_or_owner = DatasetLinkQuerySet.optimize_query(
+        DatasetLink.objects.filter(org_admin_or_owner_q).exclude(
+            dataset_id__in=org_shared_ids
```
nazarfil, line 78: Tested this in previous PR and was unable to have duplicates in the query :D

nazarfil (approved):

YolanFery:
> The queries have become quite complex and I'm having a hard time understanding the rules of who sees which dataset. Maybe it should be a docstring somewhere that explains that?
> 
> More generally speaking, I'm worried about performance on the dataset endpoints. Right now there are big queries with `OR`s and `UNION`s that could kill performance.
> 

Actually, now that I have fresh eyes on this, we don't need those changes for `workspaces.datasets`, we don't care about other datasets in other workspaces (except the one directly shared through link or through the organization sharing feature). So I reverted it.

For the rest, I agree the dataset queries/logic became quite complex but it's inherent complexity : we need to look for datasets in the workspace, shared through links and shared through organization 

Nevertheless, I am right now focusing on making the feature right but I think, like you, that we could benefit from some refactoring/cleanup/optimization. Let me have a look at that once the feature is right.

> Maybe there's no way around this, but maybe things like early returns when the user is an admin/owner of the organization can help. The current query plans are very complex and I don't think they'll scale well. But we can see how it behaves on demo.

PR #1363, opened 2025-09-15
fix: dataset link by slug should have a fallback to any of the dataset with the slug shared with the workspace (HEXA-1367)

yolanfery:
Implement successive fallbacks for `datasetLinkBySlug`
Add unit tests to cover edge cases

### DIFF
--- a/backend/hexa/datasets/admin.py
+++ b/backend/hexa/datasets/admin.py
@@ -23,12 +23,17 @@ class DatasetVersionAdmin(admin.ModelAdmin):
 
 
 @admin.register(DatasetVersionFile)
-class DatasetVersionObjectAdmin(admin.ModelAdmin):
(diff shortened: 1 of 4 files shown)


backend/hexa/datasets/schema/queries.py
```diff
@@ -64,17 +65,44 @@ def resolve_dataset_link(_, info, **kwargs):
...
+    if not workspace or not workspace.organization:
         return None
+    result = user_links.filter(
+        dataset__slug=dataset_slug,
+        dataset__shared_with_organization=True,
+        dataset__workspace__organization=workspace.organization,
+    ).first()
+    if result:
+        return result
+
```
nazarfil, line 104: Good  priority based approach

nazarfil (approved):

PR #1372, opened 2025-09-18
fix: prettier config and pretty all files in the repo

YolanFery:


frontend/eslint.config.mjs
```diff
@@ -0,0 +1,39 @@
+import { defineConfig } from "eslint/config";
```
YolanFery, line 1: New file generated from the migration script

nazarfil (approved):
i guess it is prettier now 🤷

PR #1389, opened 2025-10-01
fix: update dataset version changelog when switching versions

nazarfil:
fixes issue when changelog was not update on version chagne 

https://github.com/user-attachments/assets/ac17a3c9-4a8d-46ac-96b8-53d97fbc94ca


frontend/src/pages/workspaces/[workspaceSlug]/datasets/[datasetSlug]/from/[sourceWorkspaceSlug]/index.tsx
```diff
@@ -144,7 +154,7 @@ const WorkspaceDatasetPage: NextPageWithLayout = (
           </FormSection>
         </ItemProvider>
         {version && (
-          <ItemProvider item={version}>
+          <ItemProvider key={version.id} item={version}>
```
Comment (YolanFery) [inline frontend/src/pages/workspaces/[workspaceSlug]/datasets/[datasetSlug]/from/[sourceWorkspaceSlug]/index.tsx:157]: I think that this is the only required and minimal change : 

`<ItemProvider key={props.versionId} item={version}>`

The rest  of the PR is not needed I think

YolanFery (approved):

PR #1395, opened 2025-10-03
fix: improve logging for the pipeline heartbeat mechanism + share config to run pipelines on kubernetes locally (HEXA-1386)

YolanFery:
Improve logging on pipeline runner
And share some setup to help running pipelines on kubernetes, its not perfect but it is a starting point for helping to debug


backend/hexa/pipelines/management/commands/pipelines_runner.py
```diff
@@ -340,7 +409,9 @@ def run_pipeline(run: PipelineRun):
         elif spawner == "kubernetes":
             success, container_logs = run_pipeline_kube(run, image, env_vars)
         else:
-            logger.error("Scheduler spawner %s not found", settings.SCHEDULER_SPAWNER)
+            logger.error(
+                "Scheduler spawner %s not found", settings.PIPELINE_SCHEDULER_SPAWNER
```
YolanFery, line 413: SCHEDULER_SPAWNER does not exist, this was a small bug

PR #1398, opened 2025-10-06
chore: file browser improvements (HEXA-1360)

bramj:
2nd attempt after https://github.com/BLSQ/openhexa-app/pull/1391

This PR contains several improvements:

- Improve error message when creating a directory that already exists
- Performance: change the file icon and color helper to flat map for O(1) access
- Refactoring: Align the `File` GraphQL type with `BucketObject`. These were almost identical except for `updated` vs `updatedAt`. Now `updated` is deprecated, and both are returned for the `File` type.
- Hide the lonely "Home" icon when we're in the root folder or in search mode

Basically the same PR, but this time without breaking changes to the API.


frontend/src/workspaces/features/FileBrowserDialog/FileBrowserDialog.tsx
```diff
@@ -171,49 +170,31 @@ const FileBrowserDialog = (props: FileBrowserDialogProps) => {
...
     if (isSearchMode && searchResults) {
-      // Convert search results to bucket object format
-      return searchResults.items.map((result) => ({
-        ...result.file,
-        // Map search result fields to bucket object fields
-        updatedAt: result.file.updated,
-        type:
-          result.file.type === FileType.Directory
-            ? BucketObjectType.Directory
-            : BucketObjectType.File,
```
YolanFery, line 185: Is the type mapping covered now ?

bramj: Damn, good catch. They're not exactly the same, I had to bring back this snippet 🫤

YolanFery: Thanks

PR #1402, opened 2025-10-07
fix: workspace admins should be able to create workspaces (HEXA-1387)

yolanfery:
To create a workspace you need to have the feature flag OR be a org admin/owner OR a workspace admin of at least one workspace


bramj (approved):
Tested and working correctly 👍 
I don't really like how complex the logic has gotten, but I understand the need for it right now, and I think this should solve the users' issues for now.

nazarfil (approved):

PR #1718, opened 2026-04-01
feat: MCP extensive tool set (HEXA-1565)

yolanfery:
We previously created a basic MCP with initial tools. 
This PR is now about adding tools to make it really useful.

Please note that this PR https://github.com/BLSQ/openhexa-app/pull/1693 is adding the `createPipeline` tool already

## Changes

- Split tools in multiple files for easier review and maintenance
- Add tools based on https://github.com/BLSQ/mcp_servers/blob/main/openhexa_mcp_server.py (thanks @jhubinon and @mcrimi)
- Add basic unit tests to have an initial test setup and cover already some edge cases


backend/hexa/mcp/tools/webapps.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@tool
+def create_static_webapp(
+    user,
+    workspace_slug: str,
+    name: str,
+    files_json: str,
+    description: str = "",
+    is_public: bool = False,
```
mcrimi: @yolanfery — I wonder if we should enable publication through the MCP server or let it always a human intentional action. I'm leaning towards the latter, as the potential risks might outweigh the benefits. Wdyt?

backend/hexa/mcp/tools/help.py
```diff
@@ -0,0 +1,27 @@
+from hexa.mcp.protocol import tool
+
+
+@tool
+def get_help(user, reason: str = "") -> dict:
+    """Call this tool when you are stuck, unsure what to do next, or need guidance on OpenHEXA. Provide a reason describing why you need help (e.g. 'unsure which tool to use', 'pipeline failed', 'cannot find dataset')."""
```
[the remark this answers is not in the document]
mcrimi: This is so clever. Did you come up with this pattern?

backend/hexa/mcp/tools/datasets.py
```diff
@@ -0,0 +1,105 @@
...
+            }
+        },
+    )
+    if "errors" in data:
+        return data
+    return data["createDataset"]
+
+
+@tool
+def create_dataset_version(
```
mcrimi, line 106: @yolanfery — I think the file upload should be part of the same atomic operation as the dataset version creation. If not the system will reach an inconsistent state (a dataset version without files). See screencast:

Image

backend/hexa/mcp/tools/_graphql.py
```diff
@@ -0,0 +1,29 @@
...
+    result = graphql_sync(
+        schema,
+        _QUERIES,
+        context_value={"request": _make_request(user)},
+        variable_values=variables or {},
+        operation_name=operation_name,
+    )
+    if result.errors:
+        return {"errors": [str(e) for e in result.errors]}
+    return result.data
```
mrivar, line 30: I like having this in a separate file!

backend/hexa/mcp/graphql/queries.graphql
```diff
@@ -49,6 +49,243 @@ query ListPipelines($workspaceSlug: String, $page: Int, $perPage: Int) {
   }
 }
 
+query GetPipeline($workspaceSlug: String!, $code: String!, $runsPage: Int, $runsPerPage: Int) {
```
mrivar: I wonder if it would be nice to have these in multiple files for the same reason as with the tools

backend/hexa/mcp/tools/help.py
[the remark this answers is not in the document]
mrivar: It's such a good idea, love it! I wonder if the LLMs would be smart enough to "tag" (like reason but just in one word) each time the help tool is triggered, so that we could visualise what is triggering it the most

mrivar (approved):
Looks great! I love how it's been refactored and organised.
It's only missing the create_pipeline from #1693, which will be solved after the conflict. I only had some comments but they're suggestions more than anything, you can either take them or leave them.

backend/hexa/mcp/tools/datasets.py
mrivar: So as with the AI pipeline creation, the first version would get created at the same time as creating the dataset?

backend/hexa/mcp/tools/help.py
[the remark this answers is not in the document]
yolanfery: > This is so clever. Did you come up with this pattern?

I wish... I read it somewhere and thought we could do something similar



> It's such a good idea, love it! I wonder if the LLMs would be smart enough to "tag" (like reason but just in one word) each time the help tool is triggered, so that we could visualise what is triggering it the most

What do you mean by tag ? I am not sure I get this

backend/hexa/mcp/tools/webapps.py
yolanfery: Indeed, great point, thanks, I will change this

backend/hexa/mcp/tools/datasets.py
yolanfery: Makes total sense to keep it atomic, I overlooked this, I will change this, thanks

backend/hexa/mcp/graphql/queries.graphql
yolanfery: Makes total sense indeed, I will try to split them, thanks

backend/hexa/mcp/tools/help.py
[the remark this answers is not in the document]
mrivar: > What do you mean by tag ? I am not sure I get this

I meant something like the parameter `reason` but with just 1 or 2 words. Like some sort of classification of reasons.
If it stays consistent (chooses the same tags for same/related problems), it would be really easy to find out what problems are happening the most (in comparison to going through each help tool call and reading each `reason`)

backend/hexa/mcp/tools/pipelines.py
```diff
@@ -0,0 +1,172 @@
...
+
+@tool
+def create_pipeline(
+    user,
+    workspace_slug: str,
+    name: str,
+    description: str = "",
+    functional_type: PipelineFunctionalType | None = None,
+    source_code: str | None = None,
+) -> dict:
```
mrivar, line 123: just realized I'm missing a create_pipeline_version to update the pipeline version

mrivar (changes requested):

backend/hexa/mcp/tools/__init__.py
```diff
@@ -0,0 +1,10 @@
+from hexa.mcp.tools import (  # noqa: F401
+    connections,
+    datasets,
+    files,
+    help,
+    pipelines,
+    templates,
+    webapps,
+    workspaces,
+)
```
bramj, line 10: Good structure 👍

backend/hexa/mcp/tools/connections.py
```diff
@@ -0,0 +1,15 @@
...
+@tool
+def list_connections(user, workspace_slug: str) -> dict:
+    """List connections (external data sources) configured in a workspace. Returns connection name, slug, type (S3, GCS, POSTGRESQL, DHIS2, IASO, CUSTOM), and their fields with values. Secret fields have their values hidden. Connections are used as parameters when running pipelines — use the connection slug as the parameter value."""
+    data = execute_graphql(user, "ListConnections", {"slug": workspace_slug})
+    if "errors" in data:
+        return data
+    workspace = data.get("workspace")
+    if workspace is None:
+        return {"error": "Workspace not found"}
+    return {"connections": workspace["connections"]}
```
bramj, line 15: > Secret fields have their values hidden

So the agent cannot yet _use_ a connection, right? We'll want this of course, but not yet.

backend/hexa/mcp/tools/datasets.py
```diff
@@ -0,0 +1,135 @@
...
+
+@tool
+def preview_dataset_file(user, file_id: str) -> dict:
+    """Preview the content of a dataset file by its ID (from get_dataset's file list). Returns a sample of the data for tabular files (CSV, Parquet, etc.), file properties, and metadata. The sample status can be PROCESSING (still generating), FINISHED (sample ready), or FAILED."""
+    data = execute_graphql(user, "PreviewDatasetFile", {"id": file_id})
+    if "errors" in data:
+        return data
+    file_data = data.get("datasetVersionFile")
+    if file_data is None:
+        return {"error": "Dataset file not found"}
```
bramj, line 66: This is more "Dataset file sample not found", no? Are you sure we should consider this an error?

backend/hexa/mcp/tools/datasets.py
```diff
@@ -0,0 +1,135 @@
...
+
+@tool
+def create_dataset(
+    user,
+    workspace_slug: str,
+    name: str,
+    description: str = "",
+    files_json: str = "",
+) -> dict:
+    r"""Create a new dataset in a workspace. Optionally create an initial version (named 'v1') with files in a single atomic operation. To include files, provide files_json as a JSON array of {uri, contentType, content} objects, e.g. '[{"uri": "data.csv", "contentType": "text/csv", "content": "a,b\n1,2"}]'."""
```
bramj: > Optionally create an initial version (named 'v1')

Can we think of a situation where we would want this to be optional? I feel like it can only create orphaned datasets.

backend/hexa/mcp/tools/files.py
```diff
@@ -0,0 +1,59 @@
...
+def read_file(user, workspace_slug: str, file_path: str) -> dict:
+    """Read the content of a text file from a workspace bucket. Only works for UTF-8 text files up to 1 MB (includes .py, .csv, .json, .ipynb, .sql, .txt, etc.). Use list_files first to check file size and path. For Jupyter notebooks (.ipynb), the content is JSON that you can parse to read/modify cells."""
+    data = execute_graphql(
+        user,
+        "ReadFileContent",
+        {"workspaceSlug": workspace_slug, "filePath": file_path},
+    )
+    if "errors" in data:
+        return data
+    return data["readFileContent"]
```
bramj, line 50: 👍 we should start thinking how we can allow something like `read_file(start_line, end_line)` to allow very focused reading.

backend/hexa/mcp/tools/help.py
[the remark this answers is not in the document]
bramj: This is great! I think we can later add a similar pattern like `read_documentation` when we have migrated the docs.
And then a compressed index can be part of the initial instructions (or even fetched on the fly if it's too big).

backend/hexa/mcp/tools/templates.py
```diff
@@ -0,0 +1,53 @@
...
+    if "errors" in data:
+        return data
+    template = data.get("templateByCode")
+    if template is None:
+        return {"error": "Template not found"}
+    return template
+
+
+@tool
+def create_pipeline_from_template(
```
bramj, line 37: 👍 very useful

bramj (commented):
A very fast review: this is looking great! 👏

backend/hexa/mcp/tools/help.py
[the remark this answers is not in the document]
yolanfery: > > What do you mean by tag ? I am not sure I get this
> 
> I meant something like the parameter `reason` but with just 1 or 2 words. Like some sort of classification of reasons. If it stays consistent (chooses the same tags for same/related problems), it would be really easy to find out what problems are happening the most (in comparison to going through each help tool call and reading each `reason`)

Thanks for the idea ! For now the usage volume is so low that the more context we have the more we can improve but then yes, if we get too much we can think about a tagging system instead

backend/hexa/mcp/tools/pipelines.py
yolanfery: Will add indeed, thanks !

backend/hexa/mcp/tools/connections.py
yolanfery: The docstring is actually incorrect

I changed it and provided a unit test to verify, thanks for noticing

backend/hexa/mcp/tools/datasets.py
yolanfery: No, `datasetVersionFile ` is `None` when the file is not found , this should be correct

backend/hexa/mcp/tools/datasets.py
yolanfery: Right, better to enforce files, this will limit orphaned datasets, thanks !

backend/hexa/mcp/tools/files.py
yolanfery: Added this ability to both graphql and the tool

mrivar (approved):
Looking great!

PR #1719, opened 2026-04-01
feat: add secret parameter type for pipeline parameters

DimitriKwihangana:
Add secret parameter type to mask sensitive pipeline parameters in the UI

## Changes

- Added secret to the ParameterType enum in the GraphQL schema so the backend accepts and returns it as a valid type.
- Masked secret parameter values in the run form (password input) and run detail page (••••••).

## How/what to test

Create a pipeline that has Secret type on the CLI and on the app we will have •••••• as that parameter

## Screenshots / screencast

image
image

### DIFF
--- a/backend/hexa/pipelines/graphql/schema.graphql
+++ b/backend/hexa/pipelines/graphql/schema.graphql
@@ -67,6 +67,7 @@ enum ParameterType {
   gcs
   custom
   file
+  secret
 }
 
 """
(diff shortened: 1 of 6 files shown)


yolanfery (approved):
Great ! Shouldn't we also hide it in the default value column of the parameters section ?

mrivar (approved):
It looks great! I like that it is consistent with the look of the api key in the AI Settings as well 😄

PR #1722, opened 2026-04-03
feat: Serve static webapps from their own subdomain (HEXA-1569)

bramj:
Related Jira: https://bluesquare.atlassian.net/browse/HEXA-1569

Serving static web apps from the main [openhexa.org](http://openhexa.org/) domain has **major security implications**: cookies can bleed over, and an attacker could potentially steal a user’s session cookie.

To make sure web apps are completely isolated, we need to serve them from a separate domain. Moreover, to avoid leaking session data from one webapp to another, **every webapp needs to have its own subdomain**.

## Changes

- Added field `subdomain` on `Webapp` model, along with extensive validation to avoid future issues
- Django middleware to catch incoming requests to the `*.webapps.openhexa.org` subdomain and reroute to the correct web app
- Authentication token mechanism to handle _private_ webapps
- Small refactoring in creation of webapps

## How/what to test

The different scenarios:
- User is logged in, webapp is public
- User is logged in, webapp is private -> user has acces/no access
- User is not logged in, webapp is public
- User is not logged in, webapp is private

Important: if the `WEBAPPS_SUBDOMAIN_BASE_URL` env variable is not set, then the regular "path-serving" should be preserved.

## Screenshots

image

image


docker-compose.yaml
```diff
@@ -16,6 +16,7 @@ x-app: &common
       - CSRF_COOKIE_SECURE=False
       - SECURE_SSL_REDIRECT=False
       - PIPELINE_SCHEDULER_SPAWNER=docker
+      - WEBAPPS_SUBDOMAIN_BASE_URL=webapps.lvh.me:8000
```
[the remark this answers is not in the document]
yolanfery: I have set to use `webapps.localhost:8000` instead by default while still allowing to set `WEBAPPS_SUBDOMAIN_BASE_URL=webapps.lvh.me:8000` in `.env` , it's cleaner to not rely on such external service

yolanfery (approved):

yolanfery:
@bramj amazing start ! 

I did quite a lot of changes (removed the path serving, added tests, serving iframes and superset on the subdomains,  handling CORS across subdomains, UI for editing the subdomains...)

Hope you can find the energy/focus time to review those, thanks

bramj:
@yolanfery thanks a bunch! I'll try to look at this later today.

A quick reply on (without looking at the code):

> removed the path serving

Did you see the notice in the PR description about local hosting? I'm not sure we can be certain that a local hosting will give us the infrastructure changes that are needed for this. Also as a migration path, it would make sense to keep supporting the path-serving. For local installations, the security implications of this are a lot less relevant.
Do you think we should make the new ENV var a hard dependency for local installations?

All other changes sounds great! 👍

yolanfery:
> @yolanfery thanks a bunch! I'll try to look at this later today.
> 
> A quick reply on (without looking at the code):

Thanks @bramj  🙏

> 
> > removed the path serving
> 
> Did you see the notice in the PR description about local hosting?

Sorry, didn't see it. I removed it to be sure we get rid of this security issue in our SaaS

> I'm not sure we can be certain that a local hosting will give us the infrastructure changes that are needed for this. 

Hum I agree but we can't just enable this feature if we know for sure that it's a security breach

> Also as a migration path, it would make sense to keep supporting the path-serving. 

This feature is not in prod yet, so we shouldn't have to support a migration

> For local installations, the security implications of this are a lot less relevant.

I kinda get what you mean here but I would be against offering the ability to enable/use a feature where we know it's an openbar XSS injection 
Also, keeping this path serving is a branch that will be difficult to maintain and support (since we will not use it), I would rather have local installations either disable this feature or use the same setup as in our SaaS

> Do you think we should make the new ENV var a hard dependency for local installations? 

Only thing I can see is to disable the feature if the env is not provided (if the installer is not able to provide a domain). To not block other features if they are not able to just provide a subdomain.

We might want to discuss all this together/as a team 

> All other changes sounds great! 👍

bramj:
@yolanfery thanks!

> I kinda get what you mean here but I would be against offering the ability to enable/use a feature where we know it's an openbar XSS injection

I don't really agree with this: the context of a local installation is very different since the users on the platform are trusted within the organization, there's no random people signing up.
So by extension we can reasonably consider code in the web apps to be trusted. In this context path-based serving is not really problematic.

> Also, keeping this path serving is a branch that will be difficult to maintain and support (since we will not use it), I would rather have local installations either disable this feature or use the same setup as in our SaaS
> Only thing I can see is to disable the feature if the env is not provided (if the installer is not able to provide a domain). To not block other features if they are not able to just provide a subdomain.

I agree with you, but I just want to make sure we won't have to disable the web apps altogether for certain setups, since that would be a loss.
I'm going to discuss the local hosting migration path with Mbayang because I also wonder how complex it will be to add Forgejo to those setups.

> We might want to discuss all this together/as a team

Let's do that 👍 I should be able to attend the daily today.

backend/hexa/webapps/migrations/0014_add_subdomain_field.py
```diff
@@ -3,14 +3,17 @@
...
-    from hexa.webapps.models import Webapp
+    Webapp = apps.get_model("webapps", "Webapp")
 
     for webapp in Webapp.objects.filter(subdomain__isnull=True):
-        webapp.subdomain = create_webapp_subdomain(webapp.slug, webapp.workspace)
+        slug = webapp.slug
+        if Webapp.objects.filter(subdomain=slug).exists():
+            webapp.subdomain = f"{webapp.workspace.slug}-{slug}"
+        else:
+            webapp.subdomain = slug
```
bramj, line 16: Oups, good catch 👍

backend/hexa/webapps/migrations/0015_make_subdomain_required.py
```diff
@@ -0,0 +1,39 @@
...
+        migrations.AlterField(
+            model_name="webapp",
+            name="subdomain",
+            field=models.CharField(
+                max_length=63,
+                unique=True,
+                validators=[hexa.webapps.validators.validate_subdomain],
+            ),
+        ),
+    ]
```
bramj, line 39: Yes! That's great 👍

backend/hexa/webapps/schema/types.py
```diff
@@ -43,11 +46,23 @@ def resolve_workspace(webapp: Webapp, info, **kwargs):
...
+def resolve_preview_url(webapp: Webapp, info, **kwargs):
+    request = info.context["request"]
+    if not request.user.is_authenticated:
+        return webapp.serve_url
+
+    signer = TimestampSigner()
+    token = signer.sign_object(
+        {"user_id": str(request.user.id), "subdomain": webapp.subdomain}
+    )
+    return f"{webapp.serve_url}?{urlencode({'auth_token': token})}"
```
bramj, line 63: Very smooth 👍  this is coming together really nicely

backend/hexa/webapps/models.py
```diff
@@ -34,10 +34,18 @@ def create_webapp_slug(name: str, workspace: Workspace):
...
+    for candidate in candidates:
+        if not Webapp.all_objects.filter(subdomain=candidate).exists():
+            return candidate
+    for _ in range(max_tries):
+        candidate = f"{workspace.slug}-{slug}-{secrets.token_hex(3)}"
+        if not Webapp.all_objects.filter(subdomain=candidate).exists():
+            return candidate
+    raise ValueError(
+        f"Could not generate a unique subdomain after {max_tries} attempts"
+    )
```
bramj, line 48: 👍 good catch. I misread the unique ID on the `WebApp`:
```python
            models.UniqueConstraint(
                "workspace_id",
                "slug",
                name="unique_webapp_slug_per_workspace",
                condition=Q(deleted_at__isnull=True),
            ),
```
Honestly, I think for slugs we should just keep it simple in the future and make it unique without any combined fields or conditions. It's just so much easier to handle for adjacent cases like this.

backend/hexa/webapps/urls.py
```diff
@@ -5,7 +5,5 @@
 app_name = "webapps"
 
 urlpatterns = [
-    path("<uuid:webapp_id>/", views.serve_webapp, name="serve_webapp"),
     path("<uuid:webapp_id>/auth-token/", views.auth_token, name="auth_token"),
-    path("<uuid:webapp_id>/<path:path>", views.serve_webapp, name="serve_webapp_file"),
```
bramj, line 9: 👍

frontend/src/pages/workspaces/[workspaceSlug]/webapps/index.tsx
```diff
@@ -148,9 +148,7 @@ const WebappsPage = (props: Props) => {
               </BaseColumn>
               <LinkColumn
                 id="play"
-                url={(item) =>
-                  item.type === WebappType.Superset ? item.url : item.serveUrl
-                }
+                url={(item) => item.serveUrl ?? item.url}
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/webapps/index.tsx:151]: Love it, so much better

frontend/src/webapps/features/WebappForm/WebappForm.tsx
```diff
@@ -250,7 +250,7 @@ const WebappForm = ({ workspace, webapp }: WebappFormProps) => {
         <TextProperty
           id="url"
           accessor="url"
-          label={t("URL")}
+          label={t("Source URL")}
```
bramj, line 253: 👍 not sure if this will be enough, but the terminology gets confusing here

frontend/src/webapps/features/WebappForm/WebappForm.tsx
```diff
@@ -275,15 +275,15 @@ const WebappForm = ({ workspace, webapp }: WebappFormProps) => {
...
         )}
         {webapp && (
           <SubdomainProperty
             id="subdomain"
             accessor="subdomain"
-            label={t("Subdomain")}
+            label={t("Published URL")}
             help={t(
-              "Subdomain used to generate the URL for accessing this web app",
+              "The URL where this web app can be accessed",
```
bramj, line 286: 👍

backend/hexa/webapps/middlewares.py
```diff
@@ -0,0 +1,201 @@
...
+                session = _create_webapp_session(webapp, result)
+            elif not _check_webapp_session(request, webapp):
+                return HttpResponseRedirect(_build_auth_token_url(request, webapp))
+
+        if webapp.type == Webapp.WebappType.STATIC:
+            response = _serve_static_webapp(webapp, request)
+        elif webapp.type == Webapp.WebappType.SUPERSET:
+            response = _serve_superset_webapp(request, webapp)
+        else:
+            response = _serve_iframe_webapp(webapp)
```
bramj, line 197: Nice, this is very easy to reason about 👏

backend/hexa/templates/webapps/embed.html
```diff
@@ -0,0 +1,30 @@
...
+                width: 100%;
+                height: 100%;
+                border: none;
+            }
+        </style>
+    </head>
+    <body>
+        <iframe src="{{ url }}" allow="fullscreen"></iframe>
+    </body>
+</html>
```
bramj, line 30: I like this 👍

bramj (commented):
@yolanfery I can only comment since it's me who opened the PR, but this is awesome 👍 

I think this is really solid now. Just 2 questions:
- What if the `WEBAPPS_SUBDOMAIN_BASE_URL` ENV var isn't set? In this case, we should block the whole web apps feature, I think that's missing?
- Just a question: I notice we keep the `auth_token` in the URL now when arriving on a private web apps:
Image
Is this intentional? I feel it looks nicer to remove this query param when arriving on the page, what do you think?

Great job 👍 👏

backend/hexa/webapps/models.py
yolanfery: > Honestly, I think for slugs we should just keep it simple in the future and make it unique without any combined fields or conditions. It's just so much easier to handle for adjacent cases like this.

Totally agree with this !!

yolanfery:
> @yolanfery I can only comment since it's me who opened the PR, but this is awesome 👍
> 
> I think this is really solid now. Just 2 questions:
> 
> * What if the `WEBAPPS_SUBDOMAIN_BASE_URL` ENV var isn't set? In this case, we should block the whole web apps feature, I think that's missing?

Yes, was waiting for your review before doing this change as discussed. Done and renamed to `WEBAPPS_DOMAIN` which is more accurate and will accommodate the new domain once we have it bought

> * Just a question: I notice we keep the `auth_token` in the URL now when arriving on a private web apps:
> 
> Image Is this intentional? I feel it looks nicer to remove this query param when arriving on the page, what do you think?

Not intentional, good catch. I liked what you did for that and we need to keep it. I fixed it, thanks

> Great job 👍 👏

PR #1723, opened 2026-04-07
feat: AI Assistant - edit pipelines (HEXA-1515)

mrivar:
Added an AI assistant to edit pipelines

## Changes

- Added an edit pipeline agent that receives the current pipeline code as input / instructions
- Linked Pipeline FK to Conversation (optional, only for editing pipeline conversations)
- Pipeline Code page:
  - Added an AI Assistant button
  - Added an AI Assistant Chat Pane
  - Show proposed changes in File Editor

## How/what to test

- Enter a pipeline with code
- Open the AI assistant
- Ask him to do a modification

## Screenshots / screencast

image


backend/hexa/assistant/models.py
```diff
@@ -42,13 +43,21 @@ def create_if_has_perm(
...
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     workspace = models.ForeignKey(Workspace, on_delete=models.CASCADE)
 
+    pipeline = models.ForeignKey(
+        "pipelines.Pipeline",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="assistant_conversations",
+    )
```
bramj, line 64: I believe this should be a generic/polymorphic relation: https://docs.djangoproject.com/en/6.0/ref/contrib/contenttypes/#generic-relations

That way we can link it to any object type. In the near future, we'll want to link conversations to other resources such as web apps or even pipeline runs or datasets etc.

backend/hexa/pipelines/migrations/0006_auto_20210914_1402.py
```diff
@@ -14,6 +14,7 @@
 
 class Migration(migrations.Migration):
     dependencies = [
+        ("catalog", "0009_enable_trgm"),
```
bramj, line 17: Not sure what this one is about?

bramj (commented):
I know this is still in draft, but damn that's cool! 👏 😄 🎉 

I really dig the "diff that you need to approve" UX, that's exactly what we're going for.
I had a (very) quick look at the code, and I have only one relevant comment, see below.

backend/hexa/pipelines/migrations/0006_auto_20210914_1402.py
mrivar: Yeah I'm extremely confused about this one as well. The tests were failing (https://github.com/BLSQ/openhexa-app/actions/runs/24080400712/job/70239210855) so I tried to find out what was going on and couldn't find anything. Then I sent it to Claude and it found this, which I was very sceptic about but it did fix it..
This is the explanation it gave me after questioning him:

> The assistant app is new (created 2026-03-09). When it was added to INSTALLED_APPS, the test database had to be destroyed and recreated from scratch to apply its initial migrations.
>   That recreation is when the long-standing ordering bug in pipelines/0006 first surfaced.
> 
>   Before: the test DB was old and already had pg_trgm installed (catalog migrations ran first historically, pg_trgm stayed across runs).
>   After test DB rebuild: Django's migration executor happened to schedule pipelines/0006 (which needs pg_trgm) before catalog/0009 (which installs it), because there's no declared
>   dependency between them.
> 
>   The bug was always latent in pipelines/0006 — it's missing the dependency ("catalog", "0009_enable_trgm"). It just never mattered while the test DB persisted. Adding the assistant
>   app forced a DB rebuild and exposed it.
> 
>   The fix I applied is correct. You can revert my change to pipelines/0006 if you disagree, but then you'd need another way to guarantee ordering — like moving the TrigramExtension()
>   call to a core migration that everything transitively depends on.

bramj: Also not sure what that all means 😅 But I have a recollection of having done something similar to fix an old migration on Trypelim (also Django). It's a harmless fix anyways 👍

bramj:
@mrivar A few small glitches I notice when testing for the demo (no problem at all, just letting you know):

- Vertical scroll on the conversation when it gets longer, I think the box needs a `vheight` or something
- The "reject proposal" banner stays after I accept the changes
- I think the agent is a bit too verbose in it's reply when I ask for a change. What do you think about instructing it to be concise, especially when the result is a tool use? Subjective, I know 😉

backend/hexa/pipelines/migrations/0006_auto_20210914_1402.py
mrivar: Yep, very confusing to me as well 😅

backend/hexa/assistant/models.py
mrivar: Great idea! That makes a lot of sense, especially for the next developments. I'll be on it now.

backend/hexa/assistant/agents/create_pipeline_agent.py
```diff
@@ -3,7 +3,7 @@
 from hexa.mcp.tools.pipelines import create_pipeline
 
 
-class PipelineAgent(BaseAgent):
+class CreatePipelineAgent(BaseAgent):
     # TODO: Add reading pipeline / pipeline templates tools
```
yolanfery, line 7: [Minor] We might want to handle this now ?

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Include ALL files in the proposal, not only the ones you modified.
+    Each file must have a 'name' (e.g. 'pipeline.py') and 'content' (the full file text).
+    """
+    return {"files": [f.model_dump() for f in files]}
+
+
+class EditPipelineAgent(BaseAgent):
+    instruction_set = InstructionSet.EDIT_PIPELINE
+    tools = [propose_pipeline_version]
```
yolanfery, line 55: Cool stuff

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -0,0 +1,65 @@
...
+
+class ProposedFile(BaseModel):
+    name: str
+    content: str
+
+
+def propose_pipeline_version(files: list[ProposedFile]) -> dict:
+    """Propose a new version of the pipeline with updated files.
+
+    Include ALL files in the proposal, not only the ones you modified.
```
yolanfery: Isn't that a bad idea for token usage/performance ?

We can probably ask for the files to modify/delete and keep the rest

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    return {"files": [f.model_dump() for f in files]}
+
+
+class EditPipelineAgent(BaseAgent):
+    instruction_set = InstructionSet.EDIT_PIPELINE
+    tools = [propose_pipeline_version]
+
+    def _extra_instructions(self) -> str:
+        from hexa.pipelines.models import Pipeline
```
yolanfery: Lets try and do static imports if we can

backend/hexa/assistant/schema/mutations.py
```diff
@@ -12,6 +12,26 @@
 assistant_mutations = MutationType()
 
 
+def _resolve_linked_object(user, linked_object_type, linked_object_id):
+    """
+    Maps a (type, id) pair from the API to a model instance and the instruction set
+    that should be used for conversations about that object type.
+    Returns (linked_object, instruction_set) or raises ValueError/DoesNotExist.
+    """
+    from hexa.pipelines.models import Pipeline
```
yolanfery: Do we need static import ?

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -50,16 +54,27 @@ const FileTreeNode = ({
...
   node: FileNode;
   level?: number;
   selectedFile: FileNode | null;
   setSelectedFile: (file: FileNode | null) => void;
   modifiedFiles: Map<string, string>;
+  proposedByKey: Map<string, string>;
 }) => {
-  const [isExpanded, setIsExpanded] = useState(false);
+  const [isExpanded, setIsExpanded] = useState(
+    node.id.startsWith("proposed-dir-"),
```
yolanfery: Not a fan of inferring stuff, if proposed-dir disappear this logic will be broken silently, we might want to improve this

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -140,12 +161,82 @@ export const FilesEditor = ({
...
+    }
+
+    return result;
+  }, [proposedFiles, flatFiles]);
+
+  const augmentedFlatFiles = useMemo(
+    () => [...flatFiles, ...virtualFiles],
+    [flatFiles, virtualFiles],
+  );
+
```
yolanfery: This file becomes big and complex, we might want consider some refactoring/testing/hook usage 

I am really worried about maintenance of all of this FE

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -50,16 +54,27 @@ const FileTreeNode = ({
...
-  const [isExpanded, setIsExpanded] = useState(false);
+  const [isExpanded, setIsExpanded] = useState(
+    node.id.startsWith("proposed-dir-"),
+  );
   const isSelected = selectedFile?.id === node.id;
-  const isModified = modifiedFiles.has(node.id);
+
+  const proposedContent =
+    node.type === "file"
+      ? (proposedByKey.get(node.path) ?? proposedByKey.get(node.name))
```
yolanfery: Can we decide (and stick to it) if we use the name or the path as the key ? It adds complexity to the code to always nullish coalescing

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -50,16 +54,27 @@ const FileTreeNode = ({
...
+      : undefined;
+  const isProposed =
+    proposedContent !== undefined && proposedContent !== (node.content ?? "");
```
yolanfery: Maybe we can follow a pattern like for 

`modifiedFiles.has(node.id)` with the `proposedByKey` dict

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
```diff
@@ -352,10 +476,34 @@ export const FilesEditor = ({
               {isClient ? (
                 <div className="absolute inset-0">
                   <CodeMirror
+                    key={selectedFile.id}
```
yolanfery: A bit worried about performances of always remounting, do we really need this ?

yolanfery (changes requested):
Amazing, great job !  🚀

Tested and reviewed. 

Few things : 

- I was not able to get the diff for whatever reason : 

Image

- At some point Anthropic was down and there was no error message in the UI... We need to handle sync errors and reflecting the status in the UI somehow
- I am really worried about the Frontend. This is a complex feature and it is reflecting in the code of course. But I think we could take a bit more time to separate the logic in different files, maybe do some testing (I know we don't tend to test the fronted usually but for this feature with edge cases it might make sense). I am mostly worried about extensibility challenges and regressions we will introduce due to the complexity

backend/hexa/assistant/agents/create_pipeline_agent.py
mrivar: I thought of it and tried, but it seems to need a little bit more work thn just plugging it in.
I tested it and it didn't work exactly as I expected it to. It probably needs some prompting engineering and some trial and error.
I suggest putting it as a separate task and handling it sooner than later for sure.

backend/hexa/assistant/agents/edit_pipeline_agent.py
mrivar: Yeah you're right, we can probably just mix the proposed files with the untouched files manually to save on that. Well spotted!

backend/hexa/assistant/agents/edit_pipeline_agent.py
mrivar: true, I don't know why I had that lazily loaded

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
mrivar: You're absolutely right, it's becoming a super class. I'm gonna look for ways to refactor it in a more nicely way. Thanks for bringing it up!

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
mrivar: The main issue is that CodeMirror doesn't support server-side rendering as it accesses browser-only globals at import/init time. Thus, it needs to load lazily (in this outdated version this was applied with the isClient effect, but it was a bit hacky, so I changed it to a dynamic load with ssr disabled).

Edit: according to Claude, CodeMirror uses:
  - `document` to create DOM elements for the editor view                                                                                                                                 
  - `window` for event listeners and resize observers                                                                                                                                     
  - `navigator` for browser detection (keybindings differ between OS)

Edit 2: Refactored into CodeMirrorClient. Now that component has the full logic of being loaded dynamically to avoid SSR issues.

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
mrivar: changed for a isProposed boolean property

backend/hexa/assistant/agents/edit_pipeline_agent.py
mrivar: solved: agent just gives out the proposed changes / deleted files, and the mix is done manually, saving a lot of tokens in pipelines with many files.

frontend/src/workspaces/features/FilesEditor/FilesEditor.tsx
mrivar: refactored into hooks and smaller files (types, hooks, separated FileTree logic, CodeMirrorClient, and FileEditorPane)

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -0,0 +1,90 @@
...
+    pipeline: Pipeline,
+    modified_files: list[ProposedFile] | None = None,
+    deleted_files: list[str] | None = None,
+) -> dict:
+    """Propose a new version of the pipeline.
+
+    Always call this tool with the files you modified or created in modified_files.
+    Each entry must have a 'name' (e.g. 'pipeline.py') and 'content' (the full file text).
+    List any files to remove in deleted_files.
+    Unchanged files are preserved automatically.
```
yolanfery, line 26: Nice

frontend/src/workspaces/features/FilesEditor/useFilesEditorState.ts
```diff
@@ -0,0 +1,301 @@
...
+    return new Set(
+      flatFiles
+        .filter((f) => f.type === FileType.File && !proposedNames.has(f.path))
+        .map((f) => f.path),
+    );
+  }, [proposedFiles, flatFiles]);
+
+  // Maps file path → proposed content, but only for files that differ from the current version.
+  // Unchanged files are excluded so they don't trigger diff highlighting or amber dots.
+  // Deleted files are included with "" so the diff renders all lines as removed.
```
yolanfery, line 129: Nice trick

yolanfery (approved):
This is 🔥

Well done and thanks for the refactoring

Just 2 small UI fixes : 

- The file info for new files is not correct

Image

- [Minor suggestion] Can you please ensure the header heights are the same between the panels for consistency ? 

Image

PR #1729, opened 2026-04-09
Hexa 1428 allow users to host webapp on custom domain

DimitriKwihangana:
Added custom domain field to webapp modal and custom middleware that rewrites requests on a custom domain to webapps/uuid before django routing. Also added ADDITIONAL_ALLOWED_HOSTS in .env.dist
NB: not sure what will be changed on the infra but I think we will need  a manual cert-manager entry per domain


backend/hexa/webapps/middlewares.py
```diff
@@ -0,0 +1,30 @@
+from hexa.webapps.models import Webapp
+
+
+class CustomDomainMiddleware:
+    """
+    Rewrites requests arriving on a webapp's custom domain to the internal
+    /webapps/<id>/ path so the serve_webapp view can handle them normally.
```
yolanfery: The path `/webapps/<id>/` is now deprecated/removed for security reason, we need another mechanism

backend/hexa/webapps/middlewares.py
```diff
@@ -0,0 +1,30 @@
...
+    Host header; in Phase 2 this middleware will handle that itself.
+    """
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        host = request.META.get("HTTP_HOST", "").split(":")[0].lower()
+        try:
+            webapp = Webapp.objects.get(custom_domain=host, is_public=True)
```
yolanfery: I am not totally sure about this comment : seems like we will query webapps for ALL queries hitting our backend since we are doing this as a middleware. We can probably do something smarter/more efficient like checking that the URL is unusual or something else first. This would allow regular queries to flow naturally

backend/hexa/webapps/middlewares.py
```diff
@@ -0,0 +1,30 @@
+from hexa.webapps.models import Webapp
+
+
+class CustomDomainMiddleware:
```
yolanfery: I am proposing to add a unit test and/or integration test to cover this logic so that we verify it and don't forget about it during refactors

yolanfery (changes requested):

backend/hexa/webapps/middlewares.py
```diff
@@ -205,3 +205,49 @@ def middleware(request: HttpRequest):
...
+
+    def __call__(self, request: HttpRequest):
+        host = request.META.get("HTTP_HOST", "").split(":")[0].lower()
+
+        # Skip the DB query for known OpenHEXA hosts — only custom domains are unusual
+        webapps_domain = getattr(settings, "WEBAPPS_DOMAIN", None)
+        if host == settings.BASE_HOSTNAME or (
+            webapps_domain and host.endswith(f".{webapps_domain}")
+        ):
+            return self.get_response(request)
```
yolanfery: Cool 👌

backend/hexa/webapps/middlewares.py
```diff
@@ -205,3 +205,49 @@ def middleware(request: HttpRequest):
...
+        else:
+            response = _serve_iframe_webapp(webapp)
+
+        del response["X-Frame-Options"]
+        frame_ancestors = f"'self' {settings.BASE_URL}"
+        if hasattr(settings, "NEW_FRONTEND_DOMAIN"):
+            frame_ancestors += f" {settings.NEW_FRONTEND_DOMAIN}"
+        response["Content-Security-Policy"] = f"frame-ancestors {frame_ancestors}"
+
+        return response
```
yolanfery: This logic is repeated between the 2 middlewares, we need to avoid the code duplication (probably with a common function call, and not sure if this should be a model function or not, up to you to think and decide) 

(Also sorry but I did a commit this morning that changed the logic slightly and this need to be merged in)

backend/hexa/webapps/middlewares.py
```diff
@@ -224,12 +232,50 @@ def middleware(request: HttpRequest):
...
+
+        if request.path.startswith("/graphql/"):
+            return HttpResponseNotFound("Not available")
+
+        if webapp.type == Webapp.WebappType.STATIC:
+            response = _serve_static_webapp(webapp, request)
+        elif webapp.type == Webapp.WebappType.SUPERSET:
+            response = _serve_superset_webapp(request, webapp)
+        else:
+            response = _serve_iframe_webapp(webapp)
```
yolanfery: This logic is still duplicated. This will cause drift, errors and maintenance burden. 

For instance there is already an issue : 

Above we have 

```
        if webapp.type == Webapp.WebappType.STATIC:
            response = _serve_static_webapp(webapp, request)
            if show_powered_by:
                _inject_powered_by_banner(response)
```

And here we don't have the banner logic 

Things should be consistent. We should either call/route to the other middleware or use a helper function


Let me know if we need to go over this together

yolanfery (changes requested):

backend/hexa/webapps/middlewares.py
DimitriKwihangana: I think I'm learning more about the clean codes! Thanks @yolanfery . I've taken an approach putting together a gelper that odes the dipatch of web apps. called in both middlewares. 

Added the banner  and also converted the middleware to  a function for consistency. We can still talk on this together just felt you already had a lot on your plate.

backend/hexa/webapps/middlewares.py
```diff
@@ -215,20 +235,44 @@ def middleware(request: HttpRequest):
...
+        return response
+
+    return middleware
+
+
+def custom_domain_middleware(get_response):
+    """Intercepts requests arriving on a webapp's custom domain and serves the webapp
+    content directly bypassing normal Django URL routing.
+
+    Only public webapps can be served via a custom domain.
```
yolanfery: I don't see a reason why this should be the case 

It simplifies code but I am not sure we should enforce this false limitation. Conceptually, custom domain is just routing, so this should not affect functional requirements

WDYT ?

backend/hexa/webapps/middlewares.py
```diff
@@ -215,20 +235,44 @@ def middleware(request: HttpRequest):
...
+            return get_response(request)
+
+        request.webapp = webapp
+
+        if request.path.startswith("/graphql/"):
+            return HttpResponseNotFound("Not available")
+
+        show_powered_by = webapp.show_powered_by
+        response = _dispatch_webapp_response(request, webapp, show_powered_by)
+        _set_csp_frame_ancestors(response)
```
yolanfery: Already way better, thanks for the refactor

backend/hexa/webapps/middlewares.py
DimitriKwihangana: You're right, it's a false limitation  I'm trying to remove the is_public =True filter and pu the same token exchange  flow for private webapps.

yolanfery: Yes that would be ideal, thanks, lets not forget to not code duplicate here as well

yolanfery (approved):
Great job, you removed the logic duplication to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) ! 

Looks good to me from a code perspective

backend/hexa/webapps/models.py
```diff
@@ -140,6 +140,7 @@ class WebappType(models.TextChoices):
         unique=True,
         validators=[validate_subdomain],
     )
+    custom_domain = models.CharField(max_length=253, blank=True, null=True, unique=True)
```
bramj: I know this is an admin-only feature for the moment, but I guess we can already validate that it's a valid domain here, using something like this:

```python
from django.core.validators import DomainNameValidator
```

Source: https://docs.djangoproject.com/en/6.0/ref/validators/#domainnamevalidator

.env.dist
```diff
@@ -70,6 +70,15 @@ POSTGRES_DB=hexa-app
...
 POSTGRES_PASSWORD=hexa-app
 
+## Custom domains for web apps
+##############################
+
+# Comma-separated list of custom domains attached to public web apps.
+# Each domain set on a Webapp via Django admin must also be listed here so
+# Django accepts the Host header. Example:
+# ADDITIONAL_ALLOWED_HOSTS=carte-sanitaire.gouv.ne,dashboard.example.org
+# ADDITIONAL_ALLOWED_HOSTS=
```
bramj, line 80: 👍

backend/config/settings/base.py
```diff
@@ -274,6 +274,7 @@
     "corsheaders.middleware.CorsMiddleware",
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
+    "hexa.webapps.middlewares.custom_domain_middleware",
```
bramj: [Nitpick] Is there a reason why this middleware is so high up instead of being together with the other webapps one?

Also keeping in mind that middlewares are executed in order, and an early return here can mean we skip for example the CSRF protection.

backend/hexa/webapps/middlewares.py
```diff
@@ -69,6 +77,18 @@ def _inject_powered_by_banner(response):
     return response
 
 
+def _dispatch_webapp_response(request, webapp, show_powered_by=False):
+    if webapp.type == Webapp.WebappType.STATIC:
+        response = _serve_static_webapp(webapp, request)
+        if show_powered_by:
+            _inject_powered_by_banner(response)
```
bramj:
```suggestion
            response = _inject_powered_by_banner(response)
```

bramj (changes requested):
Nice work @DimitriKwihangana , especially on the refactoring 👍 

Just a few minor comments.

backend/config/settings/base.py
DimitriKwihangana: I don't there is a concrete reason for putting it up. I've moved it close the subdomains one

bramj (approved):
Nice one! 👍

PR #1731, opened 2026-04-10
feat: S3 file storage backend (HEXA-1585)

mrivar:
Add support for S3 file storage

## Changes

- Add S3 env vars in .env.dist
- Create S3 Storage
- Add tests using moto mock
- Tests refactor:
  - Created a base battery of tests (and removed tests that were thus duplicated afterwards). Advantages of this are:
    - Now other storage tests can inherit from it and ensure full storage testing.
    - Avoid repeating tests along each storage test class.
    - Now adding new storage tests is easier.
    - When a new storage is created, we will ensure having a whole set of tests for it to ensure proper development.
  - In order to do this, I needed to do some refactoring for consistency in some storage classes, a summary of these are:
    - Azure: raise custom `NotFound` exceptions instead of Azure SDK ones.
    - GCP: same thing.
    - GCP Mock: Fixed it so it is fully compatible with the original SDK it is mocking.
    - Generally: added typing for clarity and better code navigation


.env.dist
```diff
@@ -180,12 +181,21 @@ WORKSPACE_STORAGE_BACKEND_GCS_SERVICE_ACCOUNT_KEY=
...
 WORKSPACE_STORAGE_BACKEND_AWS_BUCKET_REGION=
+# Internal endpoint for server-side API calls — set for MinIO or custom S3 (e.g. http://minio:9000)
+WORKSPACE_STORAGE_BACKEND_AWS_ENDPOINT_URL=
+# Public endpoint used in presigned URLs returned to clients — set when internal and public hosts differ
+# Falls back to WORKSPACE_STORAGE_BACKEND_AWS_ENDPOINT_URL when not set
+WORKSPACE_STORAGE_BACKEND_AWS_PUBLIC_ENDPOINT_URL=
+# Optional IAM role ARN to assume when generating short-lived notebook credentials via STS.
+# Falls back to static access key/secret when not set (required for MinIO or simple setups).
+# S3 Files (NFS): if your infra mounts the bucket via NFS, set WORKSPACE_STORAGE_ENGINE_S3_NFS_MOUNT_POINT
+# in the notebook environment and the notebook will prefer the mount over SDK access.
```
yolanfery: Not sure I get this

We reference WORKSPACE_STORAGE_ENGINE_S3_NFS_MOUNT_POINT but it does not seem to be defined elsewhere

backend/config/settings/base.py
```diff
@@ -527,6 +527,37 @@
...
+    if not os.environ.get("WORKSPACE_STORAGE_BACKEND_AWS_ACCESS_KEY_ID"):
+        raise ValueError(
+            "WORKSPACE_STORAGE_BACKEND_AWS_ACCESS_KEY_ID environment variable is required for S3 storage backend."
+        )
+    if not os.environ.get("WORKSPACE_STORAGE_BACKEND_AWS_SECRET_ACCESS_KEY"):
+        raise ValueError(
+            "WORKSPACE_STORAGE_BACKEND_AWS_SECRET_ACCESS_KEY environment variable is required for S3 storage backend."
+        )
+    WORKSPACE_STORAGE_BACKEND = {
+        "engine": "hexa.files.backends.s3.S3Storage",
```
yolanfery, line 540: Did you try S3 Files instead of going through the SDK ?

backend/hexa/files/tests/backends/test_s3.py
```diff
@@ -0,0 +1,367 @@
+import io
+from unittest.mock import patch
+from urllib.parse import parse_qs, unquote, urlparse
+
+from moto import mock_aws
```
yolanfery, line 5: That's a nice gift from AWS :)

yolanfery (commented):
Reviewed the code 👍
Always hard to review such integrations, it's mostly plumbing but if it works it should be good
The tests are a good addition though (even if it mostly covers the AWS mock)

Just a question about S3 Files, I feel like we need to try (harder ?)

backend/hexa/files/tests/backends/test_s3.py
mrivar: Yeah, pretty amazing for testing! I've used minio for testing s3 storages before but this is more handy tbh

.env.dist
mrivar: Yeah it is more of a comment regarding infra, on how we could mount the AWS S3 Files system.
However, actually implementing it and so on would take different development and what not.
I'll go deeper into it in the next comment.

backend/config/settings/base.py
mrivar: So, regarding the AWS S3 Files, there are some things to consider. What the S3 Files do is expose a bucket as a POSIX NFS filesystem. Thus, instead of `boto3` API calls, you could use Python file I/O (`open()`, `os.path`, etc).

Now, this could make sense in something like our notebooks or pipelines. They would directly connect to S3 and use the typical Python I/O calls instead of having to go through the pain of an API and what not. Also, the connection would stay open and it would be quicker.

However, for our S3 Storage system the situation is different. As we only access it occasionally for concrete operaions (upload a file, list a directory, get a presigned URL, etc) using an API makes more sense than having an open direct connection to a bucket. Also, it gives us some advantages over S3 Files (listed by claude):
- No infrastructure dependency — just credentials and an endpoint
- Presigned URLs are an S3 SDK concept with no NFS equivalent
- Works in local dev/Docker without any mount setup
- The overhead of an API call per operation is irrelevant at this scale

Now for notebooks and pipelines, that is a conversation that might be worth opening. However, this would require development in `openhexa-sdk-python` and so on, but it would be a nice-to-have for the future.

mrivar:
> Reviewed the code 👍 Always hard to review such integrations, it's mostly plumbing but if it works it should be good The tests are a good addition though (even if it mostly covers the AWS mock)
> 
> Just a question about S3 Files, I feel like we need to try (harder ?)

Definitely plumbing haha, that's the perfect metaphor 🤣. An abstraction over an abstraction..

Regarding the tests...I have to say I'm sorry! I couldn't help myself and ended up doing some refactoring over the tests. It is more of a pain to review, but I think we will end up with a more robust testing suite. It helped me identify some inconsistencies over the storages as well (not big, but nice to have them all consistent or mostly consistent now). More info about it in the [PR description](https://github.com/BLSQ/openhexa-app/pull/1731#issue-4238635900)!

And regarding the S3 Files, I added more info [in the comments](https://github.com/BLSQ/openhexa-app/pull/1731#discussion_r3080091105) about my opinion on it after doing some research. Open to debate 😁.  We can keep the convo going there.

backend/config/settings/base.py
yolanfery: Fair points, I agree with you

Thanks for diving deeper into this and articulating the reasoning

backend/hexa/files/tests/backends/base.py
```diff
@@ -0,0 +1,159 @@
...
+
+from hexa.files.backends.base import Storage
+
+BUCKET = "test-bucket"
+
+
+class StorageTestMixin:
+    """
+    Contract tests for any Storage backend. Inherit alongside TestCase and
+    assign self.storage in setUp.
```
yolanfery, line 11: Great idea !!

backend/hexa/files/tests/backends/test_s3.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def tearDown(self):
+        self.mock.stop()
+        super().tearDown()
+
+    def test_delete_bucket(self):
+        self.storage.create_bucket(BUCKET)
+        self.storage.save_object(BUCKET, "file.txt", io.BytesIO(b"data"))
+        self.storage.delete_bucket(BUCKET, force=True)
+        self.assertFalse(self.storage.bucket_exists(BUCKET))
```
yolanfery, line 40: Do we still need those kind of tests here ?They are not in the mixin ?

backend/hexa/files/tests/mocks/bucket.py
```diff
@@ -1,56 +1,65 @@
...
+from typing import TYPE_CHECKING
+
 from google.cloud.storage._helpers import _validate_name
 
 from hexa.files.backends.exceptions import NotFound
 
 from .blob import MockBlob
 
+if TYPE_CHECKING:
+    from .client import MockClient
```
yolanfery, line 12: Why do we need to introduce this kind of pattern here ? 

It might be just me but I prefer not having the type in the file than having this, WDYT ?

backend/hexa/files/tests/mocks/blob.py
```diff
@@ -1,22 +1,26 @@
-from unittest.mock import Mock
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
 
 from google.cloud._helpers import _bytes_to_unicode
+from google.cloud.exceptions import NotFound
+
+if TYPE_CHECKING:
+    from .bucket import MockBucket
```
yolanfery, line 9: Why do we need to introduce this kind of pattern here ? 

It might be just me but I prefer not having the type in the file than having this, WDYT ?

yolanfery (commented):
Great idea on the refactoring of the tests on the base interface instead of duplication for each child class ! 👌

Jus 4 minor comments

yolanfery (approved):

backend/hexa/files/tests/backends/test_s3.py
mrivar: Yeah I just moved the ones that were already shared across the current tests, but we can add some more for sure

backend/hexa/files/tests/mocks/bucket.py
mrivar: tbh, I prefer it like this, as it makes it easier to navigate around the code using the IDE, and also knowing right away what is what instead of having to guess. I guess I got too used to the comfort of heavily typed languages after my time with Kotlin 😁
It's true that the `TYPE_CHECKING` might be slightly annoying, but being up in the imports it's a bit more "hidden"

yolanfery: Ok fair point

I don't like this because (since it's "hidden") then it's very easy to start using `MockClient` in the logic and this will trigger a runtime issue when the branch is not True

But OK no strong opinion on this one, not blocking for this

PR #1732, opened 2026-04-11
feat:show run parameters on pipeline runs list page (HEXA-1586)

DimitriKwihangana:
the PR shows run parameters on the pipeline runs list page, so users can distinguish between runs at a glance without opening each one . Each row has the parameters used, with a tooltip on hover showing the full list.

## Changes

- New Row of parameters and tooltip on the parameter cell
-  Add config and version.parameters to the WorkspacePipelineRunsPage GraphQL query

## Screenshots / screencast
image


frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
```diff
@@ -18,8 +20,51 @@ import {
...
 import PipelineLayout from "workspaces/layouts/PipelineLayout";
 
+function formatParamValue(entry: PipelineParameter & { value: any }): string {
+  if (entry.type === "bool") return entry.value ? "✓" : "✗";
+  if (entry.type === "secret" && entry.value) return "••••••";
+  if (isNil(entry.value)) return "-";
+  if (entry.multiple && Array.isArray(entry.value)) return entry.value.join(", ");
+  if (isConnectionParameter(entry.type)) return String(entry.value);
+  return String(entry.value);
+}
```
Comment (yolanfery) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx]: [Minor] Maybe we can do something more typed here using autogenerated types (https://github.com/BLSQ/openhexa-app/blob/90aaadf50b1045a8a2de8dbf00b642021d6f04bd/frontend/src/graphql/types.ts#L3860-L3874)
To be more defensive and clear

Also simple unit tests for this function for each type could help

yolanfery (changes requested):
Great !

Some visual feedback : 

I would try to ensure that truncation is not happening midline like this : 

Image

And we might not need tooltips when the content is not truncated, and a tooltip (and clear indicator that there is truncation) when its truncated

DimitriKwihangana:
> Great !
> 
> Some visual feedback :
> 
> I would try to ensure that truncation is not happening midline like this :
> 
> Image
> And we might not need tooltips when the content is not truncated, and a tooltip (and clear indicator that there is truncation) when its truncated

Thanks  @yolanfery!  I included that as a hint that users can hover to explore additional parameters, but I agree it doesn't look that good.

frontend/src/pipelines/helpers/format.test.ts
```diff
@@ -0,0 +1,90 @@
+import { ParameterType } from "graphql/types";
+import { formatParamValue } from "./format";
+
```
yolanfery, line 3: Thanks for adding tests, I think it's helpful !

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
```diff
@@ -8,18 +8,75 @@ import Time from "core/components/Time";
...
+            <span className="break-words">{formatParamValue(p)}</span>
+            {isLast && shouldShowToggle && !isExpanded && (
+              <button
+                onClick={(e) => {
+                  e.stopPropagation();
+                  setIsExpanded(true);
+                }}
+                className="shrink-0 cursor-pointer text-blue-600 hover:text-blue-800 hover:underline"
+              >
+                {t("+{{remainingCount}} more", { remainingCount })}
```
Comment (yolanfery) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx]: What do you think about having this in a new separate line instead of at the end fo the last line ? 

Also, can we ensure the table column sizes are not changing when opening closing this ?

yolanfery (changes requested):

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
Comment (DimitriKwihangana) [reply frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx]: It is better, putting on the new line.

yolanfery (approved):
Very nice ! Well done @DimitriKwihangana

PR #1741, opened 2026-04-15
 feat: select a specific version for scheduled pipelines runs(Hexa 1589)

DimitriKwihangana:
## Changes

-  Added scheduled_pipeline_version field
- Scheduler now picks pipeline.scheduled_pipeline_version otherwise  normal flow(latest version)

## Screenshots / screencast

image
image


backend/hexa/pipelines/tests/test_models.py
```diff
@@ -994,3 +997,109 @@ def test_run_uses_subscription_resource_limits(self):
...
+                schedule="0 12 * * *",
+                scheduled_pipeline_version_id=str(self.VERSION_1.id),
+            )
+            self.PIPELINE.refresh_from_db()
+            self.assertEqual(self.PIPELINE.schedule, "0 12 * * *")
+            self.assertEqual(self.PIPELINE.scheduled_pipeline_version, self.VERSION_1)
+        finally:
+            self.PIPELINE.schedule = None
+            self.PIPELINE.scheduled_pipeline_version = None
+            self.PIPELINE.save()
```
yolanfery: Why do we have try catch here ?

backend/hexa/pipelines/models.py
```diff
@@ -573,9 +581,22 @@ def upload_new_version(
     def update_if_has_perm(self, principal: User, **kwargs):
         if not principal.has_perm("pipelines.update_pipeline", self):
             raise PermissionDenied
+
+        # Resolve which version will be pinned for scheduled runs after this update.
+        if "scheduled_pipeline_version_id" in kwargs:
+            version_id = kwargs.pop("scheduled_pipeline_version_id")
+            if version_id:
```
yolanfery: Why do we need this check, can we simplify and check it at the previous if statement also ?

backend/hexa/pipelines/models.py
```diff
@@ -489,7 +496,8 @@ def is_schedulable(self):
         if self.type == PipelineType.NOTEBOOK:
             return True
         elif self.type == PipelineType.ZIPFILE:
-            return self.last_version and self.last_version.is_schedulable
+            version = self.scheduled_pipeline_version or self.last_version
```
yolanfery: Instead of duplicating this pattern everywhere, we might want to reduce duplication and define a @property in the Pipeline class to get the version_to_run that just returns self.scheduled_pipeline_version or self.last_version, it's a bit more idiomatic and reduce duplication

frontend/src/workspaces/graphql/queries.graphql
```diff
@@ -238,6 +239,31 @@ query WorkspacePipelineNotificationsPage(
...
+      id
+      versionNumber
+      versionName
+      parameters {
+        code
+        required
+      }
+      config
+    }
+    versions(perPage: 50) {
```
yolanfery: What happens when the pipeline has more than 50 versions ? 

If we return the 50 latest, it might be good enough (users want to run recent versions I guess), we could even reduce to 10 for instance

But we need to ensure this

yolanfery (changes requested):
Well done ! Just a few comments

bramj:
@yolanfery @DimitriKwihangana I'm not convinced this is a feature we really want for the following reasons:
- It feels all to easy to "fix" a schedule to a specific version and then forget about it. Some months/versions later, you wonder why the pipeline runs are not showing your latest changes.
- I think moving to a `draft` -> `published` system makes more sense to manage pipeline versions. In addition to this feature, I feel that would make things too complex.

Let's discuss during the weekly // @mcrimi

yolanfery:
> @yolanfery @DimitriKwihangana I'm not convinced this is a feature we really want for the following reasons:
> 
> * It feels all to easy to "fix" a schedule to a specific version and then forget about it. Some months/versions later, you wonder why the pipeline runs are not showing your latest changes.
> * I think moving to a `draft` -> `published` system makes more sense to manage pipeline versions. In addition to this feature, I feel that would make things too complex.
> 
> Let's discuss during the weekly // @mcrimi

Not totally aligned with this, I think it's a common pattern to build versions and then choose which one to use 
Maybe we just want to give more visibility on which version is running to avoid this issue of "forgetting about it"

Let's discuss yes ☺️

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/notifications.tsx
```diff
@@ -61,7 +61,12 @@ const WorkspacePipelineNotificationsPage: NextPageWithLayout = (
 
   const { workspace, pipeline } = data;
 
-  const versionOptions = pipeline.versions?.items ?? [];
+  const versionItems = pipeline.versions?.items ?? [];
+  const pinned = pipeline.scheduledPipelineVersion;
+  const versionOptions =
+    pinned && !versionItems.some((v) => v.id === pinned.id)
+      ? [pinned, ...versionItems]
```
Comment (yolanfery) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/notifications.tsx:68]: Cool trick !

backend/hexa/pipelines/management/commands/pipelines_scheduler.py
```diff
@@ -55,6 +55,9 @@ def handle(self, *args, **options):
                     logger.debug(f"sleep before run: {real_delay}")
                     sleep(real_delay)
 
+                pipeline_version = (
+                    pipeline.scheduled_pipeline_version or pipeline.last_version
+                )
```
yolanfery: We have a new abstraction for the version to run, let's use it 😜

yolanfery (approved):
Well done !
Sorry, still one small comment left

backend/hexa/pipelines/management/commands/pipelines_scheduler.py
yolanfery: (Everywhere it's needed)

PR #1750, opened 2026-04-20
feat: SSE in agent loop (HEXA-1554)

mrivar:
Implement SSE in the main agent loop, in favor of fire-and-wait GraphQL mutation.

## Changes

Implemented SSE in the main `ChatPane` and the `CreatePipelineUsingAI` modal. Thus, It can be viewed in `/assistant`, edit pipeline chat, and in the create pipeline modal.

- Backend:
  - POST SSE stream endpoint (/assistant/conversations/<id>/stream/)
  - Agent `run_stream` that runs the agent as an async task
  - Agent `run` refactores to sue `run_stream` synchronously to avoid repeating code
  - Refactored agent test suite (split the monolithic `test_agent.py` into focused per-concern files)
- Frontend:
  - `useStreamingFetch`: frontend hook that consumes the SSE stream over a POST fetch request
  - `useSSE`: Added close() method and fixed optimistic `isConnected` (now it is checked)
  - `useTypeWriter` and `useWordDrain`: Hooks for natural writing from agent in conversations
  - `ChatPane`:
    - Added word-by-word text drain queue (word drain hook)
    - Better visualisation of errors
  - `PipelineEditChatPane`: Conversation name got from SSE and written using the typewriter hook.
  - `PipelineFileEditor`: Get proposed files from SSE
  - `CreatePipelineUsingAI`: Added SSE, along with a set of steps that get updated using SSE. Added better visualization of errors and agent feedback.

## How/what to test

- Create a Pipeline using AI:
  - Test how it shows errors (e.g.: put a prompt with not enough information to create a pipeline)
  - Test creating a pipeline
- Edit pipeline:
  - All of the 3 changes (adding a conversation name, the agent answering, and the code proposal) should appear smoothly on the screen as they arrive via SSE
  - Errors (AI provider down, error via SSE) should show in a more visible way (used to fail silently)

## Screenshots / screencast

Creating a pipeline:
Screenshot from 2026-04-24 09-51-28

Creating a pipeline error:
image

Create a pipeline error, AI feedback:
image

Create a pipeline successfully:
Screenshot from 2026-04-24 09-51-34

Edit a pipeline, AI provider error:
image

Edit a pipeline, SSE error:
image


DimitriKwihangana (approved):
It looks really good, @mrivar. I was just wondering since we may not be using the sendAssistantMessage mutation anymore (though I might be mistaken), would it make sense to remove it from the app entirely?

mrivar:
> It looks really good, @mrivar. I was just wondering since we may not be using the sendAssistantMessage mutation anymore (though I might be mistaken), would it make sense to remove it from the app entirely?

You're probably right, I'm going to check if we use it anywhere else

backend/hexa/assistant/agents/base.py
```diff
@@ -90,98 +107,205 @@ def _context(self) -> dict:
...
         )
 
-        result = self.agent.run_sync(user_input, message_history=history)
-        logger.info(
-            "agent.run: LLM call complete, new_messages=%d", len(result.new_messages())
-        )
+        try:
+            precomputed_naming: tuple[str, RunUsage] | None = None
+            if is_first_message:
+                precomputed_naming = await self._generate_conversation_name(user_input)
```
bramj: I'll create a ticket to make sure we run the naming agent in parallel

frontend/src/assistant/features/ChatPane/index.tsx
```diff
@@ -79,18 +94,68 @@ export default function ChatPane({
...
+    localConversationIdRef.current = localConversationId;
+  }, [localConversationId]);
+
+  const handleDrained = useCallback(() => {
+    setPendingUserMessage(null);
+    setCurrentPage(1);
+    if (localConversationIdRef.current) {
+      apolloClient.refetchQueries({
+        include: [AssistantConversationMessagesDocument],
+      });
```
bramj, line 113: Is the refetch necessary? I suppose the conversation was already received over the stream?

frontend/src/core/hooks/useTypewriter.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useRef, useState } from "react";
+
+type Options = {
+  interval?: number;
+};
+
+function useTypewriter(value: string | null, { interval = 35 }: Options = {}): string | null {
```
bramj, line 18: This is a cool effect on the conversation title! 👏 

Maybe add some documentation for this sort of component so it's easy to understand what it does?

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/CreatePipelineUsingAI.tsx
```diff
@@ -51,15 +136,70 @@ const CreatePipelineUsingAI = ({ form }: CreatePipelineUsingAIProps) => {
...
+                      <ChevronDownIcon
+                        className={`h-3.5 w-3.5 transition-transform duration-200 ${showAgentResponse ? "rotate-180" : ""}`}
+                      />
+                      {t("AI feedback")}
+                    </button>
+                    {showAgentResponse && (
+                      <div className="mt-2 rounded-lg border border-gray-200 bg-white px-3 py-2 text-sm text-gray-600">
+                        <div className="prose prose-sm prose-gray max-w-none">
+                          <ReactMarkdown remarkPlugins={[remarkGfm]}>
+                            {form.agentResponse}
```
bramj, line 194: This is really great! I just had to use this:

Image

And actually, I didn't realize at first that the agent response was there. 

I think we can **make it more prominent,** like for example: instead of hiding the box behind a chevron, make the box always visible, but smaller and expandable. Actually, I think the UX in Slack is great when you have a snippet (e.g. the daily standup notes in the #blsq_developers channel)

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/CreatePipelineUsingAI.tsx
```diff
@@ -51,15 +136,70 @@ const CreatePipelineUsingAI = ({ form }: CreatePipelineUsingAIProps) => {
...
-          </p>
+
+        {isActive && (
+          <div className="rounded-xl border border-gray-200 bg-gray-50 p-4 space-y-3">
+            <Step
+              label={t("Generating pipeline code")}
+              status={generatingStatus()}
+            />
+            <Step
+              label={t("Creating pipeline")}
```
mrivar, line 168: add pipeline name?

backend/hexa/assistant/agents/base.py
yolanfery: Can't we make this non-blocking and run it in parallel ?

backend/hexa/assistant/tests/agents/test_base_agent_run.py
```diff
@@ -0,0 +1,156 @@
...
+    def setUp(self):
+        self.conversation = Conversation.objects.create(
+            user=self.user,
+            workspace=self.workspace,
+            instruction_set=InstructionSet.GENERAL,
+        )
+
+    def test_run_saves_user_message(self):
+        with _patch_builder(TestModel(custom_output_text="Hello!")):
+            agent = BaseAgent(self.conversation)
```
yolanfery: This looks like a smell that the agent needs a model as a constructor parameter, to allow dependency injection

backend/hexa/assistant/tests/agents/test_base_agent_run.py
```diff
@@ -0,0 +1,156 @@
...
+        self.conversation = Conversation.objects.create(
+            user=self.user,
+            workspace=self.workspace,
+            instruction_set=InstructionSet.GENERAL,
+        )
+
+    def test_tool_call_creates_tool_invocation_record(self):
+        model = _make_tool_call_model("_fake_tool", {"arg": "hello"})
+        with _patch_builder(model):
+            agent = _AgentWithFakeTool(self.conversation)
```
yolanfery: This looks like another smell, and therefore confirmation that an agent needs a model as a parameter

backend/hexa/assistant/tests/test_views.py
```diff
@@ -0,0 +1,139 @@
...
+            events.append(current)
+            current = {}
+    return events
+
+
+async def _collect_async_stream(streaming_content) -> bytes:
+    chunks = []
+    async for chunk in streaming_content:
+        chunks.append(chunk)
+    return b"".join(chunks)
```
yolanfery: I have seen this logic elsewhere in the PR and with different implementation, maybe we want to define a helper

frontend/src/assistant/features/ChatPane/index.tsx
mrivar: Yeah I'd say so.
The information coming from the stream is great and makes it visible for the user in real time, but it's just "plain" data (text_delta's) while the information coming from the refetch is kind of the "real state" of the data after persisting everything; including  IDs, timestamps, token counts, tool invocations, etc, which might prove useful and keeps everything consistent.

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/CreatePipelineUsingAI.tsx
mrivar: Good one!, I'll quickly change that behaviour

frontend/src/core/hooks/useTypewriter.ts
mrivar: Done!

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/useAIForm.ts
```diff
@@ -1,88 +1,245 @@
...
+    abort();
+    setPhaseWithRef(AIPhase.Idle);
     setError(null);
+    setErrorAtPhase(null);
+    setAgentResponse(null);
+    setPipelineName(null);
+    agentResponseRef.current = "";
+    navigationTriggeredRef.current = false;
+    pendingPipelineCodeRef.current = null;
+  }, [abort, setPhaseWithRef, clearNavigationTimer]);
```
yolanfery: [Nitpick] This is a bit much but ok why not (alternative is just to remount the component) 

But its repeated I think we can reduce duplication

yolanfery: It is repeated below also

yolanfery (changes requested):
Well done, that's a massive change and a complex feature ! 

I didn't get everything but this seems to make sense ☺️

Just few minor comments and one more important comment about the agent needing a dependency on the model

backend/hexa/assistant/agents/base.py
mrivar: There's a TODO and I believe a task as well to do this, but yeah I agree. I tried to play around with it a little bit in this PR but it proved a bit more complicated than I expected at first.
But I'll explore it a bit more, maybe it is more feasible than I think

Edit: pretty easy, going to push the solution later

backend/hexa/assistant/tests/agents/test_base_agent_run.py
mrivar: Good catch, working on it

backend/hexa/assistant/tests/test_views.py
mrivar: good one!, I'm seeing the code repeated as well in the  pipeline runs tests, working on it

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/useAIForm.ts
mrivar: refactored!

backend/hexa/assistant/agents/__init__.py
```diff
@@ -17,8 +18,12 @@
...
-def create_agent(conversation: Conversation) -> BaseAgent:
+def create_agent(
+    conversation: Conversation, builder: AiModelBuilder | None = None
+) -> BaseAgent:
     agent_class = _AGENT_REGISTRY.get(
         InstructionSet(conversation.instruction_set), BaseAgent
     )
-    return agent_class(conversation)
+    if builder is not None:
+        return agent_class(conversation, builder)
```
yolanfery: I propose to change to 

`return agent_class(conversation, agent)`

And the constructor to accept `None` agent

(Related to my comment below)

backend/hexa/assistant/agents/base.py
```diff
@@ -73,10 +74,14 @@ def _parse_conversation_title(text: str) -> str:
 class BaseAgent:
     instruction_set = InstructionSet.GENERAL
     tools: list = []
+    max_tokens: int = 4096
 
-    def __init__(self, conversation: Conversation):
+    @classmethod
+    def from_conversation(cls, conversation: Conversation) -> "BaseAgent":
+        return cls(conversation, AiModelBuilder.from_conversation(conversation))
```
yolanfery: Don't think we need this

backend/hexa/assistant/agents/base.py
```diff
@@ -73,10 +74,14 @@ def _parse_conversation_title(text: str) -> str:
...
+
+    def __init__(self, conversation: Conversation, builder: AiModelBuilder):
```
yolanfery: I propose to accept an agent instead of a builder and to do : 

`self._model = model or AiModelBuilder.from_conversation(conversation)`

backend/hexa/assistant/agents/base.py
```diff
@@ -154,6 +164,10 @@ async def run_stream(self, user_input: str) -> AsyncGenerator[str, None]:
                             yield sse
                 run_result = agent_run.result
 
+            if naming_task is not None:
+                precomputed_naming, sse = await self._resolve_naming_task(naming_task)
+                yield sse
```
yolanfery, line 171: Nice !

backend/hexa/assistant/tests/agents/_helpers.py
```diff
@@ -69,15 +69,12 @@ async def stream_func(messages, agent_info):
...
+def make_builder(test_model):
     mock_builder = MagicMock()
     mock_builder.model_api_name = "test"
     mock_builder.provider_id = "test"
     mock_builder.build.return_value = test_model
-    return patch(
-        "hexa.assistant.agents.base.AiModelBuilder.from_conversation",
-        return_value=mock_builder,
-    )
+    return mock_builder
```
yolanfery: Shouldn't be needed, and if we need, I much prefer mocking ai_settings

yolanfery (changes requested):

backend/hexa/assistant/tests/agents/_helpers.py
mrivar: I'm not sure I follow, why wouldn't we want to mock the builder?

Mocking the builder, we can also mock the Agent Model it builds and thus perform unit tests without depending on mocking/patching the pydantic api response

backend/hexa/assistant/agents/__init__.py
yolanfery: Sorry I meant a Model, not an agent

backend/hexa/assistant/tests/agents/_helpers.py
yolanfery: Yes, if I understand things right, your are mocking the builder here just to get a test_model, but you can supply the test_model directly instead with the changes of parameters I proposed in other comments

mrivar: Aaaah i get you now. The main issue is that we also need other info like the model provider and a model api name, which is not ensured to be given by the pydantic model abstraction. That's why I prefer to use the builder for this, because we know it already encapsulates all the info we need.
I guess we could have a constructor with every single parameter the agent needs and have a class function acting as constructor, like the `from_conversation` I had before. What do you think?

yolanfery: Hum, the version with the Model param as parameter is already better 

So it becomes [Minor] at this point

Not really fan of the Builder itself, values could get out-of-sync etc

What do you think of this ? We introduce a `BuiltModel` that is frozen and has everything needed, so that tests (or agents, like the naming agent...) can supply directly a built model ? 

```
 # model_builder.py
  from typing import NamedTuple

  from pydantic_ai.models import Model                                                                                                                                                                                                                                                                                   
  from pydantic_ai.models.anthropic import AnthropicModel
  from pydantic_ai.providers.anthropic import AnthropicProvider                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                         
  from hexa.assistant.exceptions import AssistantException
  from hexa.assistant.models import Conversation                                                                                                                                                                                                                                                                         
  from hexa.user_management.models import AiSettings
                                                                                                                                                                                                                                                                                                                         
   
  class BuiltModel(NamedTuple):                                                                                                                                                                                                                                                                                          
      model: Model           
      api_name: str
      provider_id: str
                                                                                                                                                                                                                                                                                                                         
   
  def _build_anthropic(model_api_name: str, api_key: str) -> Model:                                                                                                                                                                                                                                                      
      return AnthropicModel(model_api_name, provider=AnthropicProvider(api_key=api_key))
                                                                                                                                                                                                                                                                                                                         
   
  _PROVIDER_FACTORIES = {                                                                                                                                                                                                                                                                                                
      AiSettings.Provider.ANTHROPIC: _build_anthropic,
  }

                                                                                                                                                                                                                                                                                                                         
  def get_api_name(model: AiSettings.Model) -> str:
      model_to_api = {                                                                                                                                                                                                                                                                                                   
          AiSettings.Model.HAIKU: "claude-haiku-4-5-20251001",
          AiSettings.Model.OPUS: "claude-opus-4-6",                                                                                                                                                                                                                                                                      
          AiSettings.Model.SONNET: "claude-sonnet-4-6",                                                                                                                                                                                                                                                                  
      }                                                                                                                                                                                                                                                                                                                  
      model_api_name = model_to_api.get(model)                                                                                                                                                                                                                                                                           
      if model_api_name is None:                                                                                                                                                                                                                                                                                         
          raise AssistantException(
              f"Model {model} is not known. Accepted models are {[*model_to_api]}"                                                                                                                                                                                                                                       
          )                                                                                                                                                                                                                                                                                                              
      return model_api_name
                                                                                                                                                                                                                                                                                                                         
                             
  class AiModelBuilder:
      def __init__(self, provider: str, model: AiSettings.Model, api_key: str):
          self._provider = provider
          self._model_api_name = get_api_name(model)                                                                                                                                                                                                                                                                     
          self._api_key = api_key
                                                                                                                                                                                                                                                                                                                         
      @classmethod           
      def from_conversation(cls, conversation: Conversation) -> "AiModelBuilder":
          ai_settings: AiSettings = conversation.user.ai_settings_safe                                                                                                                                                                                                                                                   
          if not ai_settings.enabled:
              raise AssistantException("AI settings are not enabled")                                                                                                                                                                                                                                                    
          return cls(                                                                                                                                                                                                                                                                                                    
              provider=ai_settings.provider,
              model=ai_settings.model,                                                                                                                                                                                                                                                                                   
              api_key=ai_settings.api_key,
          )

      def build(self) -> BuiltModel:                                                                                                                                                                                                                                                                                     
          factory = _PROVIDER_FACTORIES.get(self._provider)
          if not factory:                                                                                                                                                                                                                                                                                                
              raise ValueError(f"Unsupported AI provider: {self._provider!r}")
          return BuiltModel(                                                                                                                                                                                                                                                                                             
              model=factory(self._model_api_name, self._api_key),
              api_name=self._model_api_name,                                                                                                                                                                                                                                                                             
              provider_id=self._provider,
          )                                                                                                                                                                                                                                                                                                              
   
  # base.py — constructor only, rest of the class unchanged                                                                                                                                                                                                                                                              
  def __init__(self, conversation: Conversation, built: BuiltModel | None = None):
      self.conversation = conversation                                                                                                                                                                                                                                                                                   
      self._model, self._model_api_name, self._provider_id = (
          built or AiModelBuilder.from_conversation(conversation).build()                                                                                                                                                                                                                                                
      )                                                                                                                                                                                                                                                                                                                  
   
      instructions = get_instructions(self.instruction_set)                                                                                                                                                                                                                                                              
      extra = self._extra_instructions()
      if extra:
          instructions += "\n\n" + extra
                                                                                                                                                                                                                                                                                                                         
      self.agent = Agent(
          model=self._model,                                                                                                                                                                                                                                                                                             
          instructions=instructions,
          tools=self._tools_with_context,
      )
```

yolanfery (approved):

backend/hexa/assistant/tests/agents/_helpers.py
mrivar: oh ok so as an inmutable object, I like it 😄, I'll change it and merge shortly after

yolanfery: Thanks and sorry for the confusing comments, I realize I could have been more explicit !

Well done anyway

PR #1752, opened 2026-04-22
fix: OH logo is stretched in the email (HEXA-1571)

DimitriKwihangana:
OH logo was stretched in the pipeline report email

## Changes

-  Setting only width lets the height scale proportionally

## How/what to test
- Configure smtp locally I've used [resend.com](resend.com)

- Trigger a pipeline run report email

## Screenshots / screencast

image


yolanfery (approved):
FYI we have mailhog configured in the project, so you shouldn't have to configure something else, an inbox should be available at http://localhost:8025/

PR #1753, opened 2026-04-22
feat: authorize webapps to access assets in the workspace (HEXA-1582)

yolanfery:
Allow webapps to get OH assets by calling a `/graphql/` endpoint on the subdomain

## Changes

- allow `/graphql/` endpoint on each webapp subdomain
- proxy graphql queries
- we just verify/filter that the queries are part of the allowed operation
- we attach the user doing the request and perform the GraphQL operation on behalf of it and return it
- allowed operations are configurable by webapp

## Screenshots / screencast

Screenshot 2026-04-22 at 14 12 22


examples/webapps/graphql-proxy-test.html
```diff
@@ -0,0 +1,146 @@
+<!DOCTYPE html>
```
yolanfery, line 1: Vibe coded example, no need for extensive review

backend/hexa/webapps/models.py
```diff
@@ -118,6 +118,13 @@ class WebappType(models.TextChoices):
         STATIC = "static", "Static"
         SUPERSET = "superset", "Superset"
 
+    class OperationScope(models.TextChoices):
+        PIPELINES_RUN = "PIPELINES_RUN", "Run pipelines"
+        PIPELINES_READ = "PIPELINES_READ", "Read pipelines"
+        FILES_READ = "FILES_READ", "Read files"
+        FILES_WRITE = "FILES_WRITE", "Write files"
+        USER_READ = "USER_READ", "Read user info"
```
bramj, line 128: No dataset read/write? (honest question, leaving it out for now is a fair call)

backend/hexa/webapps/middlewares.py
```diff
@@ -161,7 +163,17 @@ def _check_webapp_session(request, webapp):
 def _handle_webapp_request(request, webapp, *, request_has_user=True):
     """Shared auth + serve logic for both webapp middlewares."""
     if request.path.startswith("/graphql/"):
-        return HttpResponseNotFound("Not available")
+        if webapp.is_public:
+            return HttpResponseNotFound("Not available")
```
bramj, line 167: 👍

backend/hexa/webapps/graphql_proxy.py
```diff
@@ -0,0 +1,139 @@
...
+
+    event_properties = {
+        "webapp_id": str(webapp.id),
+        "webapp_name": webapp.name,
+        "workspace_id": str(webapp.workspace_id),
+        "operations": sorted(requested_fields),
+    }
+
+    if disallowed:
+        track(
```
bramj, line 134: Did you check this with @mcrimi ?

backend/hexa/webapps/graphql_proxy.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+
+    requested_fields -= INTROSPECTION_FIELDS
+    allowed_fields = {
+        f
+        for scope in webapp.allowed_operations
+        if scope in SCOPE_FIELDS
+        for f in SCOPE_FIELDS[scope]
+    }
+    disallowed = requested_fields - allowed_fields
```
bramj, line 124: I think the nature of GraphQL makes simple scoping like this insufficient down the road. At the least we should scope all queries on the `webapp.workspace`. Not sure how we'd do that, but right now a web app can for example manage pipelines that belong to other workspaces

Image

backend/hexa/webapps/graphql_proxy.py
```diff
@@ -0,0 +1,139 @@
...
+        "readFileContent",
+        "prepareObjectDownload",
+    },
+    Webapp.OperationScope.FILES_WRITE: {
+        "prepareObjectUpload",
+        "createBucketFolder",
+        "deleteBucketObject",
+        "writeFileContent",
+    },
+    Webapp.OperationScope.USER_READ: {"me", "workspace"},
```
bramj, line 51: With the `workspace` we should be careful as it allows extraction of connection credentials for example:

Image

bramj (approved):
Really cool and working well, this is a real game-changer for web apps!

I have a few important comments, but they are more pertinent in the context of public web apps. 
Given the current context of private-only, I think we can afford to go ahead with this, but it shows we need to really thread carefully if we want to make this into a public web apps feature.

backend/hexa/webapps/models.py
yolanfery: I just took the one asked by JP, but yes makes sense to add it, done

backend/hexa/webapps/graphql_proxy.py
yolanfery: Will ask if this is ok

backend/hexa/webapps/graphql_proxy.py
yolanfery: Yes, thats a shortcut I had to take to make it possible within the timeline constraints we got


I created a ticket to refine this https://bluesquare.atlassian.net/browse/HEXA-1619


I think we could create a "virtual" user with same data/scopes than the viewing user but bounded to the webapp's workspace, but we can of course brainstorm this...

backend/hexa/webapps/graphql_proxy.py
yolanfery: A bit less worried about this than the previous comment, if we scope the queries to the workspace that shouldn't be a problem



But yes this permission system is really basic and maybe not ideal for a graphql interface.. we might want take more time for the next steps

backend/hexa/webapps/graphql_proxy.py
bramj: Great :+1: I've moved the ticket to "web apps v3".

> I think we could create a "virtual" user with same data/scopes than the viewing user but bounded to the webapp's workspace, but we can of course brainstorm this...

Very clever! That sounds like a potentially great idea, to be explored.

backend/hexa/webapps/graphql_proxy.py
mcrimi: Looks reasonable. I'd only recommend adding `workspace_name` and a `is_public` property to the event, since joins in Mixpanel are very limited. Better to denormalize upfront.

yolanfery: Makes sense, will do, thanks

yolanfery: Done https://github.com/BLSQ/openhexa-app/pull/1765

PR #1761, opened 2026-04-23
fix:auto update pipeline run status on list page

DimitriKwihangana:
Active runs on the list page now poll for status updates using usePipelineRunPoller.

Also the PR fixes a race condition  that was introduced on the detail page where the poller could set isTerminal before the SSE done event arrived, causing run messages to disappear.

### DIFF
--- a/frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
+++ b/frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
@@ -7,12 +7,13 @@ import Time from "core/components/Time";
 import { createGetServerSideProps } from "core/helpers/page";
 import { formatDuration } from "core/helpers/time";
 import { NextPageWithLayout } from "core/helpers/types";
-import { PipelineParameter, PipelineRunTrigger, PipelineType } from "graphql/types";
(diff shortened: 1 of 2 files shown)


frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx
```diff
@@ -100,6 +106,9 @@ const WorkspacePipelineRunsPage: NextPageWithLayout = (props: Props) => {
 
   return (
     <Page title={pipeline.name ?? t("Pipeline runs")}>
+      {pipeline.runs.items.map((run) => (
+        <RunPoller key={run.id} run={run} />
+      ))}
```
Comment (yolanfery) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx:111]: I lost a bit track of this SSE topic, but did you try Appollo's `pollInterval` instead ?

Comment (DimitriKwihangana) [reply frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/runs/index.tsx:111]: ّI tried it first but it was causing messages to disappear on the detail page due to a race condition with SSE.

pollInterval was updating the shared Apollo cache and triggering isTerminal = true before the SSE done event could execute.

yolanfery (approved):

PR #1763, opened 2026-04-24
fix: Web app subdomain validation + logfire environment setting

bramj:
- fix: Validate the webapp subdomain on creation: the path of "create subdomain based on slug" didn't do the actual
validation. This means that if you create a new web app with name "Staging", you end up with subdomain "staging" which is a reserved field.
- chore: Add environment for logfire: for convenience, we'll reuse the Sentry variable, we'll always want to
keep those aligned anyways.


mrivar (approved):
Looks good to me 🙂

PR #1764, opened 2026-04-24
fix: preview of private static web apps (HEXA-1604)

yolanfery:
Depending on browsers, web apps previews don't always work for private static web apps

Instead of injecting the auth token through the cookies, build a preview url containing an expiring session key


bramj (approved):
That ended up more complicated than I expected :o 

Working perfectly for me locally. Let's test on demo! 👍

PR #1767, opened 2026-04-24
fix: group webapp icon name

DimitriKwihangana:
## Changes

- Group icon with name
- spacing between actions and name

## Screenshots / screencast
image


bramj (approved):
Nice! way better 👍

yolanfery (approved):

PR #1773, opened 2026-04-27
fix: psycopg2 imports

bramj:
Not 100% sure why, but some version upgrade of `psycopg2` made these imports fail.


yolanfery (approved):
Weird change but I like it better that way anyway

bramj:
@yolanfery agreed!

PR #1774, opened 2026-04-27
fix: folder cant be deleted (HEXA-1622)

yolanfery:
Deleting folders from the UI that were created without creating an empty marker file is failing because the empty marker file cant be found. Instead, list blobs and delete them

Integration to be tested in demo


bramj (approved):
Good find, and solid-looking fix 👍

PR #1775, opened 2026-04-29
chore: Make model limits specific and up the limit for pipeline creation

bramj:
A model hitting the limit will return a "cut-off" response, resulting in incorrect JSON for a tool call for example.

The default limit for Anthropic was 4096, but for the creation we can allow some more (set to double).

Example of requesting a complicated pipeline and hitting the 4096 limit, the resulting tool call is cut short:
image

**Note:** Follow-up ticket to improve this: https://bluesquare.atlassian.net/browse/HEXA-1635

### DIFF
--- a/backend/hexa/assistant/agents/base.py
+++ b/backend/hexa/assistant/agents/base.py
@@ -4,7 +4,7 @@ from collections.abc import AsyncGenerator
 from decimal import Decimal
 
 import genai_prices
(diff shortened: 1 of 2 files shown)


mrivar (approved):
Perfect 👌🏼
It's gonna be very useful to be able to have different token limits depending on the type of agent

PR #1778, opened 2026-04-30
fix: show a button to reconnect in case of pipeline logs SSE drop (HEXA-1632)

yolanfery:
## How/what to test

It is possible to test by running a pipeline and then restarting the backend


frontend/src/pipelines/hooks/usePipelineRunMessages/usePipelineRunMessages.ts
```diff
@@ -48,23 +56,25 @@ function usePipelineRunMessages(
     },
     timeout: () => {
       setStreamError("timeout");
-      Sentry.captureMessage("SSE pipeline run messages timed out", {
-        level: "warning",
-        extra: { runId },
-      });
```
bramj, line 54: 👍

bramj (approved):
Working well! 👍

PR #1779, opened 2026-04-30
fix: preview url shows outdated content (HEXA-1636)

yolanfery:
Instead of caching for 5 minutes, cache base on an ETag defined based on the published commit sha.


bramj (approved):
Wow! This is really cool, and I admit I didn't know about ETags, so I learned something, great!

Image

PR #1780, opened 2026-05-01
feature: pipeline run parameters - dynamic choices (HEXA-1620)

mrivar:
## Related PRs:
- https://github.com/BLSQ/openhexa-app/pull/1780
- https://github.com/BLSQ/openhexa-sdk-python/pull/386 (refactor)
- https://github.com/BLSQ/openhexa-sdk-python/pull/385

## Changes

  - **Dynamic choices from file**: Pipeline parameters can now declare `choices=ChoicesFromFile(path, column?)` in the SDK. The app resolves those choices at run time by reading a CSV, JSON, or YAML file from the workspace's file storage and returning a flat list of string values.
  - **New GraphQL query `pipelineParameterChoices(workspaceSlug, pipelineVersionId, parameterCode)`**: fetches the resolved choices on demand. The run-pipeline and version-config dialogs use it to populate a Select widget when a parameter has choicesFromFile set.
  - **Type mismatch warning**: When a choices file contains values that cannot be cast to the parameter's declared type (e.g. a string in an int parameter), those options are rendered **disabled in the dropdown** and a **warning** lists the invalid values.
  - **Config type validation from Backend**: Saving a pipeline version config or triggering a run now validates that each supplied value matches the parameter's declared type. Returns INVALID_CONFIG on mismatch instead of silently accepting bad data.
    -  _Note: It used to be only checked in the frontend and during the actual run._
    -  _Note 2: It is a simpler validation than the one done during the run, just checking types (such as int doesn't get a str value)._

## How/what to test

1. Create a pipeline that uses parameters with `choices=ChoicesFromFile(path, column?)` and `choices=path` (see the linked SDK PR for the API). Upload a matching CSV/JSON/YAML file to the workspace.
2. Open Run pipeline / Set default values — the parameter should show a dropdown populated from the file.
3. Upload a file whose values don't match the parameter type (e.g. strings for an int param) — invalid options should appear greyed out with a warning message below the select.
4. In the version config dialog, set a parameter to a value of the wrong type and save — expect an error.
5. Test error paths: delete the choices file, use a malformed file, or point to a non-existent column — the widget should show an inline error rather than crashing. 

## Screenshots / screencast

Example with CSV with 2 columns:
- First param has `choices=ChoicesFromFile("choices/choices_2_cols.csv")`. Not column chosen so it gives an error.
- Second param has  `choices=ChoicesFromFile("choices/choices_2_cols.csv", "two")` so it shows the option of column two.

```csv
name,two
asd,two1
qwe,two2
zxc,two3
csv,two4
```

Screenshot from 2026-05-05 16-32-28

Warning when a pipeline parameter has invalid choices (e.g.: int/float with a choices file with strings) the invalid options appear disabled:
image

Error when trying to save an invalid pipeline parameter (e.g.: an int/float pipeline parameter tries to save a string value):
Screenshot from 2026-05-06 16-58-53

Error when everything goes wrong (in theory this state should be unreachable, but it's a fail safe in case everything goes wrong 🤪)
image


.gitignore
```diff
@@ -26,6 +26,7 @@ static/uploads/*
 
 docker-compose.debug.yaml
 docker-compose.dev.yaml
+docker-compose.override.yaml
```
mrivar, line 29: Using this file to run my local `openhexa-sdk-python` and test the changes in it:

`docker-compose.override.yaml`:
```dockerfile filename="docker-compose.override.yaml"
services:
  app:
    volumes:
      - ../openhexa-sdk-python:/openhexa-sdk-python
    entrypoint: ["/bin/bash", "-c", "pip install --user -q -e /openhexa-sdk-python && exec /code/docker-entrypoint.sh \"$@\"", "--"]
```

backend/hexa/pipelines/graphql/schema.graphql
```diff
@@ -90,6 +90,24 @@ enum ParameterWidget {
   IASO_FORMS
 }
 
+"""
+Describes a dynamic choices source backed by a workspace file.
+"""
+type PipelineParameterChoicesFromFile {
+  path: String!    # Path to the file in the workspace file system.
+  format: String!  # File format: "csv", "json", or "yaml".
```
yolanfery: Not an Enum instead ?

yolanfery: Can't we infer the format from the file content/extension ?

backend/hexa/pipelines/graphql/schema.graphql
```diff
@@ -756,6 +777,12 @@ extend type Query {
   Retrieves a pipeline run by ID.
   """
   pipelineRun(id: UUID!): PipelineRun @loginRequired
+
+  """
+  Resolves the list of choices for a parameter backed by a workspace file.
+  Returns a list of string values read from the file at the time of the call.
+  """
+  pipelineParameterChoices(workspaceSlug: String!, pipelineVersionId: UUID!, parameterCode: String!): [String!] @loginRequired
```
yolanfery: Why do we need the workspace slug ? The pipeline version id should be enough

backend/hexa/pipelines/schema/queries.py
```diff
@@ -172,6 +173,44 @@ def resolve_pipeline_version(_, info, **kwargs):
...
+        version = PipelineVersion.objects.select_related("pipeline__workspace").get(
+            id=pipeline_version_id, pipeline__workspace__slug=workspace_slug
+        )
+    except PipelineVersion.DoesNotExist:
+        raise ValueError(f"Pipeline version '{pipeline_version_id}' not found.")
+
+    if not request.user.has_perm("pipelines.view_pipeline_version", version):
+        raise PermissionError(
+            "You do not have permission to view this pipeline version."
+        )
```
yolanfery: Can we use a pattern like `filter_for_user` to keep consistency in the codebase ?

backend/hexa/pipelines/models.py
```diff
@@ -39,6 +39,27 @@
...
+    "bool": lambda v: isinstance(v, bool),
+    "int": lambda v: isinstance(v, int) and not isinstance(v, bool),
+    "float": lambda v: isinstance(v, (int, float)) and not isinstance(v, bool),
+    "str": lambda v: isinstance(v, str),
+    "secret": lambda v: isinstance(v, str),
+    "dataset": lambda v: isinstance(v, str),
+    "file": lambda v: isinstance(v, str),
+}
+
+_CONNECTION_TYPES = frozenset({"dhis2", "postgresql", "iaso", "s3", "gcs", "custom"})
```
yolanfery: Can we take this list from an existing place instead fo duplicating ?

backend/hexa/pipelines/choices_from_file.py
```diff
@@ -0,0 +1,180 @@
...
+    ------
+        ValueError: If the file is missing, too large, unparseable, or the
+            column/key cannot be resolved.
+    """
+    path = choices_from_file["path"].lstrip("/")
+    fmt = choices_from_file["format"]
+    column = choices_from_file.get("column")
+
+    try:
+        obj = storage.get_bucket_object(bucket_name, path)
```
yolanfery, line 53: We need safeguard to prevent path traversal, there are examples in the code base (`is_safe_path`)

backend/hexa/pipelines/tests/test_schema/test_pipeline_parameter_choices.py
```diff
@@ -0,0 +1,549 @@
...
+            pipeline=self.PIPELINE,
+            user=self.USER,
+            parameters=parameters,
+            version_number=PipelineVersion.objects.filter(
+                pipeline=self.PIPELINE
+            ).count()
+            + 1,
+        )
+
+    def _run(self, version, code):
```
yolanfery: We might want a more descriptive naming here

backend/hexa/pipelines/tests/test_schema/test_pipeline_parameter_choices.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+def _make_storage_object(size=100):
+    obj = MagicMock()
+    obj.size = size
+    obj.type = "file"
+    return obj
+
+
+class PipelineParameterChoicesTest(GraphQLTestCase):
```
yolanfery, line 44: Seems out of place to test all those edge cases in a GraphQL test instead of unit tests on the actual methods doing the work. But ok, I guess this will survive refactoring

yolanfery: I see now that we have unit tests also, there might be overlaps but ok, this is fine, just surprised 😁

frontend/src/workspaces/features/RunPipelineDialog/FileParameterField.tsx
```diff
@@ -14,7 +14,7 @@ type FileParameterFieldProps = {
   parameter: {
     code: string;
     required?: boolean;
-    directory?: string;
+    directory?: string | null;
```
yolanfery, line 17: [Question] Can you please explain this change and the one below in the same file ? 

We already have `?` on `directory`, I don't get why we need to add this union type

yolanfery (changes requested):
Solid ! Good to see lot of tests also ☺️

No major concern, just some code quality topics and questions.

Also, we might want to add documentation as part of this PR now that the documentation is in the repo.

backend/hexa/pipelines/graphql/schema.graphql
mrivar: yesss, we were doing it from the SDK, but we can make the SDK just be "dumb" and just resend the information, and then we calculate it here

backend/hexa/pipelines/graphql/schema.graphql
mrivar: True, not really needed. At first I thought of filtering by workspace, but I don't think that's needed at all

frontend/src/workspaces/features/RunPipelineDialog/FileParameterField.tsx
mrivar: It's introduced because otherwise we get an error when running codegen:
https://github.com/BLSQ/openhexa-app/actions/runs/25879781042/job/76056394291?pr=1780
It's just because null and undefined are not considered the same apparently:
```
  Types of property 'directory' are incompatible.
    Type 'string | null | undefined' is not assignable to type 'string | undefined'.
      Type 'null' is not assignable to type 'string | undefined'. 
```

yolanfery (changes requested):
Great, thanks for the changes, just missing some documentation 🙂

mrivar:
> Great, thanks for the changes, just missing some documentation 🙂

Right! I forgot about that, thanks for reminding me

yolanfery (approved):

PR #1784, opened 2026-05-05
feat:pipeline conversation history on code screen(HEXA-1624)

DimitriKwihangana:
The PR focuses on making the Assistant remember past conversations and shows them in history list. and It enables to start new conversations.Also the conversation starts from creating the pipeline
## Changes

- The conversations are saved with a link pointing to pipeline after creating pipeline
- Added conversation history and linked to pipeline 
-  Auto open chat panel if a pipeline already had conversation with AI

## Screenshots / screencast

image


frontend/src/assistant/features/PipelineEditChatPanel/index.tsx
```diff
@@ -39,10 +73,27 @@ export default function PipelineEditChatPanel({
...
+    const newConv: PipelineConversation = {
+      id: conversation.id,
+      name: null,
+      createdAt: conversation.createdAt,
+    };
+    onConversationCreated(newConv);
+    return conversation.id;
+  }, [createConversation, workspaceSlug, pipelineId, onConversationCreated]);
+
+  const handleConversationNameChange = useCallback(
```
bramj, line 85: When I click the "new conversation" button, the header of my previous conversation is preserved until I send my first message.

This gives the perception of a "history clear" rather than a new conversation. Clicking new should clear the header.

https://github.com/user-attachments/assets/42b88f0f-3950-4c9c-a31c-85d3fc8bd333

bramj: Additionally, when switching to a previous conversation, let's remove the "typewriter" effect. It's great for the perception of an agent typing, but doesn't make much sense when opening an older conversation.

frontend/src/assistant/features/PipelineEditChatPanel/index.tsx
```diff
@@ -80,20 +131,80 @@ export default function PipelineEditChatPanel({
...
+                setShowHistory(false);
+              }}
+              className={clsx(
+                "w-full px-4 py-3 text-left text-sm border-b border-gray-100 transition-colors hover:bg-gray-50",
+                activeConversationId === conv.id
+                  ? "bg-blue-50 text-blue-700 font-medium"
+                  : "text-gray-700",
+              )}
+            >
+              {conv.name || formatDate(conv.createdAt)}
```
bramj: How about always showing the creation date from the conversation (or even better: the `updatedAt` or creation of the last message)? We can align it to the right here:

Image

backend/hexa/assistant/agents/create_pipeline_agent.py
```diff
@@ -1,10 +1,61 @@
...
+            tool_invocations,
+            usage,
+            all_messages,
+            is_first_message,
+            precomputed_naming=precomputed_naming,
+        )
+        await self._link_created_pipeline(tool_invocations)
+        return result
+
+    async def _link_created_pipeline(self, tool_invocations):
```
bramj: I feel like the complexity of this function indicates that it's maybe not the right place to add this "link the conversation to a model" feature.

Did you explore alternative routes to get there? Like for example adding something like this to the `BaseAgent`:
```python
async def _on_tool_result(self, invocation: ToolInvocation) -> None:
        pass
```

And override it in the `CreatePipelineAgent`. That way we avoid the looping over the tool invocations to get to the `pipeline_id`.

bramj (changes requested):
I think this UX looks nice is going in the right direction. 👍 

I see a few things missing, noted in the comments.

And on a general note:
When I create a pipeline with AI, I now correctly see the conversation when arriving on the edit page, that's great!
However: the linked agent should now become the edit one, or we end up being blocked on continuing the conversation:

Image

So here the agent says he can't edit, and ends up making a new pipeline with the edits :)

frontend/src/assistant/features/PipelineEditChatPanel/index.tsx
mcrimi: I think that's a good idea, but I'd be very subtle with the UI.

backend/hexa/assistant/schema/queries.py
```diff
@@ -26,6 +28,14 @@ def resolve_workspace_assistant_conversations(workspace: Workspace, info, **kwar
     )
 
 
+@pipeline_object.field("assistantConversations")
+def resolve_pipeline_assistant_conversations(pipeline: Pipeline, info, **kwargs):
```
mcrimi, line 33: The conversations need to be scoped by **user**  also. I'm able to see other user's conversation when I go to the pipeline.

cc: @bramj

mcrimi (changes requested):

frontend/src/assistant/features/PipelineEditChatPanel/index.tsx
mcrimi: I also realized that we need a way for the user to (soft) delete the convos, but that can be another ticket.

backend/hexa/assistant/schema/queries.py
bramj: Woops, good catch! That's definitely a requirement

backend/hexa/assistant/agents/create_pipeline_agent.py
```diff
@@ -1,10 +1,43 @@
...
+            self.conversation.linked_object_content_type = ct
+            self.conversation.linked_object_id = pipeline.id
+            self.conversation.instruction_set = InstructionSet.EDIT_PIPELINE
+            await self.conversation.asave(
+                update_fields=[
+                    "linked_object_content_type",
+                    "linked_object_id",
+                    "instruction_set",
+                ]
+            )
```
bramj: [Nitpick] Is there a reason why you specify the specific update fields? I feel it can sometimes just add more lines of code for little benefit, `await self.conversation.asave()` is easier to read.

backend/hexa/assistant/agents/create_pipeline_agent.py
```diff
@@ -1,10 +1,43 @@
...
+        except Exception:
+            logger.exception(
+                "CreatePipelineAgent: failed to link conversation %s to pipeline %s",
+                self.conversation.id,
+                pipeline_id,
+            )
```
bramj: [Nitpick] IMO there's no need to wrap this in a `try:`. If this fails it's a genuine bug that we should fix. (It does notify us, that's good, but I just think it's unnecessary).

backend/hexa/assistant/agents/base.py
```diff
@@ -219,9 +219,11 @@ async def _stream_model_node(node, ctx):
                         "text_delta", TextDeltaPayload(delta=event.delta.content_delta)
                     )
 
-    @staticmethod
+    async def _on_tool_result(self, invocation: ToolInvocation) -> None:
```
bramj, line 222: [Nitpick] I know we don't this in other locations in this file, but for a method like this I would add a little docstring that explains the reasoning behind this method, how it's a hook that allows you to add side-effects to a tool execution.

frontend/src/assistant/features/PipelineEditChatPanel/index.tsx
```diff
@@ -187,7 +198,14 @@ export default function PipelineEditChatPanel({
                   : "text-gray-700",
               )}
             >
-              {conv.name || formatDate(conv.createdAt)}
+              <div className="flex items-center justify-between gap-2">
+                <span className="truncate">
+                  {conv.name || t("New conversation")}
+                </span>
+                <span className="shrink-0 text-xs text-gray-400">
+                  {formatDate(conv.updatedAt)}
```
bramj: This doesn't seem to be localized correctly, in FR:

Image

I see the `formatDate()` actually comes from the hidden test-frontend (entirely vibe-coded). We'll remove that soon to avoid using it as an example.

Best to look for other example usages of `toLocaleString` in the repo.

bramj (changes requested):
Much better @DimitriKwihangana , good stuff! All mayor things looking good now, I have only a few small comments/nitpicks.

Can you rebase your branch on `main`?

backend/hexa/assistant/agents/create_pipeline_agent.py
DimitriKwihangana: Nice Idea! there was no specific reason.only wanted to update those particular fields.

backend/hexa/assistant/agents/base.py
DimitriKwihangana: Absolutely!

bramj (approved):
Great @DimitriKwihangana ! 👍

PR #1785, opened 2026-05-05
fix: Read-only user should not be able to create tables (HEXA-1630)

bramj:
### Main changes

- Explicitly revoke the `CREATE` privilige for read-only roles (with a unit test that reproduces the issue we had on prod and shows how it's fixed)
- 2 migrations:
   1. Remove the `CREATE` privilige from the public role (this effectively removes it from RO role, RW is unaffected because it sets it explicitly)
   2. Change ownership of any tables that were created by RO roles.

### How to test

Difficult, since this issue only arises on production (not even on `demo`). So I propose a code-only review and good testing/verification on production.
Wrt to the test: I did make sure the test fails when the fix is absent.

### DIFF
--- a/backend/hexa/databases/tests/test_api.py
+++ b/backend/hexa/databases/tests/test_api.py
@@ -1,4 +1,5 @@
 import psycopg2
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from psycopg2 import OperationalError, sql
 from psycopg2.errors import (
@@ -9,6 +10,8 @@ from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT, STATUS_READY
 from hexa.core.test import TestCase
 from hexa.databases.api import (
     create_database,
+    create_read_and_write_role,
+    create_read_only_role,
     delete_database,
     get_database_connection,
     get_db_server_credentials,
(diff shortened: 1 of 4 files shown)


backend/hexa/databases/tests/test_api.py
```diff
@@ -362,3 +365,58 @@ def test_read_only_role_cannot_access_other_database(self):
...
+        finally:
+            admin_conn.close()
+
+        try:
+            db_conn = get_database_connection(db_name)
+            db_conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+            try:
+                with db_conn.cursor() as cursor:
+                    cursor.execute("GRANT CREATE ON SCHEMA public TO PUBLIC;")
+                    create_read_and_write_role(db_name, pwd, cursor)
```
yolanfery, line 403: This creation might not be used in the test

Or actually better, we can check that this suer can actually create table

backend/hexa/workspaces/migrations/0058_revoke_create_on_public_from_public.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Workspace = apps.get_model("workspaces", "Workspace")
+
+    for workspace in Workspace.objects.all():
+        db_name = workspace.db_name
+        try:
+            conn = get_admin_connection(db_name)
+            try:
+                with conn.cursor() as cur:
+                    cur.execute("REVOKE CREATE ON SCHEMA public FROM PUBLIC;")
```
yolanfery, line 47: To be sure I get this : 

So it means no more implicit CREATE permission for all users
So I guess we explicitly grant this permission to RW user ?

yolanfery: Seems confirmed by https://github.com/BLSQ/openhexa-app/blob/6a8b11d2fca1bcabea1119545e52305a2bc8e5a3/backend/hexa/databases/api.py#L95-L109

backend/hexa/workspaces/migrations/0059_reassign_objects_owned_by_ro_role.py
```diff
@@ -0,0 +1,81 @@
...
+                    # REASSIGN OWNED BY requires the executor to be a member of both the source
+                    # and target roles.
+                    cur.execute(
+                        sql.SQL("GRANT {ro_role} TO {admin_role}").format(
+                            ro_role=sql.Identifier(ro_role),
+                            admin_role=sql.Identifier(admin_role),
+                        )
+                    )
+                    cur.execute(
+                        sql.SQL("REASSIGN OWNED BY {ro_role} TO {rw_role}").format(
```
yolanfery, line 60: Handy instruction 👌 perfect for our use case 😂

backend/hexa/workspaces/migrations/0059_reassign_objects_owned_by_ro_role.py
```diff
@@ -0,0 +1,81 @@
...
+            try:
+                with conn.cursor() as cur:
+                    # REASSIGN OWNED BY requires the executor to be a member of both the source
+                    # and target roles.
+                    cur.execute(
+                        sql.SQL("GRANT {ro_role} TO {admin_role}").format(
+                            ro_role=sql.Identifier(ro_role),
+                            admin_role=sql.Identifier(admin_role),
+                        )
+                    )
```
yolanfery, line 58: Just wondering if we shouldn't clean up this after the REASSIGN instruction to avoid future confusions,...

yolanfery (approved):
Tricky one
Well done and thanks for the documentation/learning 

Just a few comments/questions but not blocking

backend/hexa/workspaces/migrations/0058_revoke_create_on_public_from_public.py
bramj: Exactly! I actually ran this exact command already on the `template1` database. So now new workspace DBs (that are created as copies of `template1`) don't have implicit CREATE permission anymore.

The migration is to do the same fix on existing DBs.

backend/hexa/workspaces/migrations/0059_reassign_objects_owned_by_ro_role.py
bramj: Yes totally!

backend/hexa/workspaces/migrations/0059_reassign_objects_owned_by_ro_role.py
bramj: True, it's not really a problem, but I guess it can cause some confusion. I'll add a revoke.

backend/hexa/databases/tests/test_api.py
bramj: Good catch, added this improvement 👍

PR #1789, opened 2026-05-07
feat: add documentation on how to use the GraphQL API in web apps (HEXA-1642)

yolanfery:
Webapps can call the GraphQL layer. Users and agents are not aware of this, so : 

- Add documentation (and expose it to MCP tools) with 7 full examples (tested and validated)
- Explain the webapps permissions in the tools docstrings and allow to specify them on create and update
- Inject the following context into webapps to ease developpement
```
context = {
        "workspaceSlug": webapp.workspace.slug,
        "webappSlug": webapp.slug,
        "isPublic": webapp.is_public,
    }
```
so that it can be used as 
`const {workspaceSlug} = window.OPENHEXA`

## Screenshots

Screenshot 2026-05-07 at 15 13 18
Screenshot 2026-05-07 at 15 13 40


backend/hexa/mcp/tools/webapps.py
```diff
@@ -61,8 +81,25 @@ def update_static_webapp(
...
+    - files_json: JSON array of {path, content} objects to replace all files.
+    - name: new human-readable name.
+    - description: new description.
+    - allowed_operations: comma-separated list of API scopes the webapp's JS may call via the
+      same-origin /graphql/ proxy. Pass an empty string to leave the current scopes untouched, or
+      "NONE" to revoke all access. Valid scopes: PIPELINES_READ, PIPELINES_RUN, FILES_READ,
+      FILES_WRITE, DATASETS_READ, DATASETS_WRITE, USER_READ. Example: "PIPELINES_READ,FILES_READ".
+      For what each scope grants and example queries the webapp's JS can run, call
+      get_help_or_doc(topic="static-webapps").
+    """
```
bramj, line 91: [Nitpick] I wonder if we should try to settle on a "standardized" way for these doc strings. Here we have the "Parameters" with a bullet list, for the `create` action it's a regular text block.

backend/hexa/webapps/middlewares.py
```diff
@@ -96,9 +97,34 @@ def _inject_powered_by_banner(response):
...
+
+    content = response.content.decode(response.charset)
+    head_close = content.find("</head>")
+    if head_close == -1:
+        # No <head> — likely a fragment or non-page asset served as text/html. Skip injection.
+        return response
+    content = content[:head_close] + snippet + content[head_close:]
+    response.content = content.encode(response.charset)
+    response["Content-Length"] = len(response.content)
+    return response
```
bramj, line 121: Very clever, nice! 👍 And I see it's in the documentation as well, great!

bramj (approved):
Very nice! I'm already digging having the documentation live inside the main repo 😄

backend/hexa/mcp/tools/webapps.py
yolanfery: You are right, thanks, we should stick to one. Others are text blocks, I changed it to keep that consistent

We can move later to more structure if it becomes difficult to maintain

PR #1790, opened 2026-05-08
fix: Bring back Django auto reload for local development

bramj:
Seems like it got lost somewhere when we changed the `gunicorn` settings.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -31,6 +31,7 @@ DJANGO_SETTINGS_MODULE=config.settings.base
 # Django debugging settings
 DEBUG_LOGGING=true
 DEBUG_TOOLBAR=false
+DJANGO_HOT_RELOAD=false
 
 # Self-registration (allow users to sign up without an invitation)
 ALLOW_SELF_REGISTRATION=false
--- a/backend/docker-entrypoint.sh
+++ b/backend/docker-entrypoint.sh
@@ -36,7 +36,7 @@ case "$command" in
   ;;
 "start")
   wait-for-it ${DATABASE_HOST:-db}:${DATABASE_PORT:-5432}
-  gunicorn config.asgi:application -k config.workers.UvicornWorkerNoLifespan --bind 0:8000 --workers=3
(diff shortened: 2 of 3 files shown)


yolanfery (approved):
Nothing against this change but it seems its not needed on my side as I still have hot reload 

```
app-1                              | /opt/project/backend/hexa/assistant/instructions.py changed, reloading.

```

mrivar:
ok so what I found is:
- The hot reload broke when we switched from WSGI Django's dev server (with hot reload built-in) to gunicorn with ASGI workers (needed for SSE).
- The problem -> gunicorn + async workers hot reload doesn't work -> it detects file changes but the worker restart fails silently
- The solution:
  - ❌ ~gunicorn + sync workers~ -> doesn't really work as it would break SSE
  - ✅ uvicorn with reload for dev

I set a new env variable `DJANGO_HOT_RELOAD`, false by default. If it is true, we run the uvicorn server (only for development). If it's false or unset, it runs the proper gunicorn+async workers setup, proper for production.
So, in order for it to work you need to set DJANGO_HOT_RELOAD=true in your personal `.env`

yolanfery:
Not sure about the approach now

But @mrivar found why I don't have issues : I am using a `docker-compose.debug.yaml` (https://github.com/BLSQ/openhexa-app/blob/413169e8929d9f24f8878de2828319c7f0689608/README.md?plain=1#L241)

```
services:
  app:
    entrypoint: []
    # Used when running in normal mode.
    command: ["/code/docker-entrypoint.sh", "manage", "runserver", "0.0.0.0:8000"]
    ports:
      - "9000:8000"
    environment:
      IS_LOCAL_DEV: false
```

yolanfery:
I liked to approach of @bramj a bit better (the command is the same with additional arg for local dev) but no strong opinion 

But it needs to work 😂

bramj:
@mrivar @yolanfery Damn you're right Martin, for me it works right after startup, but then it just stops working... I'm trusting your commit more :sweat_smile: thanks for the fix! :+1:

PR #1791, opened 2026-05-08
fix: Web apps image files are corrupted upon upload

bramj:
Webapp file content was sent over GraphQL as UTF-8 strings, which silently corrupted binary files (images, fonts, etc.). Switch the encoding to base64 end-to-end so binary assets round-trip cleanly between the editor and Forgejo.
 
## Changes

- Backend now returns webapp file content as base64 over GraphQL.
- Frontend encodes uploads/edits to base64 before sending, and decodes for display in the editor.
- Only compute `line_count` for recognized text files

## How to test

- Upload images into a webapp and verify they render correctly when the webapp runs.
- Edit and save a text file in the editor, check if that's still working.

**Note:** Images uploaded under the old code are corrupted in storage and cannot be recovered, users will need to re-upload.

## Screenshot

After re-upload:
image

### DIFF
--- a/backend/hexa/git/forgejo.py
+++ b/backend/hexa/git/forgejo.py
@@ -148,10 +148,8 @@ class ForgejoClient(GitClient):
             path = file["path"]
             is_update = path in existing_tree
             content = file["content"]
-            if isinstance(content, str):
-                content = base64.b64encode(content.encode()).decode()
-            elif isinstance(content, bytes):
-                content = base64.b64encode(content).decode()
+            if isinstance(content, bytes):
+                content = base64.b64encode(content).decode("ascii")
             op = {
                 "operation": "update" if is_update else "create",
                 "path": path,
@@ -184,7 +182,7 @@ class ForgejoClient(GitClient):
         *,
         org_slug: str | None = None,
     ) -> list[dict]:
-        """Fetch the full file tree and return a flat list with path, type, and content."""
+        """Fetch the full file tree and return a flat list with path, type, and base64 content."""
         tree = self.get_files_tree(repo_name, ref, org_slug=org_slug)
         nodes: list[dict] = []
 
@@ -196,7 +194,7 @@ class ForgejoClient(GitClient):
                 nodes.append({"path": path, "type": "directory", "content": None})
             elif entry_type == "blob":
                 raw = self.get_file(repo_name, path, ref, org_slug=org_slug)
-                content = raw.decode("utf-8", errors="replace")
+                content = base64.b64encode(raw).decode("ascii")
                 nodes.append({"path": path, "type": "file", "content": content})
 
         return nodes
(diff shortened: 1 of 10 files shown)


frontend/src/webapps/helpers/base64.ts
```diff
@@ -0,0 +1,37 @@
...
+      }
+      const comma = result.indexOf(",");
+      if (comma < 0) {
+        reject(new Error("Malformed data URL"));
+        return;
+      }
+      resolve(result.slice(comma + 1));
+    };
+    reader.readAsDataURL(file);
+  });
```
yolanfery: [Minor] Quite surprised that there is no simpler/built-in/standard way for this, no ?

yolanfery (approved):
Looking good, nice improvement

frontend/src/webapps/helpers/base64.ts
bramj: Not really apparently, but there were some unnecessary checks that I just removed.

Also took the opportunity to refactor a bit:
- extracted a `base64ToBytes` helper that's reused for pipelines/template download
- move the helper to `core` since it's not just related to web apps

yolanfery: Great, thanks !

frontend/src/core/helpers/base64.ts
```diff
@@ -0,0 +1,25 @@
...
+export const fileToBase64 = (file: Blob): Promise<string> =>
+  new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onerror = () => reject(reader.error);
+    reader.onload = () => {
+      const result = reader.result as string;
+      resolve(result.slice(result.indexOf(",") + 1));
+    };
+    reader.readAsDataURL(file);
+  });
```
yolanfery, line 25: Great idea to define helpers
[Minor] Might want to add some quick unit tests to cover those
Like :
`base64ToString(stringToBase64(XXXXXXXXXXX)) = XXXXXXXXXXX`

backend/hexa/git/forgejo.py
```diff
@@ -196,7 +194,7 @@ def get_repository_files(
                 nodes.append({"path": path, "type": "directory", "content": None})
             elif entry_type == "blob":
                 raw = self.get_file(repo_name, path, ref, org_slug=org_slug)
-                content = raw.decode("utf-8", errors="replace")
+                content = base64.b64encode(raw).decode("ascii")
```
yolanfery, line 197: Actually... (and sorry for realising this now) is this a good idea for LLMs ? Reading webapps content now requires decoding which is not great at all

We need to discuss this because this seems like a blocker to me

yolanfery (changes requested):

bramj:
Closing this, rewrite here: https://github.com/BLSQ/openhexa-app/pull/1815

PR #1797, opened 2026-05-11
fix: AI agent streaming URL fallback

yolanfery:
It should fallback on empty string. Not using express proxy because of SSE


mrivar (approved):
good one, I didn't catch it!

PR #1801, opened 2026-05-12
fix: Correctly set web apps domain in ALLOWED_HOSTS + CORS

bramj:
Noticed while testing the local hosting image:

If the domain contained a port, such as for local dev, the settings didn't really work. We just didn't notice because on our local docker container because the env is set with `ADDITIONAL_ALLOWED_HOSTS=*`, which gives a freepass.

This small fix strips the port if any and correctly sets `ALLOWED_HOSTS` and `CORS_ALLOWED_ORIGIN_REGEXES` (with a small regex escape just in case).

### How to test

Remove the `ADDITIONAL_ALLOWED_HOSTS=*` setting from `docker-compose.yaml`. Without the fix from this PR, you'll get this when loading a web app:

Screenshot 2026-05-13 at 13 31 28

### DIFF
--- a/backend/config/settings/base.py
+++ b/backend/config/settings/base.py
@@ -82,11 +82,14 @@ ALLOWED_HOSTS = (
     else []
 )
 
-# Example: "webapps.openhexa.org"
+# Example: "webapps.openhexa.org" (or "webapps.localhost:8000" for local dev).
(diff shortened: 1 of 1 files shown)


yolanfery (approved):
Good catch 👌

PR #1804, opened 2026-05-12
feat: webb app diffs and browse at this point button

DimitriKwihangana:
Adds the ability to inspect what changed in each commit of a web app's file history. Clicking a commit in the history list opens a dedicated diff page that shows the commit metadata alongside a per-file unified diff, mirroring the GitHub commit view.

## Changes

-  Added  get_commit in the Forgejo Client to fetch a single commit by SHA, and get_commit_diff() on GitWebapp
-  Added  a button that  links to the file browser at that point in history.
- Clicking a commit message opens a new commitIdpage with a CommitDiff component showing per file unified diffs 

## Screenshots / screencast
image


backend/hexa/webapps/models.py
```diff
@@ -283,6 +283,81 @@ def get_files(self, ref="main"):
             )
         return nodes
 
+    def get_commit_diff(self, sha: str) -> dict:
+        import difflib
```
yolanfery: Any reason to import this dynamically ?

frontend/public/locales/fr/messages.json
```diff
@@ -1003,5 +1003,10 @@
   "Your account": "Votre compte",
```
yolanfery, line 1019: No `en/messages.json` changes ?

backend/hexa/webapps/models.py
```diff
@@ -283,6 +283,81 @@ def get_files(self, ref="main"):
             )
         return nodes
 
+    def get_commit_diff(self, sha: str) -> dict:
```
yolanfery, line 290: Can we do more unit tests for this function in particular ? It's a tricky one

backend/hexa/webapps/models.py
```diff
@@ -283,6 +283,81 @@ def get_files(self, ref="main"):
...
+        parent_files = (
+            {
+                f["path"]: f["content"] or ""
+                for f in self.get_files(parent_sha)
+                if f["type"] == "file"
+            }
+            if parent_sha
+            else {}
+        )
+
```
yolanfery: I think it should be possible to get a diff from Forgejo directly 

https://git.demo.openhexa.org/api/swagger#/repository/repoDownloadCommitDiffOrPatch

This would avoid maintenance for us.

If not (but I would be surprised), we should at least try to separate the diff logic from the commit fetching logic. This improves testing as well

frontend/src/webapps/features/CommitDiff/CommitDiff.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const webapp = data?.webapp;
+  const diff = webapp?.commitDiff;
+  const files = diff?.files ?? [];
+  const source = webapp?.source;
+  const publishedVersionId =
+    source?.__typename === "GitSource" ? source.publishedVersion : null;
+  const isPublished = publishedVersionId === commitId;
+
+  const totalAdditions = files.reduce((sum, f) => sum + f.additions, 0);
+  const totalDeletions = files.reduce((sum, f) => sum + f.deletions, 0);
```
yolanfery: This kind of data should come from the backend in my opinion

yolanfery (changes requested):
This is just too cool, well done 😎🚀

A bit of rework is required around the `get_commit_diff` method but except that, looking great

backend/hexa/webapps/models.py
DimitriKwihangana: No specific reason

frontend/public/locales/en/messages.json
```diff
@@ -1,31 +1,23 @@
 {
```
yolanfery, line 1: Why is the diff in this file quite big ? Isn't it possible to just add required changes for this PR ? Without moving stuff around ? Genuinely curious

yolanfery (changes requested):
I think we need to step back a bit here 

The result from `/repos/{org_slug}/{repo_name}/git/commits/{sha}.diff` already look better.

But I am wondering if we shouldn't just pass this result to the frontend and render it. Instead of parsing the Forgejo response, creating a structured object and then deconstructing it in the frontend. 

What do you think ?

backend/hexa/webapps/tests/test_models.py
```diff
@@ -1008,3 +1008,127 @@ def test_filter_for_user(self):
 
         webapps = Webapp.objects.filter_for_user(self.user_viewer)
         self.assertIn(git_webapp.webapp_ptr, webapps)
+
+
+class ParseGitDiffTest(TestCase):
```
yolanfery: `_parse_git_diff` does not exist anymore, we can remove this test suite

frontend/src/webapps/features/CommitDiff/CommitDiff.tsx
```diff
@@ -0,0 +1,86 @@
...
+              className="shrink-0 rounded border border-gray-300 bg-white px-3 py-1.5 text-xs font-medium text-gray-700 shadow-xs hover:bg-gray-50"
+            >
+              {t("Browse at this point")}
+            </Link>
+          </div>
+        </div>
+      )}
+
+      {diffHtml ? (
+        <div dangerouslySetInnerHTML={{ __html: diffHtml }} />
```
yolanfery, line 77: I am not totally sure now. Code looks way better (goodjob !) but we end up with a disjointed UI between this and the Codemirror diff/view we have with the AI pipeline agent. Maybe it is ok here for just showing a diff and probably more efficient. What do think of this ? 

Also we need to find a solution to avoid `dangerouslySetInnerHTML` (sanitize, use a react library,...) because in my opinion we can not trust `diff2html` to be bug free and always return a safe `html`

We might want to discuss those topics during the daily

backend/hexa/webapps/tests/test_models.py
DimitriKwihangana: Indeed

frontend/src/webapps/features/CommitDiff/CommitDiff.tsx
DimitriKwihangana: I agree about the UI inconsistency.  I went for diff2html for simplicity. We can always imitate the AI agent UI. no strong opinion. on dangerouslySetInnerHTML would wraping the diff2html output with DOMpurify before injecting be a good solution?

yolanfery: Ok, I am aligned with you :)  I propose to move forward with this, this is not a core feature, we can still bring consistency later, easier in this direction than the opposite 


Yes, using `DOMpurify` should be good enough imo

yolanfery (approved):

PR #1805, opened 2026-05-12
chore: Harden ourselves against supply chain attacks

bramj:
By setting the min release age to 1 week for both npm and renovate.

By not installing versions created in the last 7 days, we lower the possibility of getting hit by a supply chain attack.

This feels like a reasonable trade-off against slower security updates.

Sources:
- https://gajus.com/blog/3-pnpm-settings-to-protect-yourself-from-supply-chain-attacks
- https://docs.renovatebot.com/key-concepts/minimum-release-age/#minimum-release-age

### DIFF
--- a/frontend/.npmrc
+++ b/frontend/.npmrc
@@ -1,0 +1,2 @@
+# Only install package versions _older_ than 7 days to mitigate supply chain attacks
+min-release-age=7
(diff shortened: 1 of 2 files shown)


yolanfery (approved):
Makes sense

PR #1806, opened 2026-05-12
docs: fix incorrect Connection import path in toolbox docs

mcrimi:
## Summary

The DHIS2 toolbox examples in both the English and French docs imported `DHIS2Connection` from a module path that does not exist:

```python
from openhexa.sdk.workspaces.connections import DHIS2Connection 
```
The actual SDK module is singular — `openhexa.sdk.workspaces.connection` — as defined in [openhexa-sdk-python/openhexa/sdk/workspaces/connection.py](https://github.com/BLSQ/openhexa-sdk-python/blob/main/openhexa/sdk/workspaces/connection.py).

### DIFF
--- a/docs/en/toolbox.md
+++ b/docs/en/toolbox.md
@@ -35,7 +35,7 @@ Outside an OpenHEXA workspace, a connection can be manually created using the SD
 
 
 ``` python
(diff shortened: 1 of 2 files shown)


yolanfery (approved):
Thanks !!

PR #1807, opened 2026-05-12
feat: Pipelines template version name & documentation (HEXA-1546)

mrivar:
## Related PRs
- https://github.com/BLSQ/openhexa-app/pull/1807 (Merge first)
- https://github.com/BLSQ/openhexa-app/pull/1812

Adds a name and documentation value to the Pipeline Template versions.

## Changes

- Add `name` and `documentation` to pipeline template versions
- In the frontend:
  - Template Versions view: Show the name and documentation (editable only here)
  - Pipeline Template view: See documentation of last version (read-only)
  - Pipeline made from template view: See documentation of last applied version (read-only)

## How/what to test

1. Create a Pipeline Template and version
2. Add a documentation, if possible a long one to show that it is collapsible
3. Create a Pipeline from the template
4. Check how the new values are shown along the site (in the pipeline view, template view and template versions view)

## Screenshots / screencast

Template Versions view:
image

Pipeline Template view:
image

Pipeline from template view:
image


backend/hexa/pipeline_templates/models.py
```diff
@@ -103,7 +103,7 @@ class Meta:
 
     name = models.CharField(max_length=200)
     code = models.CharField(max_length=200, default="")
-    description = models.TextField(blank=True)
+    description = models.TextField(blank=True, null=True)
```
bramj: This will give your field 2 blank states: `None` and `""`, which is not great.

Better would be:
```suggestion
    description = models.TextField(blank=True, null=False)
```

TBH: I find the `blank` and `null` distinction is mega-confusing in Django, I found this helpful: https://www.jetbrains.com/guide/django/tips/django-blank-null/

backend/hexa/pipeline_templates/models.py
```diff
@@ -231,6 +235,8 @@ class Meta:
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created_at = models.DateTimeField(auto_now_add=True)
     version_number = models.PositiveIntegerField(editable=False)
+    name = models.CharField(max_length=200, blank=True)
+    documentation = models.TextField(blank=True, null=True)
```
bramj, line 239: I'm torn on this one!

I can see why you went for `documentation` for the naming, but I do wonder if we should go for consistency and choose `description`. We already have:
- `pipeline.description`
- `pipeline_version.description`
- `template.description`

So it feels like it's the sort of thing that can become confusing. In the end `description` of a pipeline version and `documentation` of a template version serve exactly the same purpose.

backend/hexa/pipeline_templates/models.py
```diff
@@ -231,6 +235,8 @@ class Meta:
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created_at = models.DateTimeField(auto_now_add=True)
     version_number = models.PositiveIntegerField(editable=False)
+    name = models.CharField(max_length=200, blank=True)
```
bramj: Maybe align on the length of the existing `pipeline_version.name`?
```python
# backend/hexa/pipelines/models.py
name = models.CharField(max_length=250, null=True, blank=True)
```

backend/hexa/pipeline_templates/migrations/0013_version_name_documentation.py
```diff
@@ -0,0 +1,26 @@
...
+            name="name",
+            field=models.CharField(blank=True, max_length=200, default=""),
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name="pipelinetemplateversion",
+            name="documentation",
+            field=models.TextField(blank=True, null=True),
+        ),
+    ]
```
bramj, line 26: WDYT of adding a migration to set the name + documentation of the template pipeline on the latest version of that same template pipeline?

frontend/src/pages/workspaces/[workspaceSlug]/templates/[templateCode]/index.tsx
```diff
@@ -144,6 +145,16 @@ const WorkspaceTemplatePage: NextPageWithLayout = (props: Props) => {
...
+          <RenderProperty
+            id="documentation"
+            accessor="documentation"
+            label={t("Version Documentation")}
+            readonly
+          >
+            {(property) => (
+              <CollapsibleMarkdown content={property.displayValue} />
+            )}
+          </RenderProperty>
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/templates/[templateCode]/index.tsx:157]: [Just a thought] I think at some point the `template.description` will become obsolete, when it's all set up correctly I'd be only interested in the "current" docs/description of the template.

frontend/src/core/components/CollapsibleMarkdown/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+          onClick={() => setExpanded(!expanded)}
+        >
+          {expanded ? t("Show less") : t("Show more")}
+        </button>
+      )}
+    </div>
+  );
+};
+
+export default CollapsibleMarkdown;
```
bramj, line 48: Nice addition! 👍

bramj (changes requested):
Nice one @mrivar ! Only some minor comments.

My main doubt is around the naming of `documentation`/`description`.

After giving it some more thought however, I wonder if at some point we should just only have "documentation" on the versions of pipelines + templates, and have the details page just show the info from the latest version.

I feel like that makes more sense, having the field duplicated on both the template/pipeline + version feels like a smell.
And in that case I like the `documentation` naming as a clear departure from `description`. I would like to hear your thoughts on this before we merge this one!

backend/hexa/pipeline_templates/models.py
```diff
@@ -291,6 +297,9 @@ def create_pipeline_version(
             if principal is None:
                 raise ValueError("principal is required when creating a new pipeline")
             pipeline = self._create_pipeline(principal, workspace)
+        elif self.documentation:
+            pipeline.description = self.documentation
+            pipeline.save(update_fields=["description"])
```
bramj: to double check

backend/hexa/pipeline_templates/models.py
mrivar: Yeah I actually had doubts in this one. I leaned towards it being `None` meant no description was set and `""` when it ahs been set as an empty string, but no strong opinion tbh. Keeping it as `null=False` works perfectly fine for me.

backend/hexa/pipeline_templates/models.py
mrivar: As spoken before, after checking user's behaviour it could make sense to have:
- `pipeline.description` and  `template.description`: the user can set what they need the pipeline/template to do, why their using that pipeline, etc
- `template_version.documentation` (and eventually `pipeline_version.documentation`, out of scope of this PR) for versions: Information about the specific version, usually set automatically by the README.md

backend/hexa/pipeline_templates/migrations/0013_version_name_documentation.py
mrivar: I'm not sure because I don't know if they would be fully related.
One idea we could have for a follow-up task tho, could be to add an action button in the versions view to automatically update the `documentation` of a version taking the README.md information (when the version has no documentation).

backend/hexa/pipeline_templates/models.py
mrivar: Good catch! It was indeed old code that didn't adjust anymore to the latest functionality of the description/documentation

backend/hexa/pipeline_templates/migrations/0013_version_name_documentation.py
bramj: Agreed, let's go with it like this for now 👍  The more I think about it, I think we'd ideally end up with only a `pipeline.description` and `README.md` in the version at some point in the future.

bramj (approved):
🚀  Good to go! 👏

PR #1811, opened 2026-05-13
fix:prevent scroll glitch in response streaming in chatpanel

DimitriKwihangana:
We were having issue of glitch where the agent was trying to smoothly scroll after each word. I  replaced  the smooth scrolling with instant during streaming to prevent animating glitches, added a isPinnedToBottom ref to stop auto scroll when a user  goes up.


bramj (commented):
Better! But did you look into removing the "refresh" of the whole conversation after each message? There's still a little flicker after a message because of this, and I wonder if it could be safely removed.

bramj (approved):
👍  Nice! Looking good

PR #1813, opened 2026-05-15
feat: Create Pipeline Agent: Disallow empty pipelines (HEXA-1634)

mrivar:
Don't allow the agent to create an empty pipeline

## Changes

- Modify MCP create pipeline tool to not accept an null source code

## How/what to test

Try to create a pipeline with the agent with an empty source code.

NOTE: Wasn't sure if it's better to modify the MCP tool itself or to add a wrapper in the agent that makes sure there is source code. Open to debate


DimitriKwihangana (approved):

yolanfery (approved):
Sounds like a good approach to me 👍

PR #1814, opened 2026-05-15
fix: Improve agents change proposals (HEXA-1646)

mrivar:
- Bug 1: After dismissing the proposal, it really is dismissed (Doesn't appear again after sending more messages to the agent)
- Bug 2: When the agent proposes a change, you can iterate over the proposed changes until accepting/dismissing them (making them vissible to the agent)

## Changes

- Tool Invocation: added a `resolved` attribute. Set to True when the tool proposal is dismissed or accepted.
- For bug 1: Set `resolved` to True when dismissing the proposal, it will be ignored from them on.
- Edit Agent: When a proposal hasn't been resolved yet, the agent sees those changes and proposes over them instead of the original files

### Bug 1
1. Go to a pipeline to edit with an agent
2. Propose a change
3. Dismiss it
4. Send a message to the agent without proposal (e.g.: "hi")

### Bug 2
1. Go to a pipeline to edit with an agent
2. Propose a change
3. Ask him to make a change over the proposal (it should iterate over the proposal instead of acting as if it doesn't know about it)


backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -84,7 +110,28 @@ def _extra_instructions(self) -> str:
...
-        if current_version.zipfile:
+        pending = (
+            ToolInvocation.objects.filter(
+                message__conversation=self.conversation,
+                tool_name="propose_pipeline_version",
+                success=True,
+                resolved=False,
+            )
+            .order_by("-created_at")
+            .first()
```
bramj, line 133:
```suggestion
            .latest("created_at")
```

https://docs.djangoproject.com/en/6.0/ref/models/querysets/#latest

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -84,7 +110,28 @@ def _extra_instructions(self) -> str:
...
+        if pending and pending.tool_output:
+            proposed_files = pending.tool_output.get("files", [])
+            if proposed_files:
+                lines.append("")
+                lines.append(
+                    "### Files (Pending Proposed Version — not yet accepted by the user)\n"
+                    "Use these files as the base for any further changes."
+                )
+                for f in proposed_files:
+                    lines.append(f"\n#### {f['name']}\n```\n{f['content']}\n```")
```
bramj, line 145: This works really well, love it! 👍 

It took me a moment to realize that the `pending.tool_output` actually contains _all_ files, and not just the ones that the agent decided to edit. Something to keep in mind when we move pipelines to git versioning, I suppose that could change this behaviour.

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx
```diff
@@ -117,6 +142,19 @@ const WorkspacePipelineCodePage: NextPageWithLayout = (props: Props) => {
...
+    const idToResolve = proposedToolInvocationId;
+    setProposedToolInvocationId(null);
+    if (idToResolve) {
+      resolveProposal({
+        variables: { toolInvocationId: idToResolve },
+        update(cache) {
+          cache.modify({
+            id: cache.identify({ __typename: "AssistantToolInvocation", id: idToResolve }),
+            fields: { dismissed: () => true },
+          });
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx]: This manual updating of the `cache` smells too complex, and it's duplicated in the `handleDismiss` function. There's probably room for improvement and at the very least pulling it into a helper.

backend/hexa/assistant/models.py
```diff
@@ -188,6 +188,7 @@ class ToolInvocation(Base):
     tool_input = models.JSONField()
     tool_output = models.JSONField(null=True, blank=True)
     success = models.BooleanField(default=True)
+    resolved = models.BooleanField(default=False)
```
bramj: I have an issue with this new field: it only makes sense for the `propose_pipeline_version` tool. E.g. records for `create_pipeline` are now `resolved=False`. That feels like something that will end up confusing later on and makes me feel like it's the wrong abstraction.

On the other hand, I feel like introducing a proper "Proposal" model feels like overkill. So what do you think about making the naming more explicit? Something like:

```python
proposal_pending = models.BooleanField(null=True, blank=True, default=None)
```

And here I'm making it explicitly a 3-state boolean. Normally I would avoid that, but here it could signal that for example a `create_pipeline` `ToolInvocation` doesn't have a proposal.
WDYT?

backend/hexa/assistant/graphql/schema.graphql
```diff
@@ -66,8 +67,19 @@ type CreateAssistantConversationResult {
...
+}
+
+type ResolveAssistantProposalResult {
+  success: Boolean!
+  errors: [ResolveAssistantProposalError!]!
+}
+
 extend type Mutation {
   createAssistantConversation(input: CreateAssistantConversationInput!): CreateAssistantConversationResult!
+  resolveAssistantProposal(toolInvocationId: UUID!): ResolveAssistantProposalResult!
```
bramj, line 83: What do you think of renaming?
```suggestion
  dismissAssistantProposal(toolInvocationId: UUID!): DismissAssistantProposalResult!
```
The resolver actually already carries this name.

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx
Comment (bramj) [reply frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx]: Actually, what do you think about moving this dismissal to the backend? Maybe it can be simpler as well: instead of dismissing a specific conversation, whenever I save a new version, we dismiss all potential pending proposals.

And ever further: I would maybe even dismiss even all pending proposals related to the _pipeline_ whenever we save a version. That ensures a "clean slate" every time we create a new version. It would be technically simpler and avoid people saving proposals of outdated versions.

backend/hexa/assistant/schema/mutations.py
```diff
@@ -89,4 +89,22 @@ def resolve_create_assistant_conversation(_, info, input, **kwargs):
...
+            id=tool_invocation_id
+        )
+    except ToolInvocation.DoesNotExist:
+        return {"success": False, "errors": ["NOT_FOUND"]}
+
+    if invocation.message.conversation.user != request.user:
+        return {"success": False, "errors": ["PERMISSION_DENIED"]}
+
+    invocation.resolved = True
+    invocation.save(update_fields=["resolved"])
```
bramj: Maybe safer: we can mark **all** pending proposals related to the conversation (or even the pipeline 🤔 ) as rejected.

bramj (changes requested):
This functionality was a lot more involved than I expected! 😅 
Good work, I really like how you enable the agent to not forget the unsaved changes 👍 

A bug that remains: Unresolved tool invocations can still pop up:
1. Edit a pipeline with AI -> proposal 1 (unresolved)
2. Don't accept or reject, but say: "Actually let's do it differently...." -> proposal 2 (unresolved)
3. Then you accept/reject the presented proposal 2
4. Refresh the page and you'll be presented with the proposal 1
Expected behaviour: creating proposal 2 implies a rejection of proposal 1.

I think my main doubt is around the new `resolved` field: it only makes sense for the `propose_pipeline_version` tool. E.g. records for `create_pipeline` are now `resolved=False`. That feels like something that will end up confusing later on and makes me feel like it might be the wrong abstraction.

I added a few comments about renaming that I think can improve this, curious to hear what you think! We should try to get it right since this we'll get similar tools such as `propose_webapp_version` etc later on.

backend/hexa/assistant/agents/edit_pipeline_agent.py
mrivar: I didn't know about `.latest()`! However it raises a `DoesNotExist` exception if there are no tool invocations, while order_by + first returns None, so I think first would be better in this case?

From an efficiency point of view there's no change, as they both do `ORDER BY created_at DESC LIMIT 1` behind doors

backend/hexa/assistant/graphql/schema.graphql
mrivar: this mutation is also called when a version is accepted, so I actually think the name of the resolver is wrong, which I hadn't noticed!

I will think of a better naming than resolved tho. I like `pending` as you proposed before

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx
Comment (mrivar) [reply frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx]: You're right, this cache modification is a code smell, the frontend shouldn't be in charge of this logic.
I'm going to change it to a backend reload so it's not hardcoded

Agree on the dismissing all proposal approach as well

backend/hexa/assistant/models.py
mrivar: I went for `pending` over `proposal_pending` but added a Django's `help_text` that will appear in the admin section explaining the field

backend/hexa/assistant/agents/edit_pipeline_agent.py
bramj: 👍 gotcha, let's keep it like this!

backend/hexa/assistant/migrations/0004_toolinvocation_resolved.py
```diff
@@ -11,7 +11,14 @@ class Migration(migrations.Migration):
...
-            field=models.BooleanField(default=False),
+            name="pending",
+            field=models.BooleanField(
+                default=False,
+                help_text=(
+                    "True while the proposal is awaiting user action; False once accepted or dismissed. "
+                    "Only meaningful for tool invocations that produce a user-facing proposal "
+                    "(e.g. propose_pipeline_version)."
+                ),
+            ),
```
bramj, line 22: Sorry to give pushback on this, but I think `pending` is confusing here. It makes it sound like the tool invocation itself is "pending", but the tool actually executed correctly, it just comes with a proposal that is pending.
I think the naming should really make clear that this field is strictly related to a proposal.

bramj (changes requested):
Sorry to send this back after so long, I'll try to be faster with my reviews.

But I have 2 comments: 
- I think the naming should really reflect that it concerns a "proposal". 
- I think something got lost with the rename: when a `propose_pipeline_version` `ToolInvocation` is created, it's `pending` state should be `True`, right? Maybe there's a `_on_tool_result` hook missing for this?

bramj (approved):
Great! 👍

PR #1815, opened 2026-05-18
fix: Don't corrupt web apps images upon upload (HEXA-1564)

bramj:
## Changes

- New file `encoding` GraphQL enum (`TEXT` / `BASE64`) lets webapp files round-trip binary content (e.g. images) instead of corrupting it as UTF-8.
- Forgejo client detects binary blobs on read and encodes on write.
- Frontend uploads always base64-encode with new helper `fileToBase64` (extracted from `PipelineFilesEditor`), but uploads as text when doing file edits.
- Note: Binaries are not correctly displayed yet, this will be done in a later PR.

## How/what to test

- Upload an image into a static webapp, verify that it renders correctly after save.
- Edit index.html inline, this still saves as text.


backend/hexa/git/tests/test_forgejo.py
```diff
@@ -574,6 +574,138 @@ def test_unarchive_repository_not_found(self):
...
+        client = ForgejoClient(url=FORGEJO_URL, username=USERNAME, password=PASSWORD)
+        client.commit_files(
+            "my-repo",
+            [{"path": "hello.txt", "content": "héllo", "encoding": "TEXT"}],
+            "msg",
+            "u",
+            "u@example.com",
+        )
+
+        import json
```
yolanfery: [Nitpick] no reason to dynamically do import here

Maybe we want to add this to CLAUDE.md

backend/hexa/git/forgejo.py
```diff
@@ -148,14 +148,14 @@ def commit_files(
             path = file["path"]
             is_update = path in existing_tree
             content = file["content"]
-            if isinstance(content, str):
-                content = base64.b64encode(content.encode()).decode()
-            elif isinstance(content, bytes):
-                content = base64.b64encode(content).decode()
+            if file.get("encoding") == "BASE64":
```
yolanfery: [Minor] We might want to use a enum for BASE64 and TEXT bound to the GraphQL type like : 

```
pipeline_notification_level_enum = EnumType(
    "PipelineNotificationLevel", PipelineNotificationLevel
)
```

yolanfery (approved):
Highly prefer this approach, well done ! Just 2 minor comments

Tested, it looks better : 
Image

backend/hexa/git/forgejo.py
bramj: Good catch, that's a good idea, thanks :+1:

CLAUDE.md
```diff
@@ -5,6 +5,7 @@
 Follow these guidelines when writing code:
 
 -   **Comments**: Do not write comments that merely restate what the code does. Only add comments when explaining _why_ something is done a certain way, not _what_ it does.
+-   **Imports**: Place imports at the top of the file. Do not use inline/dynamic imports inside functions or methods unless strictly necessary (e.g., to break a circular import or to defer an optional/expensive dependency).
```
yolanfery, line 8: Thanks 👌

yolanfery (approved):

PR #1818, opened 2026-05-20
fix: docker pipeline runner is restarting all running pipelines on restart (HEXA-1660)

yolanfery:
We previously reworked the pipeline runner to be stateless (it can be restarted without effect on running pipelines) but we left out the docker path.

Fix it for docker.

Tested locally : 
- Started a long-running pipeline; confirmed a single container named `pipeline-<workspace>-<code>-<runid>` with label `hexa-run-id=<uuid>`.
- `docker compose restart pipelines_runner`
- After restart, docker ps still showed the same single container.
- Runner logs confirmed the re-attach path:
```
  Found 1 orphaned RUNNING pipeline(s) to re-attach gradually
  Re-attaching to orphaned run: 5af14409-...
  Re-attached to container 1bd804599dc7... for run 5af14409-...
```


bramj (approved):
Tested and working great! Thanks for adding this code path 👍

PR #1819, opened 2026-05-20
fix: allow to create a complete version from a pipeline creation (HEXA-1661)

yolanfery:
We previously allowed to create a version using a zipfile. 

But we want to extend that to allow creating a version with all the required details to allow the CLI to create a pipeline atomically

### DIFF
--- a/backend/hexa/pipelines/graphql/schema.graphql
+++ b/backend/hexa/pipelines/graphql/schema.graphql
@@ -390,6 +390,20 @@ enum PipelineOrderBy {
 }
 
 """
+Configures the first pipeline version, created atomically alongside the pipeline.
+Providing this sub-input signals that a first version should be created.
+"""
+input CreatePipelineVersionInput {
+  zipfile: String!  # Base64-encoded ZIP file containing the version code.
+  name: String  # The name of the first pipeline version.
+  description: String  # The description of the first pipeline version.
+  externalLink: URL  # The external link associated with the first pipeline version.
+  parameters: [ParameterInput!]  # The parameters of the first pipeline version (defaults to those parsed from the zipfile).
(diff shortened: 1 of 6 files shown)


backend/hexa/pipelines/graphql/schema.graphql
```diff
@@ -400,7 +414,7 @@ input CreatePipelineInput {
   notebookPath: String  # The path to the notebook file for the pipeline.
   functionalType: PipelineFunctionalType  # The functional type of the pipeline.
   tags: [String!]  # The names of tags to associate with the pipeline.
-  zipfile: String  # Optional base64-encoded ZIP file to upload as the first pipeline version.
+  version: CreatePipelineVersionInput  # Optional first pipeline version, created atomically with the pipeline.
```
yolanfery, line 417: We could keep zipFile and avoid the breaking change. But I think it is worth doing this clean up now that there is low usage of it

bramj: Agreed 👍

backend/hexa/pipelines/schema/mutations.py
```diff
@@ -386,19 +426,7 @@ def resolve_upload_pipeline(_, info, **kwargs):
...
-                max_allowed_timeout, subscription.max_pipeline_timeout
-            )
-
-        if input.get("timeout") and (
-            input.get("timeout") < 0 or input.get("timeout") > max_allowed_timeout
-        ):
-            raise InvalidTimeoutValueError(
-                "Pipeline timeout value cannot be negative or greater than the maximum allowed value."
-            )
+        _validate_pipeline_version_timeout(input.get("timeout"), pipeline.workspace)
```
bramj, line 429: 👍

bramj (approved):
Looking good, working great!

PR #1820, opened 2026-05-21
fix: increase request size limit (HEXA-1662)

yolanfery:
Some users face issues when trying to upload big webapps.

There are actually 2 set limits : 
- Next.js, if not explicitly set will limit requests to 1Mb.
- Django by default will limit to 2Mb.

When the Next.js limit is met, no sentry issue was sent. 

## Changes

- Ensure that the Next.js limit is set and larger than the Django one
- Double the Django limit
- Add a middleware to convert DataTooBig exceptions to a 413 that can be caught in the Frontend
- Graceful error parsing in the Frontend to handle 413

## Screenshots / screencast

Screenshot 2026-05-21 at 11 04 45


backend/config/settings/base.py
```diff
@@ -424,6 +426,11 @@
 GRAPHQL_DEFAULT_PAGE_SIZE = 10
 GRAPHQL_MAX_PAGE_SIZE = 10_000
 
+DATA_UPLOAD_MAX_MEMORY_SIZE = int(
+    os.environ.get(
+        "DATA_UPLOAD_MAX_MEMORY_SIZE", 2 * global_settings.DATA_UPLOAD_MAX_MEMORY_SIZE
+    )
+)
```
bramj, line 430: Why not set it explicitly? I feel like this makes a bit hard to understand what the current setting is.
Also: this means we set at 5MB in Django and 10MB on the frontend?

bramj (changes requested):
I am a bit hesitant on the middleware complexity, and my main issue is that I think there's misalignment between the front (10MB) and back (5MB).

What if:
- we try to unify into 1 single ENV var that sets the max upload size
- and then we expose it to the frontend as well so we can check if a file is too big _before_ uploading. That way we won't need any middleware at all.

WDYT?

backend/config/settings/base.py
yolanfery: Changed to explicit number. Yes FE limit is larger on purpose to hit the backend error and generate a sentry issue for visibility

yolanfery:
> I am a bit hesitant on the middleware complexity, and my main issue is that I think there's misalignment between the front (10MB) and back (5MB).
> 
> What if:
> 
> * we try to unify into 1 single ENV var that sets the max upload size
> * and then we expose it to the frontend as well so we can check if a file is too big _before_ uploading. That way we won't need any middleware at all.
> 
> WDYT?

Yes this is also an option. I am bit torn on this one. 

The middleware ensures we get a descriptive error when hitting such errors from any source (the CLI, the MCP, the frontend,...) instead of getting a generic 400 so that's why I took this route.

I don't think misalignement between the BE and the FE is an issue. We actually lived with such a disalignement so far but I can understand the urge to get it sorted for clarity. Also it saves bandwidth and time.

My opinion is to keep the current approach but I won't be against the frontend check.

bramj (approved):
> Yes this is also an option. I am bit torn on this one.
> 
> The middleware ensures we get a descriptive error when hitting such errors from any source (the CLI, the MCP, the frontend,...) instead of getting a generic 400 so that's why I took this route.
> 
> I don't think misalignement between the BE and the FE is an issue. We actually lived with such a disalignement so far but I can understand the urge to get it sorted for clarity. Also it saves bandwidth and time.
> 
> My opinion is to keep the current approach but I won't be against the frontend check.

Fair enough. I'm not really sure what's the best approach, but I don't have a strong opinion on the alternative, and I think both options are a big improvement on what we have now.
So agreed, let's keep it like this 👍

PR #1822, opened 2026-05-21
feat: Sort pipelines by creation date (HEXA-1655)

mrivar:
Add the ability to sort pipelines by creation date

## Changes

- Add the possibility to sort by creation date
- Resolve creation date also in card view

## How/what to test

- Try to sort pipelines by creation date


frontend/src/pipelines/features/Pipelines/GridView.tsx
```diff
@@ -127,9 +127,9 @@ const GridView = ({
           )}
         </BaseColumn>
         <DateColumn
-          accessor={"currentVersion.createdAt"}
```
DimitriKwihangana, line 130: I like that we are not using the current version but the pipeline creation date!.

DimitriKwihangana (approved):
It look very good and simple! Nice one!

frontend/src/pipelines/features/Pipelines/GridView.tsx
mrivar: Yes! I feel if we want the `currentVersion.createdAt`, that should be an Updated At column or somthing like that 🤔

bramj (approved):
Great! 👍

PR #1824, opened 2026-05-21
fix: split toolbox documentation (HEXA-1653)

yolanfery:
Split toolbox documentation to have a better index, especially for AI


bramj (approved):
Nice one! 👍

PR #1825, opened 2026-05-21
fix: add GraphQL documentation for static webapps + remove delete operations from webapp scope (HEXA-1653)

yolanfery:
Users and agent need to be able to understand the schema -> Adding collapsible details for each section

Also ensure delete operations are not part of webapp' scopes

Screenshot 2026-05-21 at 16 25 20

### DIFF
--- a/backend/hexa/webapps/graphql_proxy.py
+++ b/backend/hexa/webapps/graphql_proxy.py
@@ -29,7 +29,6 @@ SCOPE_FIELDS = {
     Webapp.OperationScope.FILES_WRITE: {
         "prepareObjectUpload",
         "createBucketFolder",
-        "deleteBucketObject",
         "writeFileContent",
     },
     Webapp.OperationScope.DATASETS_READ: {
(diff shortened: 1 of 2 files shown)


yolanfery:
I considered and explored giving a dynamic full breakdown of the GraphQL tree for each scope but it's quite big and was only serving AI agents through a tool. This simpler solution is not complete but is already helping to understand the basis and is same for human and ai agents

backend/hexa/webapps/graphql_proxy.py
```diff
@@ -41,12 +40,9 @@
     Webapp.OperationScope.DATASETS_WRITE: {
         "createDataset",
         "updateDataset",
-        "deleteDataset",
         "createDatasetVersion",
         "updateDatasetVersion",
-        "deleteDatasetVersion",
         "createDatasetVersionFile",
-        "deleteDatasetLink",
```
bramj, line 49: Auwch, didn't catch that last time...

bramj (approved):
> I considered and explored giving a dynamic full breakdown of the GraphQL tree for each scope but it's quite big and was only serving AI agents through a tool. This simpler solution is not complete but is already helping to understand the basis and is same for human and ai agents

I also feel there's room for improvement on this, but not sure yet what the right way forward would be.
So agreed on the simpler solution to help our human and non-human users out.

PR #1826, opened 2026-05-21
fix: improve get_help_or_doc index of reachable documentation + improve documentation summaries (HEXA-1653)

yolanfery:
- Improve summaries (for both human and machine)
- Improve the index of available topic of `get_help_or_doc`
- Update the snapshot test


backend/hexa/mcp/tests/test_help.py
```diff
@@ -57,6 +57,14 @@ def test_docstring_lists_curated_topics(self):
             "- Pass a topic name to fetch that doc page in full. Reason is optional here.\n"
             "\n"
             "\n"
-            "    Available topics: cli, notebooks-advanced, sdk, static-webapps, toolbox-dhis2, toolbox-hexa, toolbox-iaso, writing-pipelines."
+            "    Available topics:\n"
+            "    - cli: OpenHEXA comes with a CLI you can install globally on your system. This CLI allows you to interact with the OpenHEXA API and perform various tasks such as creating and managing pipelines, running local jobs and more.\n"
```
yolanfery, line 61: It makes it a bit more tedious/repetitive to update the documentation but I think it's great to view what is fed to the AI and be aware than modifying the docs impacts the AI context. 

But we can remove/downscope this test anytime if needed 🙂

bramj: Agreed! I actually wanted to show it to JP yesterday and indeed it was hard to understand what was being added exactly to the context.

backend/hexa/mcp/tests/test_help.py
```diff
@@ -57,6 +57,14 @@ def test_docstring_lists_curated_topics(self):
...
+            "    - sdk: The `openhexa.sdk` Python library lets you write pipelines and notebook code against OpenHEXA: a `workspace` helper for files, database, connections, datasets and webapps, plus a typed `OpenHexaClient` for programmatic and GraphQL access.\n"
+            "    - static-webapps: Static webapps host an HTML/CSS/JavaScript bundle inside an OpenHEXA workspace, served on its own subdomain, and can query OpenHEXA on behalf of viewing users via a pre-authenticated GraphQL endpoint — great for dashboards and custom forms.\n"
+            "    - toolbox-dhis2: The `openhexa.toolbox.dhis2` library connects to a DHIS2 instance with metadata caching, queries org units, data elements and indicators, fetches dataValueSets and analytics, and handles ISO periods and pyramid enrichment.\n"
+            "    - toolbox-hexa: DEPRECATED — the `openhexa.toolbox.hexa` GraphQL client is no longer maintained. Use the typed `OpenHexaClient` documented in the [SDK guide](sdk.md#using-the-openhexa-client) instead.\n"
```
bramj: I wonder if it makes sense to keep this one 🤔  I suppose you could be editing a legacy pipeline and need it, but it also uses context ⚖️ . Not sure what the best approach will be, just thinking out loud here.

bramj (approved):
I think this is really starting to come together nicely! 👍

backend/hexa/mcp/tests/test_help.py
yolanfery: As discussed, I have put forward the new client instead and pushed down the deprecated client to the bottom of the page so that it's still avialable

PR #1830, opened 2026-05-25
fix: unknown format in dynamic choices param (HEXA-1620)

mrivar:
Found a small bug when trying out weird things in the dynamic choices parameter. We currently let the user set a format (in case he has a file without the extension in the name or what not, just for edge cases tbh, otherwise it is automatically handled).
I found an edge case in which, if the user somehow manages to set a `ChoicesFromFile` with an unsupported format, the whole app would clash. The user would not be able to navigate to the pipeline code. I've changed it so if we don't recognise the format we just treat as if it was not given on the first place.
Here is an example of a parameter that would cause this error:

```python
@parameter(
    "wrong_param",
    name="Wrong Param",
    choices=ChoicesFromFile("choices.csv", format="non-existing-format"),
    type=int,
)
```

### DIFF
--- a/backend/hexa/pipelines/choices_from_file.py
+++ b/backend/hexa/pipelines/choices_from_file.py
@@ -32,7 +32,11 @@ def resolve_choices_from_file(bucket_name: str, choices_from_file: dict) -> list
         raise ValueError(f"Invalid file path '{path}'.")
 
     raw_fmt = choices_from_file.get("format")
-    fmt = PipelineParameterChoicesFileFormat.from_extension(raw_fmt) if raw_fmt else PipelineParameterChoicesFileFormat.from_path(path)
+    fmt = (
+        PipelineParameterChoicesFileFormat.from_extension(raw_fmt)
+        if raw_fmt
+        else PipelineParameterChoicesFileFormat.from_path(path)
+    )
     column = choices_from_file.get("column")
 
     try:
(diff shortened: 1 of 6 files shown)


backend/hexa/pipelines/choices_from_file.py
```diff
@@ -7,20 +7,22 @@
...
 
 MAX_CHOICES_FILE_SIZE = 5 * 1024 * 1024  # 5 MB
 
-_SUPPORTED_FORMATS = {"csv", "json", "yaml"}
-
 
 def _detect_format(path: str) -> str:
     ext = path.rsplit(".", 1)[-1].lower() if "." in path else ""
     if ext == "yml":
         return "yaml"
```
yolanfery: Handling of the format is disjointed (there are repetitions and scattered logic) and untyped (we rely on str) 
See : 
- https://github.com/BLSQ/openhexa-app/blob/a64ad50ff527d697863867c05138e0f4d706a59e/backend/hexa/pipelines/choices_from_file.py#L75-L82
- https://github.com/BLSQ/openhexa-app/blob/a64ad50ff527d697863867c05138e0f4d706a59e/backend/hexa/pipelines/choices_from_file.py#L49-L52

I propose to : 
- Define the parsing logic in the Enum class
- Rely on Enums for comparisons/branching
- Avoid duplicate logic like     
      ```
          if fmt == "yml":
               fmt = "yaml"
       ```

mrivar: I like that, and deeply prefer it, thanks for the suggestion!

yolanfery (approved):
Code looks way better, well done !

PR #1831, opened 2026-05-25
fix: Long Runs parameters not breaking layout (HEXA-1656)

mrivar:
Pipeline Run parameters that are very long (such as the DHIS2 Connection) break the layout of the pipeline runs page. This PR solves that by truncating the parameters text and, if hidden, showing the information on a tooltip.

## Changes

- Added a component that truncates text
- Used the component to truncate run parameters

## How/what to test

Run a pipeline with really long parameters. DHIS2 connections is an example of parameters being really long.

## Screenshots / screencast

image

Old Run page:
image

Fixed Run page:
image

### DIFF
--- a/frontend/src/core/components/DataCard/__tests__/__snapshots__/DataCard.test.tsx.snap
+++ b/frontend/src/core/components/DataCard/__tests__/__snapshots__/DataCard.test.tsx.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
 exports[`DataCard renders 1`] = `
 <div>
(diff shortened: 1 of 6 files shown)


DimitriKwihangana (commented):
Image

It does look good @mrivar  ! I at some point tested the name of the parameter instead of the value.😂 . Do you think it at some point we would need to truncate the names of parameters? cc: @mcrimi

mrivar:
That's some good testing @DimitriKwihangana ! I think it should truncate it as well, or at least make it look nicer hahah, even if it's an edge case

DimitriKwihangana (approved):
This looks cooler! Nice @mrivar

bramj (approved):
Nice improvements @mrivar ! Thanks for adding this 👍

PR #1835, opened 2026-05-27
refactor: `filter_for_user` pattern to support `ServicePrincipal` (HEXA-1619)

yolanfery:
We want to enable webapps to query the graphql layer on behalf of viewing user (in https://github.com/BLSQ/openhexa-app/pull/1836). We need to restrict access for those request to the webapp's workspace. We need to refactor the `filter_for_user` pattern to allow this.

At first we wanted to add two accessors (`accessible_workspaces()`, `accessible_organizations()`) and a `is_service_principal` flag to UserInterface. But there is no reason to add a submethod that actually represent `Workspace` and `Organization` `filter_for_user`. Also Inheriting from `ServicePrincipal` is more practical and idiomatic than the `is_service_principal` flag 

## Changes

- Build a solid and tested `Workspace` and `Organization` `filter_for_user` that can be used in other `filter_for_user` 
- Refactor `filter_for_user` accross assets to make use of `Workspace` and `Organization` `filter_for_user`
- Add some tests to avoid regressions
- Ensure the `ServicePrincipal` real user is used when possible creating assets etc


backend/hexa/datasets/tests/test_models.py
```diff
@@ -898,3 +898,41 @@ def test_pipeline_user_cannot_access_non_shared_dataset_links_from_same_organiza
         links = DatasetLink.objects.filter_for_user(self.pipeline_user)
         link_datasets = [link.dataset for link in links]
         self.assertNotIn(self.DATASET_IN_WORKSPACE_3, link_datasets)
+
+    def test_pipeline_user_sees_org_shared_datasets_via_dataset_queryset(self):
+        """A pipeline run sees datasets shared with its workspace's organization."""
```
yolanfery, line 942: New feature with this refactor, this is a great win on this edge case

backend/hexa/user_management/models.py
```diff
@@ -87,6 +87,20 @@ def is_authenticated(self):
...
+    Marker mixin for principals that impersonate a workspace rather than a
+    real user account. Subclasses must expose `real_user` — the underlying
+    user that ultimately triggered the action — for audit and attribution
+    (e.g. `created_by`). May be `None` when no human triggered the action,
+    such as scheduled pipeline runs.
+    """
+
+    @property
+    def real_user(self) -> User | None:
+        raise NotImplementedError
```
bramj, line 103: This is great!

What do you think about adding 2 more properties here to expose the `workspace` and `workspace_id`? 

Implementation for `PipelineRunUser` can be something like

```python
  @property
  def workspace(self):
      return self.pipeline_run.pipeline.workspace

  @property
  def workspace_id(self):
      return self.pipeline_run.pipeline.workspace_id
```

By doing the same thing for the `WebappUser`, we could get rid of the distinction between the two in `WorkspaceQuerySet` and just check on `ServicePrincipal`.

backend/hexa/workspaces/models.py
```diff
@@ -337,12 +347,9 @@ def generate_new_database_ro_password(self, *, principal: User):
...
 class WorkspaceMembershipQuerySet(BaseQuerySet):
-    def filter_for_user(self, user: AnonymousUser | User) -> models.QuerySet:
-        return self._filter_for_user_and_query_object(
-            user,
-            Q(workspace__members=user),
-            return_all_if_superuser=False,
-            return_all_if_organization_admin_or_owner=True,
+    def filter_for_user(self, user: AnonymousUser | UserInterface) -> models.QuerySet:
+        return self.filter(
+            workspace__in=Workspace.objects.filter_for_user(user, include_archived=True)
```
bramj: Do you think we should get rid of `_filter_for_user_and_query_object` over the longer term?

bramj: This preserves the old behaviour, but does `include_archived=True` actually make sense here? 🤔

backend/hexa/pipelines/authentication.py
```diff
@@ -1,15 +1,20 @@
...
         super().__init__()
         self.pipeline_run: PipelineRun = pipeline_run
 
     is_active = True
     is_authenticated = True
 
+    @property
+    def real_user(self):
+        # The user who triggered the run, if any (None for scheduled runs).
+        return self.pipeline_run.user
```
bramj, line 16: Awesome 👍

backend/hexa/datasets/models.py
```diff
@@ -580,35 +573,22 @@ def filter_for_workspaces(self, workspaces, pinned=None, query=None):
...
+        return self.optimize_query(
+            self.filter(
+                Q(workspace__in=accessible_workspaces)
+                | Q(
+                    dataset__shared_with_organization=True,
+                    workspace__organization__in=accessible_organizations,
                 )
-            )
+            ).distinct()
+        )
```
bramj, line 584: So much simpler 👍

backend/hexa/datasets/models.py
```diff
@@ -37,90 +42,69 @@ def create_dataset_slug(name: str, workspace):
...
 
-        # FIXME: Use a generic permission system instead of differencing between User and PipelineRunUser
-        if isinstance(principal, PipelineRunUser):
-            if principal.pipeline_run.pipeline.workspace != workspace:
+        if isinstance(principal, ServicePrincipal):
+            if (
+                not Workspace.objects.filter_for_user(principal, include_archived=True)
+                .filter(pk=workspace.pk)
+                .exists()
+            ):
```
bramj: This could also become then `if principal.workspace != workspace:` with the refactoring I proposed

backend/hexa/datasets/models.py
```diff
@@ -37,90 +42,69 @@ def create_dataset_slug(name: str, workspace):
...
-                ),
-                return_all_if_superuser=False,
-            )
+        # Datasets auto-link to their primary workspace, so `links__workspace`
+        # uniformly covers both primary and shared scopes — no separate
+        # primary-workspace clause needed.
+        return (
+            self.filter_for_user(user)
+            .filter(links__workspace__slug__in=workspace_slugs)
+            .distinct()
```
bramj, line 77: ❤️

backend/hexa/datasets/tests/test_models.py
bramj: That's great!

backend/hexa/user_management/models.py
```diff
@@ -822,12 +840,8 @@ class OrganizationInvitationStatus(models.TextChoices):
...
 
 class OrganizationInvitationQuerySet(BaseQuerySet):
-    def filter_for_user(self, user: AnonymousUser | User) -> models.QuerySet:
-        return self._filter_for_user_and_query_object(
-            user,
-            Q(organization__organizationmembership__user=user),
-            return_all_if_superuser=False,
-        )
+    def filter_for_user(self, user: AnonymousUser | UserInterface) -> models.QuerySet:
+        return self.filter(organization__in=Organization.objects.filter_for_user(user))
```
bramj:
```suggestion
        return self.filter(
            organization__in=Organization.objects.filter_for_user(
                user, direct_membership_only=True
            )
        )
```

Without this, via a GraphQL query like
```graphql
query {
  organization(id: "<uuid>") {
    invitations {
      items {
        email
        role
        invitedBy {
          email
        }
      }
    }
  }
}
```
an external collaborator could see all organization invitations.
Best to add a test for this as well.

backend/hexa/datasets/models.py
```diff
@@ -37,90 +42,69 @@ def create_dataset_slug(name: str, workspace):
...
+        return self.optimize_query(
+            self.filter(
+                Q(links__workspace__in=accessible_workspaces)
+                | Q(
+                    shared_with_organization=True,
+                    workspace__organization__in=accessible_organizations,
                 )
-            )
+            ).distinct()
+        )
```
bramj, line 66: I'm not sure where, but I think a regression slipped in: external collaborators can now see datasets from _other workspaces_ that are "shared with organization". Perhaps this was already a bug, but an external collaborator should not see these.

To reproduce:
1. Invite external collaborator to workspace A
2. Create and share dataset in workspace B
3. Log in as external collaborator and you'll see the dataset

bramj (changes requested):
@yolanfery 👏 Honestly, this is a massive improvement of a refactoring. It greatly simplifies this complex logic and makes it more readable 👍 

This was a though one to review 😅 
I see two real issue (datasets + invitations), and a few minor comments.

backend/hexa/user_management/models.py
yolanfery: Great idea thanks. Done in next rev :-)

backend/hexa/workspaces/models.py
yolanfery: > Do you think we should get rid of _filter_for_user_and_query_object over the longer term?

Yes I think the new pattern is better but we can keep it for now to reduce regression risks..

> This preserves the old behaviour, but does include_archived=True actually make sense here? 🤔

Good point, yes I removed those in next rev, thanks

backend/hexa/datasets/models.py
yolanfery: Great improvement indeed, I like it, modifying this in next rev

yolanfery: It even removes dynamic imports ❤️

backend/hexa/user_management/models.py
yolanfery: Indded, fixed and added tests, thanks

backend/hexa/datasets/models.py
yolanfery: Wasn't this the case before ?

Any way I did a commit (https://github.com/BLSQ/openhexa-app/pull/1835/commits/b51b78514f175913d274948b92a5f527397b324e) to fix it and provide a unit test to cover it

BUT I am personally not sure such datasets should be hidden to external collaborators. The goal of org sharing is to share the dataset with all workspaces in the org. So a pipeline in a different workspace could use it. We probably want external collaborator to see datasets used by their workspace pipelines no ? 

Please let me know your thoughts to see if this commit needs to be reverted or not

bramj: Good point! That's another way of looking at it indeed. My feeling was that if I (for example as a BLSQ employee) share across the org, then I expect to share the dataset with my colleagues, but not with a client who was invited for collaboration on some workspace.

I guess both points are valid, and it's a product question, we should discuss this with @mcrimi .
To not block this PR, I guess blocking it as you did is a safe choice, and we can revert later.

yolanfery: > I guess both points are valid, and it's a product question, we should discuss this with @mcrimi .
To not block this PR, I guess blocking it as you did is a safe choice, and we can revert later.

Totally aligned with you there

backend/hexa/datasets/models.py
```diff
@@ -88,14 +88,8 @@ def create_if_has_perm(
...
     ):
-        from hexa.workspaces.models import Workspace
-
         if isinstance(principal, ServicePrincipal):
-            if (
-                not Workspace.objects.filter_for_user(principal, include_archived=True)
-                .filter(pk=workspace.pk)
-                .exists()
-            ):
+            if principal.workspace_id != workspace.pk:
```
bramj, line 92: ❤️

bramj (approved):
@yolanfery Great job, this turned out really nicely 👍 😁

PR #1836, opened 2026-05-27
feat: implement WebappUser class for querying GraphQL (HEXA-1619)

yolanfery:
Ensure requests from webapps are scoped to the webapp's workspace

Depends on : https://github.com/BLSQ/openhexa-app/pull/1835


backend/hexa/webapps/models.py
```diff
@@ -390,3 +390,20 @@ def update_dashboard(self, superset_instance, external_dashboard_id):
...
+
+    @classmethod
+    def from_user(cls, user: User, webapp: Webapp) -> "WebappUser":
+        """Cast an existing User row to a WebappUser view tied to `webapp`."""
+        instance = cls.objects.get(pk=user.pk)
+        instance.webapp = webapp
+        return instance
+
+    def get_username(self):
+        return f"webapp_{self.webapp.id}_as_{self.email}"
```
bramj, line 432: Very cool 👍

backend/hexa/pipelines/authentication.py
```diff
@@ -10,11 +10,6 @@ def __init__(self, pipeline_run: PipelineRun):
     is_active = True
     is_authenticated = True
 
-    @property
-    def real_user(self):
-        # The user who triggered the run, if any (None for scheduled runs).
-        return self.pipeline_run.user
-
```
bramj, line 17: Not sure why this is removed?

backend/hexa/datasets/models.py
```diff
@@ -382,11 +376,7 @@ def create_if_has_perm(
         ):
             raise PermissionDenied
 
-        created_by = (
-            principal.real_user
-            if isinstance(principal, ServicePrincipal)
-            else principal
-        )
+        created_by = None if isinstance(principal, PipelineRunUser) else principal
```
bramj: Not sure this is an improvement?

bramj (commented):
Just a few questions, big review is on the other PR 👍

backend/hexa/datasets/models.py
yolanfery: Wanted to reduce the number of changes/diff with the original codebase but yeah, might be a good abstraction to keep, will add it back

bramj (approved):
This confirms that the abstraction holds up really nicely here, awesome 👍

PR #1838, opened 2026-05-27
feat: Visualise agent tools (HEXA-1650)

mrivar:
Visualise agent tools in assistant conversations

## Changes

- **Backend: message content segments migration:** Assistant's `Message`s now store content as a json: classifying and ordering internal messages and tool calls, so they're reordered properly in the frontend
- **Added DB migration** so old `Message`'s get the proper structure
- **Visualise agent tool calls in the chat:** On tool call, show it in the frontend as a separate bubble (both for SSE and recovering old conversations). Used a dedicated UI **ToolCallCard component** with 3 states (loading, done, error).
- **Tool name translations:** Automatically mapping of tool naming in the frontend (default with messages.json, backup with automated helper `toolNames.ts`)
- **ChatPane refactor:** The chat pane was reorganised to support rendering tool call segments alongside text segments, with improved layout and streaming behaviour

Note: Added label `approved-breaking-change`, as the change is AssistantMessage which is only used internally

## How/what to test

Run some agent chats to see how it acts on tool calls

## Screenshots / screencast

Screenshot from 2026-05-29 12-16-32

image


frontend/src/assistant/helpers/toolNames.ts
```diff
@@ -0,0 +1,65 @@
...
+  toolName: string,
+  labels?: Record<string, string>,
+): string {
+  if (labels?.[toolName]) return labels[toolName];
+
+  for (const [prefix, verb] of PREFIX_VERBS) {
+    if (toolName.startsWith(prefix)) {
+      const rest = toolName.slice(prefix.length).replace(/_/g, " ");
+      return `${verb} ${rest}`;
+    }
```
yolanfery, line 58: Cool defensive trick

backend/hexa/assistant/agents/base.py
```diff
@@ -121,7 +145,7 @@ async def run_stream(self, user_input: str) -> AsyncGenerator[str, None]:
         user_msg = await Message.objects.acreate(
             conversation=self.conversation,
             role=Message.Role.USER,
-            content=user_input,
+            content=[{"type": "text", "content": user_input}],
```
yolanfery: Instead of a JSON, maybe we want to type this content ? 
I feel like we will benefit from an abstraction, it should be easier to manipulate and maintain in both BE and FE

backend/hexa/assistant/migrations/0005_message_content_segments.py
```diff
@@ -0,0 +1,52 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
yolanfery, line 4: This migration script should be simpler if we use abstraction instead of json but anyways can we have some test for it to ensure it is doing the proper thing ?

frontend/src/assistant/helpers/toolNames.ts
```diff
@@ -0,0 +1,65 @@
...
+    get_pipeline_template:          t("Reading pipeline template"),
+    create_pipeline_from_template:  t("Creating pipeline from template"),
+    list_static_webapps:            t("Listing web apps"),
+    create_static_webapp:           t("Creating web app"),
+    update_static_webapp:           t("Updating web app"),
+    get_help_or_doc:                t("Reading documentation"),
+    propose_pipeline_version:       t("Proposing pipeline version"),
+    execute_graphql:                t("Executing GraphQL query"),
+  };
+}
```
yolanfery, line 46: I am wondering if we shouldn't type (use enums) those tool names in GraphQL and use them here in the mapping to get translations ?

backend/hexa/assistant/tests/agents/test_base_agent_run.py
```diff
@@ -40,7 +43,9 @@ def test_run_saves_assistant_message(self):
             role=Message.Role.ASSISTANT
         )
         self.assertEqual(assistant_messages.count(), 1)
-        self.assertEqual(assistant_messages.first().content, "Hello!")
+        self.assertEqual(
+            assistant_messages.first().content, [{"type": "text", "content": "Hello!"}]
+        )
```
yolanfery, line 49: I feel like we could add a bit more testing on the backend side to cover the tool calls parsing and passing to the graphql layer

yolanfery (changes requested):
Great ! As discussed during the daily, it would be cool to have tool calls' args/results visible but we agreed to have a separate ticket/PR for this

Just comments/questions about : 

- Testing
- Typing the content instead of using a dynamic json

backend/hexa/assistant/agents/base.py
mrivar: Well spotted! I like that idea

frontend/src/assistant/helpers/toolNames.ts
mrivar: I'd vote for keeping it as it is. Whenever we use the tool for some reason in the frontend, we know straightaway which one it is referring to already, and it feels more programmatic, while this label would only be used here (if it's used anywhere else, we can have it as a separate helper anyway).
Also, it would add even more maintainance. I'd keep as it is, WDYT?

yolanfery: Ok we can keep it simple for now, no strong opinion at this point

backend/hexa/assistant/agents/base.py
```diff
@@ -145,7 +146,7 @@ async def run_stream(self, user_input: str) -> AsyncGenerator[str, None]:
         user_msg = await Message.objects.acreate(
             conversation=self.conversation,
             role=Message.Role.USER,
-            content=[{"type": "text", "content": user_input}],
+            content=[TextSegment(content=user_input).model_dump()],
```
yolanfery, line 159: Any reason to serialize the model into the database instead of keeping the abstractions in the database as well (new table with foreign key) ?

mrivar: I'm not sure if I follow, do you mean having the content as a list of message/tools in DB, or removing the content json entirely, and have the Message and Tools as rows in the DB?

However, I prefer the current approach to any of those tho. This content field is just for display purposes and makes the queries cheaper than having it all in DB objects (and it still encapsulates a single message anyway, even if it's divided in text and tool parts). Also, the more interesting tool info is still in a real DB table. This also adds flexibility, adding a third new message part (eg an image block) is a code change instead of a migration.

Maybe I'm missing something tho, let me know!

bramj (approved):
I did some testing (purely functional, not looking at the code) and this looks great!
Really nice improvement. When we add the details there, this will be fully polished 😁 

@mrivar nice job, okay to merge for me 👍

DimitriKwihangana (approved):
LGTM! gives more live to the agent! I think the other PR of showcasing the details tools.

PR #1839, opened 2026-05-28
fix: sentry issue no sample found for file xxx (HEXA-1672)

yolanfery:
The issue (https://bluesquareorg.sentry.io/issues/7398239096/?referrer=jira_integration) seems to come from two distinct cases : 

- Some files format are not supported
- When the sample is still queued for processing 

This PR is about removing warning for unsupported files (if we need to support a new format this might come as a feature request) and ensure we create a `PROCESSING` record in the database before actual execution to be handled correctly by the frontend

### DIFF
--- a/backend/hexa/datasets/schema/types.py
+++ b/backend/hexa/datasets/schema/types.py
@@ -13,6 +13,7 @@ from hexa.datasets.models import (
     DatasetVersion,
     DatasetVersionFile,
 )
+from hexa.datasets.queue import is_file_supported
 from hexa.files import storage
 from hexa.workspaces.models import Workspace
 from hexa.workspaces.schema.types import workspace_object, workspace_permissions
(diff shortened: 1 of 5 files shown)


backend/hexa/datasets/schema/types.py
```diff
@@ -237,7 +238,10 @@ def resolve_version_file_metadata(obj: DatasetVersionFile, info, **kwargs):
     try:
         return obj.sample_entry
     except DatasetFileSample.DoesNotExist:
-        logging.error(f"No sample found for file {obj.filename} with id {obj.id}")
+        if is_file_supported(obj.filename):
```
DimitriKwihangana, line 241: I like the change. reporting only what we support!

DimitriKwihangana (approved):
Perfect! I like the unit tests. It not something I would think of for this :). thanks for the learning @yolanfery

backend/hexa/datasets/schema/types.py
```diff
@@ -237,7 +238,10 @@ def resolve_version_file_metadata(obj: DatasetVersionFile, info, **kwargs):
...
+            logging.warning(
+                f"No sample found for supported file {obj.filename} with id {obj.id}"
+            )
```
mrivar, line 244: Why do we want to log only supported files and not unsupported files? Not questioning, just wondering

mrivar (approved):
LGTM 🚀! Just a small question for knowledge more than anything

backend/hexa/datasets/schema/types.py
yolanfery: We expect the sample to not exist for unsupported file formats

But for supported file format we should have a sample

PR #1840, opened 2026-05-29
fix: Clarify the cron schedule to avoid user confusing

bramj:
Based on some direct support requests from the DS team:

- Cron schedules are entered in UTC
- Then runs are displayed in local time

This confuses users. To help against that, add a little description to the schedule that clarifies that the time is in UTC + show the "next run" in localized time.

### Display (FR/EN)

Screenshot 2026-05-29 at 09 21 18
Screenshot 2026-05-29 at 09 23 42

### Edit (FR/EN)

Screenshot 2026-05-29 at 09 21 57
Screenshot 2026-05-29 at 09 23 25

### DIFF
--- a/frontend/public/locales/en/messages.json
+++ b/frontend/public/locales/en/messages.json
@@ -490,6 +490,7 @@
   "New Template '{{name}}' created successfully.": "New Template '{{name}}' created successfully.",
   "New Template Version for '{{templateName}}' created successfully.": "New Template Version for '{{templateName}}' created successfully.",
   "Next": "Next",
+  "Next run: {{date}} ({{timeZone}})": "Next run: {{date}} ({{timeZone}})",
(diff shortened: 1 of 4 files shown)


bramj:
@mcrimi Assigned you as well for feedback on the UX proposal, thanks!

mrivar (approved):
LGTM! Just as a comment, I guess we don't want to show the Schedule result in locale as well?

bramj:
> LGTM! Just as a comment, I guess we don't want to show the Schedule result in locale as well?

You mean the part where I added "(UTC)"? That was the alternative option actually. But I figured it would be a bit confusing to put a certain hour, and then see a different hour next to it. So that's why I went for the "next run" visualization.

mcrimi:
Lgtm!

PR #1841, opened 2026-06-01
fix:error on hitting maximum tokens is not well handled on ai agent (Hexa-1635) 

DimitriKwihangana:
When the AI agent hits its token output limit, users were seeing a generic "An error occurred" with no explanation. This PR replaces that with specific, actionable messages for every token-related failure mode — and adds a loop guard to prevent the agent from spinning indefinitely.

## Changes

- Loop guard: Added UsageLimits(request_limit=10) — stops the agent after 10 model calls. Without this, hitting max_tokens with end_strategy="exhaustive" causes infinite retries.

- **4 specific** error handlers replacing the old generic catch:
Request limit hit (loop) → "I got stuck in a loop..."
Token output limit hit → "I hit the maximum token limit..."
IncompleteToolCall (defensive, see below) → "I hit the maximum token limit..."
Other unexpected model behavior → "An error occurred"
- **finally** block to cancel the naming background task on errors.

-  Removed max_tokens removed duplicate'

#### Note on[ IncompleteToolCall](https://pydantic.dev/docs/ai/api/pydantic-ai/exceptions/#pydantic_ai.exceptions.IncompleteToolCall)
pydantic-ai raises IncompleteToolCall when the model stops mid-tool-call due to token limits. However, with end_strategy="exhaustive" (our current setup), pydantic-ai retries instead of raising this exception so the loop guard fires first in practice. The handler is included as a defensive fallback and is covered by unit tests.


backend/hexa/assistant/agents/base.py
```diff
@@ -192,14 +200,46 @@ async def run_stream(self, user_input: str) -> AsyncGenerator[str, None]:
                 ),
             )
 
-        except json.JSONDecodeError:
```
mrivar, line 195: I wonder if we want to keep this one for visibility

backend/hexa/assistant/agents/base.py
```diff
@@ -192,14 +200,46 @@ async def run_stream(self, user_input: str) -> AsyncGenerator[str, None]:
...
             )
+        except UnexpectedModelBehavior:
+            logger.exception("agent.run_stream: unexpected model behavior")
             yield format_sse("error", ErrorPayload(message="An error occurred"))
         except Exception:
             logger.exception("agent.run_stream: error during streaming")
             yield format_sse("error", ErrorPayload(message="An error occurred"))
+        finally:
+            if naming_task is not None and not naming_task.done():
+                naming_task.cancel()
```
mrivar, line 239: I guess this is quite unlikely due to the naming task being "small", but just wondering, why do we cancel it? just wondering, not saying it's wrong

backend/hexa/assistant/tests/agents/_helpers.py
```diff
@@ -73,6 +73,23 @@ def make_built_model(test_model) -> BuiltModel:
...
+        stream_calls.append(1)
+        yield {
+            0: DeltaToolCall(
+                name=tool_name,
+                json_args='{"arg": "trun',  # deliberately truncated
+                tool_call_id="call-truncated-001",
+            )
+        }
+
+    return FunctionModel(stream_function=stream_func)
```
mrivar, line 90: nice one!

mrivar (commented):
Looks nice! just a couple of questions 😁

frontend/src/assistant/features/ChatPane/index.tsx
```diff
@@ -174,9 +159,10 @@ export default function ChatPane({
       if (name) onConversationNameChange?.(name);
       markDone();
     },
-    error: () => {
+    error: (data) => {
       clear();
-      setSendError(t("The AI service encountered an error. Please try again."));
+      const { message } = (data ?? {}) as { message?: string };
+      setSendError(message ?? t("The AI service encountered an error. Please try again."));
```
yolanfery: This means the message will not be translated

We have a pattern in the code base to define enums in graphql for the errors and then provide a (translated) explanation in the frontend for each enum. We might want to follow this

yolanfery (changes requested):
Cool tests :) 

Just an additional comment about error handling (translation)

backend/hexa/assistant/sse_types.py
```diff
@@ -32,5 +32,12 @@ class DonePayload(TypedDict):
     name: str | None
 
 
+class ErrorCode(str):
```
yolanfery: [Minor] I think `Enum` is more appropriate than `str`

frontend/src/workspaces/features/CreatePipelineDialog/CreatePipelineUsingAI/useAIForm.ts
```diff
@@ -114,9 +114,20 @@ export function useAIForm(
...
+      const errorMessages: Record<string, string> = {
+        AGENT_STUCK_IN_LOOP: t(
+          "I got stuck in a loop — try breaking your request into smaller steps.",
+        ),
+        MAX_TOKENS_REACHED: t(
+          "I hit the maximum token limit — try breaking your request into smaller steps.",
+        ),
+        UNEXPECTED_MODEL_BEHAVIOR: t("An unexpected error occurred. Please try again."),
+        UNKNOWN_ERROR: t("An error occurred. Please try again."),
+      };
```
yolanfery: [Minor] This mapping is duplicated, we can extract it into a helper file and function

yolanfery (approved):

PR #1842, opened 2026-06-01
feat: Add WHO CIAM login (HEXA-1606)

mrivar:
Add WHO CIAM as an identity provider for single sign on (SSO).
This is a requirement for the local hosting of OpenHEXA on WHO's servers. Users need to be able to use their WHO accounts to log in directly on OpenHEXA.

**Note:** Related PR: https://github.com/BLSQ/iaso/pull/3024/

## Changes

- Added `django-allauth[socialaccount]` library to manage OIDC/SSOs
- `sso_adapter`: manages any type of sso
- Added env variables to add any OIDC provider (WHO added by default)
- Added frontend providers logins
- Hide password login if providers are given

## How/what to test

- Log out
- Go to login page
- Select WHO login

## Screenshots / screencast

image
image


backend/hexa/user_management/schema/mutations.py
```diff
@@ -144,6 +144,9 @@ def resolve_login(_, info, **kwargs):
     request: HttpRequest = info.context["request"]
     mutation_input = kwargs["input"]
 
+    if not settings.PASSWORD_LOGIN_ENABLED:
+        return {"success": False, "errors": ["INVALID_CREDENTIALS"]}
```
bramj, line 148: 👍

backend/config/settings/base.py
```diff
@@ -337,6 +393,45 @@
...
+LOGIN_REDIRECT_URL = "/"
+
+# Required by django.contrib.sites (used by allauth)
+SITE_ID = 1
+
+# allauth configuration
+# https://docs.allauth.org/en/latest/socialaccount/configuration.html
+ACCOUNT_USER_MODEL_USERNAME_FIELD = None
+ACCOUNT_LOGIN_METHODS = {"email"}
+ACCOUNT_SIGNUP_FIELDS = ["email*", "password1*", "password2*"]
```
bramj, line 412: What does this mean?

backend/config/settings/base.py
```diff
@@ -337,6 +393,45 @@
 # Auth settings
 LOGIN_URL = "core:login"
 LOGOUT_REDIRECT_URL = "core:login"
+LOGIN_REDIRECT_URL = "/"
+
+# Required by django.contrib.sites (used by allauth)
+SITE_ID = 1
```
bramj, line 406: We should not forget to update the `example.org` domain in the Django admin.

backend/hexa/user_management/sso/sso_adapter.py
```diff
@@ -0,0 +1,110 @@
...
+        sociallogin.save(request)
+
+        try:
+            self._send_new_account_email(user, sociallogin.account.provider)
+        except Exception:
+            logger.exception(
+                "Failed to send SSO new-account notification for %s (provider: %s)",
+                user.email,
+                sociallogin.account.provider,
+            )
```
bramj, line 99: Good call to catch those errors and not break the account creation 👍

backend/hexa/user_management/migrations/0041_ensure_default_site.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def ensure_default_site(apps, schema_editor):
+    # AccountMiddleware (allauth) does a Site lookup on every request using SITE_ID=1.
+    # django.contrib.sites normally creates this row via post_migrate, but pg-dump
+    # restores and some CI setups skip that signal.  This migration guarantees the row
+    # exists regardless of how the database was provisioned.
+    Site = apps.get_model("sites", "Site")
+    Site.objects.get_or_create(
+        id=1, defaults={"domain": "example.com", "name": "example.com"}
```
bramj: I wonder if we can get the correct domain from `settings.BASE_URL` here if that's set?

backend/hexa/core/schema.py
```diff
@@ -31,4 +32,23 @@ def resolve_config_allow_self_registration(_, info):
     return settings.ALLOW_SELF_REGISTRATION
 
 
+@config_object.field("passwordLoginEnabled")
+def resolve_config_password_login_enabled(_, info):
+    return settings.PASSWORD_LOGIN_ENABLED
+
+
+@config_object.field("oidcProviders")
+def resolve_config_oidc_providers(_, info):
```
bramj, line 41: We should have some tests for this endpoint since there's quite a bit going on when you call `settings.OIDC_PROVIDERS`

backend/config/settings/base.py
```diff
@@ -273,6 +323,11 @@
     "django_otp",
     "django_otp.plugins.otp_static",
     "django_otp.plugins.otp_email",
+    # allauth (always installed for clean migrations; OIDC providers enabled via WHO_CIAM_ENABLED)
```
bramj: > OIDC providers enabled via WHO_CIAM_ENABLED

This seems outdated

backend/hexa/user_management/sso_adapter.py
```diff
@@ -0,0 +1,110 @@
...
+                "next"
+            ] = f"{settings.NEW_FRONTEND_DOMAIN.rstrip('/')}{next_url}"
+
+        if sociallogin.is_existing:
+            return
+
+        # Only link/create accounts when the provider has verified the email.
+        # Defaults to True when the claim is absent (Entra ID verifies at IdP level
+        # and often omits the field entirely).
+        if not sociallogin.account.extra_data.get("email_verified", True):
```
bramj: I don't think we'll ever add a provider that's less trustworthy, so I think it's okay to default to `True`

backend/hexa/user_management/sso/sso_adapter.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def is_auto_signup_allowed(
+        self, request: HttpRequest, sociallogin: SocialLogin
+    ) -> bool:
+        data = sociallogin.account.extra_data
+        email = (data.get("email") or "").strip()
+        email_verified = data.get("email_verified", True)
+        return bool(email and email_verified)
+
+    def save_user(
```
bramj, line 70: Does `allauth` wrap this in a DB transaction? Just checking to make sure we don't end up with users without `sociallogin` attached for example.

backend/hexa/user_management/sso/sso_adapter.py
```diff
@@ -0,0 +1,110 @@
...
+        send_mail(
+            title=f"New OpenHEXA account created via {display_name}",
+            recipient_list=recipients,
+            template_name="user_management/mails/sso_new_account",
+            template_variables={
+                "new_user": user,
+                "provider_name": display_name,
+            },
+            attachments=get_email_attachments(),
+        )
```
bramj, line 124: I would like a bit of documentation on the flow here, WDYT?

E.g. from Claude, but maybe less verbose 😅 :
```
User clicks "Login with WHO/Entra"
          │
          ▼
    [ Redirect to IdP, user authenticates, IdP redirects back ]
          │
          ▼
    allauth receives the callback with OIDC claims
          │
          ▼
  ┌───────────────────────────────────────────────┐
  │  pre_social_login()          ← ALWAYS runs     │
  │  • fix up the "next" redirect URL              │
  │  • if account already linked → return          │
  │  • else: find local User by email → connect()  │
  └───────────────────────────────────────────────┘
          │
          ▼
    Is this login already linked to a user?
          │
     ┌────┴─────────────────────────┐
     │ YES                          │ NO
     │ (existing, or just           │ (brand-new identity,
     │  connected above)            │  no matching email)
     ▼                              ▼
   LOGIN the user            ┌──────────────────────────────┐
   (no more adapter          │ is_auto_signup_allowed()     │
    methods called)          │ • email present + verified?  │
                             └──────────────────────────────┘
                                     │
                            ┌────────┴────────┐
                            │ True            │ False
                            ▼                 ▼
                    ┌─────────────────┐   show signup form
                    │  save_user()    │   (manual flow)
                    │  • create User  │
                    │  • set unusable │
                    │    password     │
                    │  • save login   │
                    │  • call ↓       │
                    │ _send_new_      │
                    │  account_email()│
                    └─────────────────┘
                            │
                            ▼
                       LOGIN the user
```

bramj (changes requested):
@mrivar This all looks very solid, and generalized to support multiple providers, nice job! 👏 

I took the liberty to add a commit with a mock server to test the OpenID Connect callback. It did reveal the main issue with the flow: it should be invite-only.

Right now when I invite user A, I see:
1. User A receives invitation email and clicks the link
2. User A lands on a page that allows them to set up a password

Instead, they should land on the "Login with WHO" page and so if there's an invitation + validation from WHO -> create the user.

If I as a user click on "Login with WHO" and there's no trace of me in the system (so no existing user or invitation), I should be presented an error message.

Let me know what you think and we can discuss in person if anything's not clear!

backend/config/settings/base.py
mrivar: - `ACCOUNT_USER_MODEL_USERNAME_FIELD = None`: by default allauth stores a User with a username field, but we don't have that field in OpenHexa
- `ACCOUNT_LOGIN_METHODS = {"email"}`: Similar to the one above. Tells allauth that when users log in with a password (the classic login form), they identify themselves by email, not username. 
- `ACCOUNT_SIGNUP_FIELDS`: Again, related with avoiding allauth to use username
- `SOCIALACCOUNT_LOGIN_ON_GET`: We use get endpoints for login, this just allows that

The rest of the configuration is more "usual" (avoid email confirmation, automatically create account on provider signup)

backend/hexa/user_management/migrations/0041_ensure_default_site.py
mrivar: good call, that makes more sense 👌

backend/config/settings/base.py
mrivar: what do you mean?

backend/hexa/user_management/sso/sso_adapter.py
mrivar: love it, I'll add a simpler version 😁

backend/hexa/user_management/sso/sso_adapter.py
mrivar: Good call, will check that

backend/hexa/user_management/sso/sso_adapter.py
```diff
@@ -68,12 +81,13 @@ def save_user(
         first_name = (data.get("given_name") or data.get("first_name") or "").strip()
         last_name = (data.get("family_name") or data.get("last_name") or "").strip()
 
-        user = User(email=email, first_name=first_name, last_name=last_name)
-        user.set_unusable_password()
-        user.save()
+        with transaction.atomic():
```
bramj, line 84: Great 👍

docs/en/sso.md
```diff
@@ -0,0 +1,184 @@
...
+    IDP     --> AUTH
+    AUTH    -->|"3 · redirect to /callback"| OH2
+    OH2     --> OH3
+    OH3     <-.->|"token endpoint"| IDP
+    OH3     --> DEC
+    DEC     -->|"yes"| LINK
+    DEC     -->|"no"| CREATE
+    LINK    --> DONE
+    CREATE  --> DONE
+```
```
bramj, line 48: Love the mermaid diagram! ❤️

bramj (approved):
This is looking great! 👏 Good job @mrivar !!

As discussed we'll do auto-accepting of invitation in a separate PR.

PR #1843, opened 2026-06-01
feat: ai can make mistakes disclaimer (HEXA-1675)

DimitriKwihangana:
Added a small paragraph that warns that AI can make mistakes

### DIFF
--- a/frontend/public/locales/en/messages.json
+++ b/frontend/public/locales/en/messages.json
@@ -71,6 +71,7 @@
   "After regenerating the password, you will have to:": "After regenerating the password, you will have to:",
   "AI Agents": "AI Agents",
   "AI Assistant": "AI Assistant",
+  "AI can make mistakes. Always verify important information.": "AI can make mistakes. Always verify important information.",
   "AI features are not enabled": "AI features are not enabled",
   "AI feedback": "AI feedback",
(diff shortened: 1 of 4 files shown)


mrivar (approved):
Looks good! More of a question for @mcrimi : do we want this as well in the Create Pipeline with AI modal, or only in the chat panes is enough?

EDIT: Actually, I'm wondering if we could have it inside the textarea, so it doesn't decrease the height of the chat pane (next to the "Send" button)
image

DimitriKwihangana:
> Looks good! More of a question for @mcrimi : do we want this as well in the Create Pipeline with AI modal, or only in the chat panes is enough?
> 
> EDIT: Actually, I'm wondering if we could have it inside the textarea, so it doesn't decrease the height of the chat pane (next to the "Send" button)

Thanks @mrivar . I agree the disclaimer should be on creating pipeline as well

mcrimi:
Nice! Yes @mrivar good point, it should be in the create modal as well.  I feel like having the disclamer in the text area would be a bit too invasive.

mcrimi (changes requested):
Looks great. We only need to add the disclamer in the "Create pipelines with AI" modal and we're good to go.

PR #1844, opened 2026-06-03
chore:remove hidden assistant (HEXA-1652)

DimitriKwihangana:
Deleted the /assistant page


mrivar (approved):

bramj (changes requested):
Good! But I think there's a GraphQL query that's no longer used and that can be removed:
`AssistantPage` in `frontend/src/assistant/graphql/queries.graphql`.
Can you double-check?

bramj (approved):
LGTM! 👍

PR #1845, opened 2026-06-04
feat: add GCP vertex integration for AI assistant (HEXA-1613)

yolanfery:
Make it possible to use Claude through Vertex  

Screenshot 2026-06-04 at 11 01 03

Related to infra change : https://github.com/BLSQ/openhexa-infra/pull/400


backend/hexa/user_management/models.py
```diff
@@ -228,9 +236,21 @@ def model_choices() -> list[dict[str, str]]:
         ]
 
     def validate(self):
-        if self.enabled and not (self.provider and self.model and self.api_key):
+        if not self.enabled:
+            return
+        if not (self.provider and self.model):
+            raise ValidationError("Incomplete config")
+        if self.provider != AiSettings.Provider.ANTHROPIC_VERTEX and not self.api_key:
```
mrivar, line 243: It would be nice to have a logical separation for providers that don't need an API key. Just from the top of my mind I'd say having some sort of logical class with info like this one that then transform into text choices for the DB model? WDYT?

bramj (changes requested):
Sorry for the late review on this one, but I had to think it through a bit.
The code is solid for this, but I feel like functionally, we need to go back to the drawing board.

The way I see it, an OpenHEXA instance should have one of two "modes", which can be set by an ENV var like `ASSISTANT_MODE=managed|byok`:

1. **Managed**: For the end user, the technical side is fully abstracted, they don't need to worry about anything except for hitting their limits.
    - An organization has a monthly "credit" limit, based on their plan. Additionally, a per-user limitation (as done in this PR) is a great idea.
    - Monthly limits are expressed as "credits", the actual cost is our problem.
    - We provide the agents as a service, and we choose the models that live behind it: This allows us to use different models for different use cases, and swap models as we see fit to balance performance with price.
    - Users have no "AI settings", just an indication of their credit consumption.
    - It is Vertex AI at the moment, but it can change in the future, so I'd stick to the abstraction of "managed"
2. **BYOK**: The feature as we know it, everyone is free to add their own key if they have one. (At some point we should probably expand her to allow other providers, OpenRouter, etc.)

To discuss:
- All of the above :)
- Credits: how do we define this? I suppose token use makes more sense.
- For the models used in the "managed" setup: I think we should have transparency, but maybe more like a dedicated page where we go into more details about which models are used and the rationale, instead of a select in the user profile.
- Currently for our SaaS: I know we said we'd open up only for BLSQ, but this makes me feel like we should just open up with low limits to avoid complicated logic for this.

yolanfery:
Closing in favor of https://github.com/BLSQ/openhexa-app/pull/1888

PR #1847, opened 2026-06-05
fix: silence webapp GraphQL errors to warning in Sentry

DimitriKwihangana:
The aim is to tag Graphql errors that comes from web apps and then ignore them before send.

### DIFF
--- a/backend/config/test_sentry.py
+++ b/backend/config/test_sentry.py
@@ -39,6 +39,13 @@ class SentryBeforeSendTest(TestCase):
         event = {"tags": {"connection_test": True}}
         self.assertIsNone(self.before_send(event, self._make_hint(exc)))
 
+    def test_downgrades_webapp_graphql_errors_to_warning(self):
+        exc = Exception("Unknown type 'PipelineRunOutputFile'")
+        event = {"tags": {"webapp_graphql": True}, "level": "error"}
+        result = self.before_send(event, self._make_hint(exc))
+        self.assertIsNotNone(result)
+        self.assertEqual(result["level"], "warning")
+
     def test_circular_exception_chain_does_not_loop(self):
         exc_a = ValueError("a")
         exc_b = ValueError("b")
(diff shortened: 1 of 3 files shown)


backend/config/test_sentry.py
```diff
@@ -39,6 +39,13 @@ def test_filters_connection_test_errors(self):
         event = {"tags": {"connection_test": True}}
         self.assertIsNone(self.before_send(event, self._make_hint(exc)))
 
+    def test_downgrades_webapp_graphql_errors_to_warning(self):
+        exc = Exception("Unknown type 'PipelineRunOutputFile'")
+        event = {"tags": {"webapp_graphql": True}, "level": "error"}
+        result = self.before_send(event, self._make_hint(exc))
+        self.assertIsNotNone(result)
+        self.assertEqual(result["level"], "warning")
```
yolanfery: Are we sure we want still get warnings ? I thought we would just ignore them 

Let's confirm with @bramj

backend/hexa/webapps/graphql_proxy.py
```diff
@@ -150,4 +151,6 @@ def handle_graphql_proxy(request: HttpRequest, webapp: Webapp):
         )
 
     track(request, "webapp_graphql_query", {**event_properties, "status": "allowed"})
-    return _graphql_view(request)
+    with sentry_sdk.new_scope() as scope:
+        scope.set_tag("webapp_graphql", True)
```
yolanfery, line 155: Can we use a shared constant across the implementation to avoid Magic strings ? (https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings)

yolanfery (changes requested):

backend/config/test_sentry.py
bramj: @DimitriKwihangana @yolanfery Indeed let's just ignore them

backend/hexa/webapps/graphql_proxy.py
yolanfery: (Not needed anymore if we simply ignore the errors)

bramj: Agreed, but since we'll be certainly refactoring this when we add user-managed Sentry notifications, let's go with this for now.

bramj (approved):
LGTM! I'll merge this to add to the release.

PR #1848, opened 2026-06-05
fix: MCP create pipeline version only takes a single file (HEXA-1648)

yolanfery:
The MCP was only able to create pipelines with one file. 

Allow creating using a dict of files. Implementation in the GraphQL layer fro keeping the MCP layer as thin as possible

### DIFF
--- a/backend/hexa/mcp/tools/pipelines.py
+++ b/backend/hexa/mcp/tools/pipelines.py
@@ -1,7 +1,4 @@
-import base64
-import io
 import json
-from zipfile import ZipFile
 
 from hexa.mcp.protocol import tool
 from hexa.pipelines.models import PipelineFunctionalType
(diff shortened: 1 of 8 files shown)


backend/hexa/mcp/tools/pipelines.py
```diff
@@ -243,28 +241,38 @@ def create_pipeline(
...
+    r"""Create a new pipeline in the current workspace and upload its files as the first version (v1).
 
     Always provide a meaningful description summarizing what the pipeline does.
     If the pipeline has no clear purpose or is blank, use "" as the description.
-    The source_code must follow the OpenHEXA SDK structure (use @pipeline and @task decorators).
+
+    Provide files_json as a JSON array of {path, content, encoding?} objects, e.g.
+    '[{"path": "pipeline.py", "content": "from openhexa.sdk import pipeline\n..."},
+      {"path": "helpers.py", "content": "def clean(df): ..."},
+      {"path": "requirements.txt", "content": "pandas\nrequests"}]'.
```
mrivar, line 256: We could add an example with encoding

mrivar (approved):
Really cool feature 😁

PR #1849, opened 2026-06-05
fix: stringify object values in file sample for rendering (HEXA-1676)

yolanfery:
When the data is a dict (nested structure), it is not rendering correctly. We stringify it

## How to test

- Generate a dataset with data such as `{"id":"abc-123"}` in some cells
- Go to preview
- No more console error should appear


mrivar (approved):
looks good to me!

PR #1853, opened 2026-06-08
fix: Small Django admin improvements

bramj:
The pipeline details page actually timed out due to the N+1 queries to construct the dropdowns for pipeline versions. (read-only prints just the resource, removing the dropdown altogether)

Changes:
- Performance improvements on pipelines, pipeline versions and pipeline runs pages
- More date filtering on pipeline runs, remove the pipeline filter (unusable anyways)
- Show if user is active

**Before:**
image

**After:**
image


DimitriKwihangana (approved):
Looks good! I was wondering if the pipeline templates would need the same treatment

bramj:
> I was wondering if the pipeline templates would need the same treatment

Possibly! Could you add those if needed? You can push on this same PR :)

bramj:
@DimitriKwihangana thanks for the improvements! 🙏

PR #1856, opened 2026-06-11
fix:remove autos-scroll down when assistant is responding

DimitriKwihangana:
Fix auto-scroll in the AI assistant chat that was forcing users back to the bottom even when they tried to scroll up during streaming. Added a wheel listener to unpin auto scroll on an upward scroll


mrivar (approved):
Works great!
Note: I had to merge main into it because of some db mismatch 😁

bramj (approved):

PR #1858, opened 2026-06-11
chore: Speed up Django test suite

bramj:
I noticed that the backend tests **grew from ~8 mins to ~12 mins over the last 12 months**.
Since I found myself often waiting for tests to finish running, I decided to have a go at it with Claude and ended up with a **new test duration of ~6 mins**. :tada: :rocket: 

### Changes

- Change pw hasher to speed up tests (interesting, this is a big one!)
- Make sure Logfire is disabled in tests
- Run tests in parallel
- Speed up coverage calculation
- Speed up Postgres for dev/CI by disabling crash-durability features

image

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -47,31 +47,3 @@ ignore = [
 [tool.ruff.lint.pydocstyle]
 convention = "numpy" # Accepts: "google", "numpy", or "pep257".
 
-[tool.coverage.report]
-exclude_lines = [
-    # Have to re-enable the standard pragma
-    "pragma: no cover",
-
-    # Don't complain about missing debug-only code:
-    "def __repr__",
-    "if self\\.debug",
-
-    # Don't complain if tests don't hit defensive assertion code:
-    "raise AssertionError",
-    "raise NotImplementedError",
-
-    # Don't complain if non-runnable code isn't run:
-    "if 0:",
-    "if __name__ == .__main__.:",
-]
-
-omit = [
-    "*/__init__.py",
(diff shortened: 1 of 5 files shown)


backend/pyproject.toml
```diff
@@ -47,31 +47,3 @@ ignore = [
...
-omit = [
-    "*/__init__.py",
-    "*/migrations/*",
-    "*/tests/*",
-    "config/*",
-    "*/admin.py",
-    "*/apps.py",
-    "*/urls.py",
-    "manage.py",
-]
```
bramj, line 77: This file is no longer used, `.coveragerc` takes precedence

mrivar: just wondering, we don't use this config but does it have anything we want to recycle into the `.coveragerc` file, or is it all useless/not neeeded for our `.coveragerc` config?

mrivar (approved):
This is so cool!! Going from 12 mins to 6 mins is nuts 😄

backend/pyproject.toml
bramj: Honest answer: I don't know. Maybe the coverage was used some time in the past, and this was the config that was convenient.

I think there's some value in looking into this again, but only if it results in something useful like a CI test that checks if the coverage hasn't gone down. ([something like this](https://docs.openedx.org/en/latest/developers/how-tos/use-python-coverage-comment.html) I guess).

But so I'd argue that we're not using it right now, so no need to dig deeper, but feel free to look into it and see if we can make it useful.

PR #1859, opened 2026-06-11
Auto accept all pending invitations upon user registration (HEXA-1696)

bramj:
This PR adds a helper to accept all pending invitations and uses it whenever a new user is created.

### Changes

It is used whenever a new user is created. More specifically, this means:
- when a user registers via user/pass
- when a user is created via SSO login

### How to test

- CIAM account creation after a workspace invitation should arrive directly on the workspace
- Regular signup with or without invitation should be unaffected


mrivar (approved):
Looks good! Exciting to see the OICD/SSO working!

PR #1860, opened 2026-06-11
feat: Conditional parameters (HEXA-1687)

mrivar:
## Related PRs
- https://github.com/BLSQ/openhexa-sdk-python/pull/396
- https://github.com/BLSQ/openhexa-app/pull/1860

NOTE: Waiting for new SDK version to merge this one, will stay in Draft until then (need to add new sdk version in requirements first). However, it can be reviewed already. 

Add parameters conditional to others

## Changes

- Added GraphQL schemas with updated SDK param values (disables/disable_when)
- In frontend, disable values when a param sets it to be disabled
- In backend, check if a pipeline is schedulable or if a new config is valid taking into account disabled values
- [In SDK, check if a pipeline run is valid taking into account disabled values](https://github.com/BLSQ/openhexa-sdk-python/pull/396)
- Added documentation

## How/what to test

You can test it if you have the branch version of the sdk in local using a `docker-compose.override.yaml`:
```yaml
services:
  app:
    volumes:
      - ../openhexa-sdk-python:/openhexa-sdk-python
    entrypoint: ["/bin/bash", "-c", "pip install --user -q -e /openhexa-sdk-python && exec /code/docker-entrypoint.sh \"$@\"", "--"]
```
Also, run the following command from the pip you're using to run the openhexa-app:
```shell
pipx install --force --editable ~/dev/bluesquare/openhexa-sdk-python
```

- Create different params
- Add param to disable
- Test disabling it
- Required params that are disabled should not be required anymore (avoiding having to set dummy values)
- Try hacking the frontend (Inspect -> remove disable). When running the pipeline, the set value for that param should be lost.

Example of disabling params:
```python
@parameter(
    "disabling_param",
    name="Disabler",
    help="",
    type=bool,
    required=False,
    disables=["int_val"]
)
@parameter(
    "int_val",
    name="Integer Value",
    help="Int.",
    choices=[0, 1, 2, 3],
    type=int,
    required=True,
)
```

## Screenshots / screencast

image

image


frontend/src/workspaces/features/RunPipelineDialog/parameterFixtures.ts
```diff
@@ -0,0 +1,26 @@
...
+  help: null,
+  choices: null,
+  widget: null,
+  connection: null,
+  directory: null,
+  disables: [],
+  disableWhen: true,
+  choicesFromFile: null,
+  ...overrides,
+});
```
bramj, line 26: Nice improvement! 👍

bramj (changes requested):
Nice work! This is looking really polished, and thorough with the addition of `disable_when` and lot's of tests and documentation 👍 

The code looks solid to me, nothing really to remark there.

I just have a few issues with edge cases when testing locally:

- When I use `disable_when=False`, this doesn't really seem to work. I'd expect the condition to invert with regards to the default.
- Default parameter setting: the disabled conditions aren't enforced in the modal it seems (see screenshot)

Image

but when running the pipeline, it works correctly:

Image

 Can you double check these?

mrivar:
@bramj
The default setting not disabling is intentional. I feel like users should be able to change values even if the value is disabled, as it is just a fixed config. We could show in a way that those values would be disabled in a "default run" but letting the user edit them. Or do you think we should fully disable them to avoid confusion? WDYT?

On the `disable_when=False`, you're right, there seems to be a bug, I'll look into it.

EDIT:
- Added an explanatory text in default param setting, but if we want to fully disable the values lmk.
- For the bug, after checking it might be an issue of the code not picking up the new sdk version, try doing `pipx install --force --editable ~/dev/bluesquare/openhexa-sdk-python` on the pip you're using to run it (it might be `.venv/bin/pip`)

bramj (approved):
@mrivar 

> The default setting not disabling is intentional. I feel like users should be able to change values even if the value is disabled, as it is just a fixed config

Yes I actually thought about this a bit later, my initial reaction was of surprise, but it makes sense to not do enforce anything for the default values.
Thanks for adding the clarifying sentence, that's great 👍 

LGTM! 👏

PR #1862, opened 2026-06-12
fix:allow re-publishing pipeline after deletion

DimitriKwihangana:
When a pipeline template is soft-deleted, its PipelineTemplateVersion rows are not removed. The createTemplateVersion permission check was seeing the stale relation and blocking the button permanently.

Fixed by skipping the relation when the linked template's deleted_at is set, so the permission is correctly reenabled after deletion.


mrivar (approved):
Nice one! This "allowed" option has so many edge cases, it feels like a tricky one, took me a minute to understand the whole logic hahah

bramj (approved):
Great! Tested, working well 👍

PR #1871, opened 2026-06-16
fix: sentry issue smtp authentication failed (HEXA-1673)

yolanfery:
This PR does not solve the underlying [SMTP issue](https://bluesquareorg.sentry.io/issues/7505130121/?referrer=jira_integration) (it might be a transient issue) but ensures the pipeline run does not fail because of notification

Open to opinions on this approach


DimitriKwihangana (approved):
I like the implementation!!

mrivar (commented):
Way better than what we had, having a pipeline run failing for a mail error feels awkward.
But I wonder if we want to additionally have a retry/backoff for these errors to make sure we don't lose the mail sending (Although we could just open a separate task for that if we want this merged asap). WDYT?

EDIT: Interesting feedback from Claude for long-term development, although I'm not sure if it's worth tbh:
> notifications should be out-of-band. The deeper architectural fix is to send pipeline notifications via the async task/queue layer rather than inline in the runner process. That isolates failures and gives you retries for free. Bigger scope — fine as a follow-up, not this PR.

backend/hexa/core/utils.py
```diff
@@ -55,4 +59,10 @@ def send_mail(
             else:
                 mail.attach(filename, content, mimetype)
 
-    return mail.send()
+    try:
+        return mail.send()
+    except Exception as e:
+        if not fail_without_raising:
+            raise
+        logger.error("Failed to send mail '%s' to %s: %s", title, recipient_list, e)
```
mrivar: minor, but `logger.exception` might be more suitable

yolanfery: Changed, thanks

yolanfery:
> Way better than what we had, having a pipeline run failing for a mail error feels awkward. But I wonder if we want to additionally have a retry/backoff for these errors to make sure we don't lose the mail sending (Although we could just open a separate task for that if we want this merged asap). WDYT?
> 
> EDIT: Interesting feedback from Claude for long-term development, although I'm not sure if it's worth tbh:
> 
> > notifications should be out-of-band. The deeper architectural fix is to send pipeline notifications via the async task/queue layer rather than inline in the runner process. That isolates failures and gives you retries for free. Bigger scope — fine as a follow-up, not this PR.

Yep I agree with this and explored a bit.

This would need a change in the infra to have a new pod for the async worker. I don't think it is worth our time and the additional infra cost (+ change in localhosting instances) at this point since this is happened only once in months.
I also explored saving the failed email so that we can manually resend it from the admin panel. I will go that route. It seems a great middle ground

mrivar (approved):
Love the idea of being able to resend the email from the admin console! That along with the sentry exception whenever there's a failed email sounds good to me 😁

PR #1872, opened 2026-06-16
feat:  agent tools details (HEXA-1682)

mrivar:
Show the agent tool details, both inputs and outputs.

## Changes

All frontend changes, as we already send the tool inputs/outputs via graphQL and SSE.
- Add dropdown to open tool to see details
- Add option to se details in a modal to go deeper into the information and seeing it in a bigger screen (chat is too cramped)
- Show JSON by default (JsonView)
- On specific tools have special viewers (File Trees, Markdown, Tables, etc) that fallback to JSON if broken
- Option to see the raw/JSON information always
- On the special case of `propose_pipeline_version` tool, show only the output (input & output are the same, but output includes all files; details can be viewed better in the diff anyway)

## How/what to test

- Start a conversation and ask the agent to explore the different tools in the app.
- Check the details
- Try to break it

## Screenshots / screencast

image
image

image
image

image

image


DimitriKwihangana (approved):
The intergration on the frontend looks very great! Just a quick question why do we need to visualise the raw content/json?

frontend/src/assistant/helpers/tools.ts
```diff
@@ -0,0 +1,36 @@
+// Canonical roster of assistant tool machine names — the single source of truth
+// for tool identities. Every tool-keyed table (labels in `toolNames`, per-tool
+// config in `toolConfig`, renderer overrides) keys off these constants instead
+// of repeating the strings, so a name lives in exactly one place. Mirrors the
+// backend tools in `hexa/mcp/tools/`; keep in sync when tools are added/renamed.
```
yolanfery, line 5: I think this was discussed in the past before my vacations but it would be better to expose those tools in the GQL layer to keep the sync and not having (and forgetting) to maintain this list at 2 places

frontend/src/assistant/helpers/toolConfig.ts
```diff
@@ -0,0 +1,28 @@
...
+  hideOutput?: boolean;
+  // Override the generic "Input" header when that word would mislead — e.g. the
+  // changeset is the agent's *proposed change*, not a function argument.
+  inputLabel?: ToolInputLabel;
+};
+
+const TOOL_CONFIG: Record<string, ToolUiConfig> = {
+  [TOOL.PROPOSE_PIPELINE_VERSION]: {
+    hideOutput: true,
+    inputLabel: "Proposed changes",
```
yolanfery: Does this end up translated ?

frontend/src/assistant/helpers/toolConfig.test.ts
```diff
@@ -0,0 +1,17 @@
+import { getToolConfig } from "./toolConfig";
+
+// Tests pin the literal tool-name contract (the string the backend sends) rather
+// than reusing the TOOL constant, so a wrong rename is caught instead of passing
+// tautologically.
```
yolanfery: > tautologically

Doing literature at this point

frontend/src/assistant/features/ChatPane/ToolValueSection.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    toolName: "",
+    success: true,
+    input: {},
+    output: {},
+    ...overrides,
+  };
+}
+
+describe("ToolValueSection", () => {
+  it("falls back to raw JSON (no view toggle) for an unrecognized shape", () => {
```
yolanfery, line 17: That's cool for maintenance, nice

frontend/src/assistant/features/ChatPane/ToolCallDetails.tsx
```diff
@@ -0,0 +1,88 @@
...
+function inputLabel(
+  t: TFunction,
+  override: ToolInputLabel | undefined,
+): string {
+  if (!override) return t("Input");
+  const labels: Record<ToolInputLabel, string> = {
+    "Proposed changes": t("Proposed changes"),
+  };
+  return labels[override];
+}
```
yolanfery: I am pretty sure we can do cleaner than this double mapping for transaltions

frontend/src/assistant/features/ChatPane/renderers/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+// until it is registered here.
+export function getRendererLabel(t: TFunction, label: RendererLabel): string {
+  const labels: Record<RendererLabel, string> = {
+    Files: t("Files"),
+    Code: t("Code"),
+    Document: t("Document"),
+    Table: t("Table"),
+  };
+  return labels[label];
+}
```
yolanfery: Same comment for translations, not fan at all of this new pattern

frontend/src/assistant/features/ChatPane/renderers/RendererBoundary.tsx
```diff
@@ -0,0 +1,20 @@
+import { ErrorBoundary } from "@sentry/nextjs";
+import { ReactNode } from "react";
+import JsonView from "core/components/JsonView";
+
+// Guards a semantic renderer against tool input/output shapes that drift from
+// what it expects: if rendering throws, we fall back to the raw JSON view
+// (and Sentry records the failure so we notice the structure change).
+export default function RendererBoundary({
```
yolanfery, line 8: Nice pattern

yolanfery (changes requested):
Looking very neat, good job ! Not reviewed deeply as this is frontend heavy. 
Great to see we are adding back some FE unit tests, we lost this habit...

Few comments about translations, typing and some UI/UX improvements

- Can we make the modal resizable to make it easier to see big queries ?   
- [Minor] Changing the view modes between `Files` and `Raw` is changing the modal and the conversation snippet at the same time, I think they should be independent
- Can we have some margin at the bottom ?
Image

frontend/src/assistant/helpers/toolConfig.ts
mrivar: It does, but I'll think of a better more readable solution

frontend/src/assistant/features/ChatPane/renderers/index.tsx
mrivar: yeah it looks ugly and not very readable, thanks for pointing it out, working on a better solution

mrivar:
Thanks for the reviews guys!

> Just a quick question why do we need to visualise the raw content/json?

@DimitriKwihangana the user should always be able to visualise the original raw JSON content that came from the tool. Everything else (tables, markdown, file listing special styling) are just nice to have for better readability. Also, the JSON serves as a fallback in case there's no special rendering for the tool.

> Few comments about translations, typing and some UI/UX improvements
> 
> * Can we make the modal resizable to make it easier to see big queries ?
> * [Minor] Changing the view modes between `Files` and `Raw` is changing the modal and the conversation snippet at the same time, I think they should be independent
> * Can we have some margin at the bottom ?

@yolanfery Added some changes to the PR:

- Modal state independent from the conversation snippet
- Added space around the modal so it's not crammed to the bottom of the screen
- Added functionality to resize the modal (from dragging corner + right/bottom sides)
- Refactored the translations labels to a more simple and readable feature, while keeping the configurability of it (from the complex remapping and repetition of labels to `label: (t) => t("Table")`)
- Side quest: updated sdk to have the latest version with the [conditional params fix](https://github.com/BLSQ/openhexa-sdk-python/pull/402)

frontend/src/assistant/helpers/tools.ts
mrivar: Regarding this, I'm giving it some thought.
If we have a GQL layer that exposes the tool and we get them from there, we lose the typed constants into the tools enum, losing the capability of just having an enum to use specific tools with specific uses in the FE code. Whenever we want to add specific FE logic for a tool ew would have to hardcode the tool, sometimes in several places, which is what I'm trying to avoid here (eg having `ctx.toolName === TOOL.GET_HELP_OR_DOC` instead of `ctx.toolName === "get_help_or_doc"`). This also makes it easier to track when do we have specific logic for specific tools.
If we however want to make sure we don't forget to have specific tools in the FE as well, we can add a lightweight test that checks for this. I think I'll go this route for now, wdyt?

yolanfery: GQL Enums are code-generated into a `type.ts` file that can be used in the frontend

An example is https://github.com/BLSQ/openhexa-app/blob/a6a90648b39853791b321a0dbe619c8ed35bb44d/frontend/src/graphql/types.ts#L6113-L6118

That's a pattern we have in the code especially for error handling

mrivar: huh, alright I'll look into it

mrivar: I'll elaborate/investigate about this in a separate branch, so this one can be reviewed as-is

frontend/src/assistant/features/ChatPane/renderers/index.tsx
```diff
@@ -38,10 +37,15 @@ function deletedFiles(value: unknown): string[] {
...
+//
+// `label` is a translator thunk (`(t) => t("Files")`) rather than a plain key:
+// the i18next parser only extracts literal `t("…")` calls, so keeping the call
+// here — co-located with the renderer — lets the key be picked up while staying
+// the single source of truth (no separate label→t() map to keep in sync).
 const RENDERERS: SemanticRenderer[] = [
   {
     id: "files-changeset",
-    label: "Files",
+    label: (t) => t("Files"),
```
yolanfery, line 48: 👍

yolanfery (commented):
Looking good, well done !

yolanfery (approved):

PR #1873, opened 2026-06-17
feat: api endpoint to allow executing a sql query against the workspace database (HEXA-1665)

yolanfery:
Endpoint for executing read-only SQL queries

## How/what to test

Unit tests should cover everything ☺️


backend/hexa/databases/utils.py
```diff
@@ -73,6 +73,63 @@ def get_workspace_database_connection(workspace: Workspace):
...
+        user=workspace.db_ro_username,
+        password=workspace.db_ro_password,
+    )
+
+
+def execute_database_query(
+    workspace: Workspace,
+    query: str,
+    timeout_ms: int = 10_000,
+    max_rows: int = 1000,
```
mrivar: Do we want to set a cap for this so it doesn't go rogue on an immense amount of rows?

backend/hexa/databases/utils.py
```diff
@@ -73,6 +73,63 @@ def get_workspace_database_connection(workspace: Workspace):
...
+            # Fetch one extra row to detect (without returning) that more exist.
+            fetched = cursor.fetchmany(max_rows + 1) if cursor.description else []
+        truncated = len(fetched) > max_rows
+        rows = fetched[:max_rows]
+        return {
+            "columns": columns,
+            "rows": rows,
+            "row_count": len(rows),
+            "truncated": truncated,
+        }
```
mrivar, line 156: do we want pagination for truncated queries? or the query is too broad to add generic pagination?

backend/hexa/databases/utils.py
```diff
@@ -73,6 +73,63 @@ def get_workspace_database_connection(workspace: Workspace):
...
+            # cursor.description is None for statements that do not return rows
+            columns = (
+                [column.name for column in cursor.description]
+                if cursor.description
+                else []
+            )
+            # Fetch one extra row to detect (without returning) that more exist.
+            fetched = cursor.fetchmany(max_rows + 1) if cursor.description else []
+        truncated = len(fetched) > max_rows
+        rows = fetched[:max_rows]
```
mrivar: For `rows` there might be an issue, we are enforcing it as a JSON but sometimes some columns might return non-serializable content such as bytes (e.g. a pipeline zipfile), which would raise an error (returning a 500 error)
I'd say it's worth it to add a test to verify this and if so look for solutions on how to address it

mrivar (dismissed):
Very solid code!
Just some minor comments and one possibel issue about binary/byte content columns

backend/hexa/databases/schema.py
```diff
@@ -8,12 +8,15 @@
 )
 from django.core.exceptions import PermissionDenied
 from django.http import HttpRequest
+from psycopg2 import Error as Psycopg2Error
+from psycopg2.errors import QueryCanceled
```
bramj, line 12: 😬 I find this American English spelling of "canceled" disturbing 😅

backend/hexa/databases/utils.py
bramj: Absolutely, we should have a hard limit (which I think can be a simple ENV variable) of for example 100K.
And I'd put the default a lot lower actually, more like 50 for example, as a regular "pagination".

This PR actually implements almost everything described in this other ticket :p https://bluesquare.atlassian.net/browse/HEXA-1666

backend/hexa/databases/utils.py
```diff
@@ -73,6 +73,63 @@ def get_workspace_database_connection(workspace: Workspace):
...
+    A per-statement timeout is set so that a single request cannot hold database
+    resources for an extended period of time. At most ``max_rows`` rows are
+    returned; ``truncated`` indicates whether the result was capped.
+    """
+    conn = None
+    try:
+        conn = get_workspace_database_ro_connection(workspace)
+        with conn.cursor(cursor_factory=RealDictCursor) as cursor:
+            cursor.execute(
+                sql.SQL("SET LOCAL statement_timeout = {timeout};").format(
```
bramj, line 136: 👍

backend/hexa/databases/utils.py
bramj: I think we can leave that burden on the user for now, if they want to paginate they can do so as they wish by setting a `LIMIT` and `OFFSET` on their query.
When we add the frontend, we'll see if it becomes part of the interface there as well I guess.

backend/hexa/databases/tests/test_schema.py
```diff
@@ -327,6 +333,108 @@ def test_get_database_table_rows(self):
                 r["data"]["workspace"]["database"],
             )
 
+    def _seed_demo_table(self, rows):
```
bramj: [Nitpick] This one is copied in `test_utils.py` as well, maybe we can extract it?

backend/hexa/databases/utils.py
```diff
@@ -73,6 +73,63 @@ def get_workspace_database_connection(workspace: Workspace):
...
+            )
+            cursor.execute(query)
+            # cursor.description is None for statements that do not return rows
+            columns = (
+                [column.name for column in cursor.description]
+                if cursor.description
+                else []
+            )
+            # Fetch one extra row to detect (without returning) that more exist.
+            fetched = cursor.fetchmany(max_rows + 1) if cursor.description else []
```
bramj, line 148: This is interesting, I think it could be the fallback for when we are unable to append a `LIMIT` to the query (https://bluesquare.atlassian.net/browse/HEXA-1666)
But as I understand it, this alone won't be enough since the DB will still fetch all rows, and it's only the Python part that's optimized.
But okay for this PR!

bramj (changes requested):
@yolanfery This is an exciting feature 😁 

I think we should add that hard limit, but other than that this looks like a great start for the Data Studio! 👏

backend/hexa/databases/utils.py
yolanfery: Sounds good, I will do the changes

Funny that I ended up implementing the same ticket specs, I am taking it up to implement it fully

backend/hexa/databases/utils.py
yolanfery: Yes that was my reasoning as well

backend/hexa/databases/utils.py
yolanfery: Completely, thanks for catching and raising this, will do

backend/hexa/databases/tests/test_schema.py
yolanfery: Yep makes sense

backend/hexa/databases/utils.py
yolanfery: I will look into cascading this into the DB query as part of this PR, I didn't want to parse/modify/wrap the query at first but I think it makes sense ultimately

yolanfery: Ok, I looked a bit at it with Claude

Wrapping/adding the limit at the SQL level requires to limit what the SQL query can do (duplicate column names would not work [like `SELECT 1,1`], `EXPLAIN` queries would be a problem as well,..). Error messages are less clear also. We can try to overcome all those cases one by one but it feels hacky to me at this point

I have added the limitation of running one statement to prevent overriding the timeout. I think that with this timeout and materialization limitation at the python level we should be in a good spot. But we can still revisit this later...

bramj: :+1: agreed. It does feel like a whole can of worms trying to manage all edge cases for that. We can look further into that when we feel the need for it, maybe I went a bit too premature optimization there.
I think my main concern was that queries on a huge table could consume too much memory on the server, but maybe the `statement_timeout` is enough.

bramj (approved):
Nice work! 👍

PR #1874, opened 2026-06-17
feat: editing web apps with assistant(HEXA-1699)

DimitriKwihangana:
Added assistant to edit static web app next to code editor

## Changes
-  Added AI assistant to static web app editor (next to code)
- Propose and preview file changes before applying them
- Extracted shared  components reused by both pipeline and webapp  assistants
- Added  migration for EDIT_WEBAPP instruction set choice


backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,126 @@
...
+            .first()
+        )
+
+    if pending and pending.tool_output:
+        for f in pending.tool_output.get("files", []):
+            current_files[f["path"]] = f["content"]
+    else:
+        for f in webapp.get_files():
+            if f.get("encoding") == FileEncoding.TEXT and f.get("content") is not None:
+                current_files[f["path"]] = f["content"]
```
bramj, line 93: Nice how much simpler this is without the zipfile :)

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,126 @@
...
+        current_files.pop(path, None)
+
+    return {"files": [{"path": k, "content": v} for k, v in current_files.items()]}
+
+
+class EditWebappAgent(BaseAgent):
+    instruction_set = InstructionSet.EDIT_WEBAPP
+    tools = [
+        get_help_or_doc,
+        get_static_webapp,
```
bramj: Not sure this one makes sense actually, the instructions should contain the web app, no?

I do think there's a need for something like `get_static_webapp_file` with a from/to line number.
Web apps can get very large and I just hit the max context by just reading an app.

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,126 @@
...
+        if pending and pending.tool_output:
+            proposed_files = pending.tool_output.get("files", [])
+            if proposed_files:
+                lines.append("")
+                lines.append(
+                    "### Files (Pending Proposed Version — not yet accepted by the user)\n"
+                    "Use these files as the base for any further changes."
+                )
+                for f in proposed_files:
+                    lines.append(f"\n#### {f['path']}\n```\n{f['content']}\n```")
```
bramj: I think the actual contents are missing here, the agent's first action is always `get_static_webapp`.

As for the content returning for the files: we should be careful, since right now we'd also dump in binary files like images, which we don't really want the agent to manipulate directly.

Managing large files for web apps will be a bigger challenge, but I'd say for this first version we can try something like:
- Check the `lineCount` of the returned files, if it's `null` (indicating it's not a text file) or larger than a certain threshold: we only return the metadata, and not the contents.
- Add a tool to fetch the contents of a specific file so the agent can read it.

Then we should start thinking about how we'll handle large files like CSVs with lots of lines, the agent should be able to read just the "head" or "tail" of those files, as well as do "grep" operations.

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,126 @@
...
+    Unchanged files are preserved automatically.
+    """
+    current_files: dict[str, str] = {}
+
+    pending = None
+    if conversation is not None:
+        pending = (
+            ToolInvocation.objects.filter(
+                message__conversation=conversation,
+                tool_name="propose_webapp_changes",
```
bramj: For consistency, why not `"propose_webapp_version"`?

backend/hexa/assistant/schema/queries.py
```diff
@@ -38,6 +40,15 @@ def resolve_pipeline_assistant_conversations(pipeline: Pipeline, info, **kwargs)
     )
 
 
+@webapp_object.field("assistantConversations")
+def resolve_webapp_assistant_conversations(webapp: GitWebapp, info, **kwargs):
```
bramj, line 44: I think there's not test coverage for this resolver (and same for the other related ones), I think it makes sense to add some tests to verify that the conversations are scoped on a webapp/pipeline + user.

backend/hexa/assistant/instructions.py
```diff
@@ -68,11 +69,34 @@ class InstructionSet(TextChoices):
...
+Static web apps consist of HTML, CSS, and JavaScript files served as-is. An `index.html` file at the root is required.
+The web app may also call OpenHEXA's GraphQL API via a same-origin proxy at POST /graphql/ — no auth token needed, the user's session handles it.
+"""
+
 _INSTRUCTION_SETS: dict[InstructionSet | tuple[str, str], str] = {
     InstructionSet.GENERAL: _BASE,
     InstructionSet.CREATE_PIPELINE: _BASE + _CREATE_PIPELINE + _PIPELINE_DOCS,
     InstructionSet.EDIT_PIPELINE: _BASE + _EDIT_PIPELINE + _PIPELINE_DOCS,
     InstructionSet.CREATE_WEBAPPS: _BASE + _WEBAPPS,
+    InstructionSet.EDIT_WEBAPP: _BASE + _EDIT_WEBAPP,
```
bramj, line 106: We should add the web app docs here: https://docs.openhexa.com/static-webapps/

frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx
```diff
@@ -188,18 +189,10 @@ const WorkspacePipelineCodePage: NextPageWithLayout = (props: Props) => {
...
-                    className="flex items-center gap-1 text-blue-500 hover:text-blue-700 text-xs"
-                  >
-                    <XMarkIcon className="h-3.5 w-3.5" />
-                    {t("Dismiss")}
-                  </button>
-                </div>
+                <AssistantProposalBanner
+                  label={t("Proposed version from AI assistant")}
+                  onDismiss={handleDismiss}
+                />
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/pipelines/[pipelineCode]/code.tsx:211]: I think I liked the alignment of the before-version better, but I think adding the "Accept" button to it makes sense 👍

bramj (changes requested):
It's amazing how fast you could bring this functionality to the web apps editor 😁 

A couple of comments, I think the main most important question is how we handle potentially large files for web apps.

Let's try to start simple: I have a proposition in a comment below, and then we can gradually add complexity in other PRs. I would advise to try to edit a "real" app with images and several larger files to be able to test how well the feature holds up. If you find a better way for this v1, I'd be curious to hear that!

Another note: I think the agent is currently unable to delete files, but that it's fixed in this PR https://github.com/BLSQ/openhexa-app/pull/1877. Good idea to rebase your PR once the other one is merged and double-check that deleting files is working correctly.

backend/hexa/assistant/agents/edit_webapp_agent.py
yolanfery: At the minimum we could include the `index.html` this is the one served everywhere.

We need to use a constant for this so that if this logic changes, it is reflected here

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,215 @@
...
+            if proposed_files:
+                lines.append("")
+                lines.append(
+                    "### Files (Pending Proposed Version — not yet accepted by the user)\n"
+                    "Use these files as the base for any further changes."
+                )
+                for f in proposed_files:
+                    content = f["content"]
+                    line_count = content.count("\n") + 1
+                    if line_count > _MAX_INLINE_LINES:
```
bramj, line 201: We can handle that in a follow up PR, I guess this will still easily blow up when you have e.g. minified JS where the whole file is on a single line.
I suppose the safer way would be to look at character count.

backend/hexa/mcp/tools/webapps.py
```diff
@@ -195,3 +197,51 @@ def edit_static_webapp_file(
...
+
+
+@tool
+def get_static_webapp_file(
+    user,
+    workspace_slug: str,
+    webapp_slug: str,
+    path: str,
+    start_line: int | None = None,
+    end_line: int | None = None,
```
bramj, line 211: Great!

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,215 @@
...
+        current_files[patch.path] = original.replace(
+            patch.old_string, patch.new_string, 1
+        )
+
+    for f in modified_files or []:
+        current_files[f.path] = f.content
+    for path in deleted_files or []:
+        current_files.pop(path, None)
+
+    return {"files": [{"path": k, "content": v} for k, v in current_files.items()]}
```
bramj, line 117: I might be mistaken, but I think we can return just something like `{"status":"ok"}` here. While testing locally I think returning the full contents here blows up the context for no good reason.
What do you think?

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,215 @@
...
+    """Propose changes to the web app files.
+
+    Two ways to describe changes — use whichever fits:
+
+    - modified_files: pass the full new content for files you created or rewrote entirely
+      (each entry needs a 'path' and 'content').
+    - file_patches: for targeted edits to existing files, pass {path, old_string, new_string}.
+      The backend finds old_string in the current file and replaces it with new_string.
+      Prefer this over modified_files when changing a few lines of a large file — you
+      only need to read and reproduce the lines that actually change.
```
bramj, line 48: This is awesome! I bet the `file_patches` way will become very much the preferred way.

backend/hexa/assistant/agents/edit_webapp_agent.py
```diff
@@ -0,0 +1,215 @@
...
+    tools = [
+        get_help_or_doc,
+        get_static_webapp_file,
+        list_datasets,
+        get_dataset,
+        preview_dataset_file,
+        list_connections,
+        list_files,
+        read_file,
+        propose_webapp_version,
```
bramj, line 134: Can we get the `list_pipelines` and `get_pipeline` here as well? I think a common question will be to make a web app to launch a pipeline. With this the agent will be able to check what it looks like and what the expected parameters are.

bramj (changes requested):
Really awesome @DimitriKwihangana , big improvements since the last review 👍 

I worry a bit about the context size for ongoing conversations, but I think the reading with specific start/end lines and doing patch updates should improve things a lot.

I think the main culprit is the return statement of `propose_webapp_version`. If you do a conversation of several back-and-forth edits, you can see the context increase very fast because of this. In logfire you can see that all tool input and outputs from the entire conversation are sent each time. Can you confirm this and change it to see if it effectively reduces the context accumulation?

And then a few small UX issues:

The save button is overlapped on my machine:
Image

Small spacing issue and I think there was an "Apply" button there somewhere?
Image

backend/hexa/mcp/tools/webapps.py
yolanfery: Nice to see that you are progressing on [this ticket](https://bluesquare.atlassian.net/browse/HEXA-1707) as part of this PR ☺️

Just a comment for this tool, I would consider if it is possible to make it a graphQL query and calling it here. It's the pattern we use for the MCP tools since it allows to maintain one layer of logic (if the frontend/cli/sdk ever needs to read one file, the GQL resolver will be there already)

backend/hexa/assistant/agents/edit_webapp_agent.py
DimitriKwihangana: I do think this is a good Idea!

DimitriKwihangana:
> Really awesome @DimitriKwihangana , big improvements since the last review 👍
> 
> I worry a bit about the context size for ongoing conversations, but I think the reading with specific start/end lines and doing patch updates should improve things a lot.
> 
> I think the main culprit is the return statement of `propose_webapp_version`. If you do a conversation of several back-and-forth edits, you can see the context increase very fast because of this. In logfire you can see that all tool input and outputs from the entire conversation are sent each time. Can you confirm this and change it to see if it effectively reduces the context accumulation?
> 
> And then a few small UX issues:
> 
> The save button is overlapped on my machine: Image
> 
> Small spacing issue and I think there was an "Apply" button there somewhere? Image

Thanks for the comments @bramj. For the "Apply" I did remove it intentionally it was playing the same role as the "save" button. I was also aligning with the UI on the piplines

bramj:
> Thanks for the comments @bramj. For the "Apply" I did remove it intentionally it was playing the same role as the "save" button. I was also aligning with the UI on the piplines

:+1: Consistency is good. I did like the idea, we might introduce it later again. The current UI where the "save" button appears linked to a specific file, but actually saves all files is not great, but good enough for now.

bramj (approved):
Nice work @DimitriKwihangana on finishing up this big one! 👏 

I think we'll feel the pain of a need for more context management very soon, but really cool to see how this is already working.

E.g. the agent appears a bit unsure about the current state of the code due to the change in the tool result. There's room for improvement, but it's better than before.

PR #1877, opened 2026-06-18
fix: improve `udpate_static_webapp` MCP tool docstring and allow deleting files (HEXA-1703)

yolanfery:
- Add a `filesToDelete` attribute in the GQL layer 
- Wire it to the MCP tool
- Update Forgejo client to accept deletions
- Update MCP tool docstring


backend/hexa/mcp/tools/webapps.py
```diff
@@ -98,13 +98,16 @@ def update_static_webapp(
...
 
-    Provide the webapp UUID (from list_static_webapps) and any fields to change. Only provided non-empty fields are updated. Pass files_json as a JSON array of {path, content} objects to replace all files (e.g. '[{"path": "index.html", "content": "<html>...</html>"}]'). Pass name to change the human-readable name, and description to change the description.
+    Provide the webapp UUID (from list_static_webapps) and any fields to change. Only provided non-empty fields are updated. Pass name to change the human-readable name, and description to change the description.
+
+    Files are updated incrementally — you do NOT need to resend the whole bundle. Pass files_json as a JSON array of {path, content} objects containing only the files you want to add or change (e.g. '[{"path": "app.js", "content": "..."}]'); any file you omit is left exactly as it is. Matching is by path: an existing path is overwritten, a new path is created. To remove files, pass files_to_delete_json as a JSON array of paths (e.g. '["old.js", "legacy/style.css"]'); paths that don't exist are ignored. You can combine files_json and files_to_delete_json in a single call — they are applied as one commit.
```
bramj, line 110: I think soon we should experiment with a tool that allows a precision edit like "replace lines 40-45 with X". That would also be relevant for #1874 

What do you think?

yolanfery: That can be useful and is probably worth the cost of adding more tool/complexity on the MCP side, yes 

What do you think about a new tool `edit_static_webapp_file` doing a text search and replace ? Probably easier to understand/debug than line number edits

yolanfery: (And I think we can keep this PR, just adding a new tool as a separate PR)

bramj: Sounds good, I think that's how local agent edit files as well 👍

yolanfery: Merging, will open a PR for precise edits ad defined here

PR #1878, opened 2026-06-19
feat: `edit_static_webapp_file` MCP tool (HEXA-1703)

yolanfery:
Allow to do a search and replace with this new MCP tool

Implementation through the GQL interface to make this available to other clients (FE, SDK,...)

### DIFF
--- a/backend/hexa/git/client.py
+++ b/backend/hexa/git/client.py
@@ -45,6 +45,7 @@ class GitClient(ABC):
         *,
         org_slug: str | None = None,
     ) -> bytes:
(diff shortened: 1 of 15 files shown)


bramj (approved):
This is awesome and it aligns perfectly with how Claude Code does it: https://code.claude.com/docs/en/tools-reference#edit-tool-behavior !

I think this will become the general preferred way to edit files and not just "small changes".

@DimitriKwihangana I think this tool is very relevant for your PR, we should add it later or you can rebase on this.

PR #1882, opened 2026-06-22
Add a script + admin page to duplicate a workspace (v1) (HEXA-1663)

bramj:
This PR introduces the `workspace_duplicator` Django app. It's a (right now) Django app without models that contains a script that allow duplicating a workspace with the following resources:
- Workspace metadata
- Files
- Pipelines
- Connections

Other resources such as Database, Datasets and Web apps will come in later iterations.

Notes:
- Tested and working well for copying large workspaces (with a lot of files including large ones, as well as pipelines) from prod to demo or from demo to demo.
- Copying from localhost to localhost should also work well with the examples below. Copying files from prod/demo to localhost however has some issues, but I'm not sure if that's worth fixing.
- The final commit adds a few fields that don't exist yet on demo and prod, and can produce errors because of this.

## Changes

Detailed changes are in the new README.md file: `backend/hexa/workspace_duplicator/README.md`

## How to test

There are 2 ways to execute the script:

### CLI

Example usage:

```bash
# Localhost to localhost
./manage.py copy_workspace \
	--source-workspace-slug my-workspace \
	--source-url http://app:8000/graphql/ \
	--source-token 'source-service-account-token' \
	--target-url http://app:8000/graphql/ \
	--target-organization org-uuid \
	--target-workspace-name "My Workspace (copy)" \ # optional
	--target-token 'target-service-account-token'

# Production to demo
./manage.py copy_workspace \
	--source-workspace-slug my-workspace \
	--source-url https://api.openhexa.org/graphql/ \
	--source-token 'my-prod-service-account-token' \
	--target-url https://api.demo.openhexa.org/graphql/ \
	--target-organization 002f2f74-7cdb-452c-8ef5-28cc27c04fbe \ # BLSQ org
	--target-workspace-name "My Workspace (copy)" \
	--target-token 'my-demo-service-account-token'
```

Note: `--target-workspace-name` is optional and defaults to the same name as the source workspace.

### Django Admin:

On the workspaces list page, there's a new button "Migrate workspace". This provides a view to easily run the CLI script.

The view is Superuser-only and entered credentials are used transiently and never persisted.

### CLI

image

### Django admin

image


backend/hexa/workspace_copier/management/commands/copy_workspace.py
```diff
@@ -0,0 +1,106 @@
...
+                resources=resources,
+                reporter=reporter,
+            )
+        except CredentialError as exc:
+            raise CommandError(
+                "Endpoint verification failed:\n"
+                + "\n".join(f"  - {e}" for e in exc.errors)
+            )
+        except GraphQLError as exc:
+            raise CommandError(str(exc))
```
yolanfery, line 108: This might leak passwords in sentry no ?

backend/hexa/workspaces/templates/admin/workspaces/workspace/change_list.html
```diff
@@ -0,0 +1,13 @@
+{% extends "admin/change_list.html" %}
+{% load i18n %}
+
+{% block object-tools-items %}
+{% if user.is_superuser %}
+<li>
+    <a href="{% url 'admin:workspaces_workspace_migrate' %}" class="addlink">
+        {% trans 'Migrate workspace' %}
+    </a>
```
yolanfery: We could tie the button to selected workspaces also instead of having a generic "free-floating" button

backend/hexa/workspace_copier/management/commands/copy_workspace.py
yolanfery: That's why I don't like the password approach and claims that `entered credentials are used transiently and never persisted.`, it's very difficult (impossible ?) to guarantee that. Passwords are very sensitive and private (users might use same passwords on other products). Probably a good idea to use tokens instead at the minimum

backend/hexa/workspace_duplicator/tests/resources/test_pipelines.py
```diff
@@ -0,0 +1,135 @@
...
+    def setUp(self):
+        self.source = Endpoint.remote(MagicMock(), "src")
+        self.target = Endpoint.remote(MagicMock(), "tgt")
+        self.result = DuplicationResult()
+
+    @patch("hexa.workspace_duplicator.resources.pipelines._update_settings")
+    @patch("hexa.workspace_duplicator.resources.pipelines._upload_versions")
+    @patch("hexa.workspace_duplicator.resources.pipelines._create_on_target")
+    @patch("hexa.workspace_duplicator.resources.pipelines._fetch_source_detail")
+    @patch("hexa.workspace_duplicator.resources.pipelines._list_source_ids")
```
yolanfery: That's a lot of mocks, can't we do our tests at a higher level of abstraction ? A bit like we do for the MCP. Genuinely  wondering

Because this seems to be difficult to maintain and this does not catch issues like breaking changes

backend/hexa/workspace_duplicator/tests/test_forms.py
```diff
@@ -0,0 +1,54 @@
...
+        "target_email": "",
+        "target_password": "",
+        "target_organization": "",
+        "resources": ["connections"],
+    }
+    data.update(overrides)
+    return data
+
+
+class MigrateWorkspaceFormTest(SimpleTestCase):
```
yolanfery: Cool tests 👍

backend/hexa/workspace_copier/progress.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def log(self, message: str, *, level: str = "INFO") -> None:
+        prefix = "" if level == "INFO" else f"[{level}] "
+        self.stream.write(f"{prefix}{message}\n")
+
+
+class BufferReporter(BaseReporter):
+    """Collect lines in memory for rendering after the run.
+
+    This is used in the Django admin view. Later, we'll adapt this reporter
```
yolanfery, line 63: Cool stuff

backend/hexa/workspace_copier/transport.py
```diff
@@ -0,0 +1,76 @@
...
+        client,
+        "mutation Login($input: LoginInput!) { login(input: $input) { success errors } }",
+        {"input": {"email": email, "password": password}},
+        "Login",
+    )
+    if not data["login"]["success"]:
+        raise GraphQLError(
+            f"{label} login failed: " + ",".join(data["login"]["errors"] or [])
+        )
+    return client
```
yolanfery: I think we might be able to replace this with the SDK client and a token

backend/hexa/workspace_duplicator/README.md
```diff
@@ -0,0 +1,97 @@
+# workspace_duplicator
+
+Copy an OpenHEXA workspace (and its resources) from one server to another, or between two workspaces on the same server.
```
yolanfery: Here you use duplicate and copy wording but you mostly use `duplicate` in the codebase. I think we should settle for one. There are also occurrences of migrate

And for me duplicate was not really accurate because I thought it meant in place. So I asked Claude and found the answer pretty convincing : 

> Is duplicate a good name for this feature ? What about copy ?                                                                                                                                                                                                                                                          

> ⏺ Short answer: "copy" is the best of the three, and the bigger problem is that you're currently using all of them. The same feature is called workspace_duplicator (app), migrate_workspace (command), run_migration (service), duplicate_workspace (orchestrator), and WORKSPACE_COPIERS (copiers). Four verbs across
>   five layers — that's the thing actually worth fixing.
> 
>   On the verb itself, the deciding question is: does the source survive? It does — this reads the source and writes a copy to the target, leaving the source intact. That semantics rules them out in this order:
> 
>   - migrate — implies a move: source gets decommissioned or repointed afterward. Misleading here, since nothing happens to the source. (It's the Jira ticket title, which is probably why it leaked into the command name — but the ticket describes the goal, not what the code does.)
>   - duplicate — accurate, but it conventionally means cloning in place / same context ("duplicate this row"). Across two servers it reads slightly off, and the noun/verb forms are clunky (duplicator, duplicate_workspace).
>   - copy — exactly right: source preserved, a copy created elsewhere, neutral about whether "elsewhere" is the same server or another. Cleanest verb and noun forms (copy_workspace, WORKSPACE_COPIERS already fits).
> 
>   So I'd standardize on copy: workspace_copy (or workspace_copier) app, copy_workspace command, copy_workspace() service, WORKSPACE_COPIERS stays as-is. That also means renaming CredentialError/run_migration and the migrate_workspace.py file/template.
> 
>   One caveat worth confirming before you rename everything: is the real intent actually a move? If the use case is "stand up a workspace on a new server and stop using the old one" (server consolidation, prod cutover), then "migrate" is the honest product name even if the code doesn't delete the source — and
>   you'd unify on migrate instead. If it's "I want a second copy" (templating, backup, clone-for-testing), it's "copy."
> 
>   Which is the actual use case — moving workspaces to a new home, or making reusable copies? That determines copy vs. migrate; either way, pick one and unify.

backend/hexa/workspace_duplicator/forms.py
```diff
@@ -0,0 +1,90 @@
...
+        help_text="GraphQL endpoint of the target server. Leave blank for the local server.",
+    )
+    target_email = forms.EmailField(required=False, label="Target superuser email")
+    target_password = forms.CharField(
+        required=False, label="Target superuser password", widget=forms.PasswordInput
+    )
+    target_organization = forms.CharField(
+        required=False,
+        label="Target organization id",
+        help_text="Optional UUID of the organization to create the workspace under.",
```
yolanfery: I think this is required no ? I tried without it and had an error, worth double-checking

backend/hexa/workspaces/templates/admin/workspaces/workspace/change_list.html
yolanfery: Actually, after testing, your button is really generic and allows bi-directional copies, so it's probably great like this already

yolanfery (changes requested):
Well done this is a huge step forward 🚀

Tested, feels like magic 

A bit of concern about maintenance since with this we are mixing everything (using the SDK client to do perform some dynamic GQL queries, using SDK client types,...), that's why I suggest to try and spend a bit more time on testing at a higher level to ensure breaking changes are not bitting us through this command (if this is possible ofc) and refactoring is simplified with test coverage.


But besides that and some minor stuff around auth and naming, it's a great first step ! 
It was a good call moving this in within this repo and allowing execution through the admin panel

backend/hexa/workspace_copier/management/commands/copy_workspace.py
bramj: I agree with you. Additionally, when this becomes a async job, this becomes more important since we'll need to persist them at some point (and probably clean up after the run, but it's risky business).
Will change to use tokens.

backend/hexa/workspaces/templates/admin/workspaces/workspace/change_list.html
bramj: Yep I agree it's not very polished, but it's really and admin-only entrypoint so flexible and "good enough" :)

backend/hexa/workspace_duplicator/README.md
bramj: Arg this is a big one. But I agree and Claude makes some convincing arguments 😅 

Claude and I will rename everything to copy 👍

backend/hexa/workspace_duplicator/forms.py
bramj: Agreed!

bramj:
@yolanfery Changes wrt last review:

- Full rename: `copy` and `copier` everywhere
- Target org uuid is now required
- Swapping out user/pass for tokens (and improve the `transport.py` layer to better use the SDK)
- Small fixes and improvements:
  - better handle pipelines with exactly the same name
  - improved error handling for file copying (with adding to summary)

For the test strategy: TBH I haven't found a better way to avoid the mocking for now. I had Claude refactor a couple of times, but the end results were always convoluted. I'm not really happy with all of that for now, but I can live with it in this state for now if okay for you.

backend/hexa/workspace_copier/transport.py
```diff
@@ -0,0 +1,75 @@
...
+    label: str,
+    http_client: httpx.Client | None = None,
+) -> Client:
+    """Build an SDK client authenticated with a Bearer token and verify it.
+
+    The token is a ServiceAccount token, which the backend authenticates per
+    request via the ``Authorization: Bearer`` header — so there is no login
+    round-trip or session cookie to manage. A cheap ``me`` query is run up front
+    so an invalid or under-permissioned token surfaces here rather than mid-copy.
+
```
yolanfery, line 57: 👌

yolanfery (approved):
> For the test strategy: TBH I haven't found a better way to avoid the mocking for now. I had Claude refactor a couple of times, but the end results were always convoluted. I'm not really happy with all of that for now, but I can live with it in this state for now if okay for you.

Sure ! 

Looking great now , I like the simpler auth system with the token

PR #1885, opened 2026-06-24
feat: agent tools automagic (HEXA-1682)

mrivar:
Autogenerate agent tools, using command `dump_agent_tools` and `npm run codegen`.
Added a CI test to ensure we don't forget to run it.
Added a test to ensure all tools have assigned translated names (so we don't forget to assign them to new added tools)


yolanfery (commented):
Wow, you went further than I expected, I didn't envision generating the GQL enums based on the backend tools list ! 

The thing that is bothering me with this approach is that we don't reference the enum in the mutation input/output and are still relying on strings. And it also adds CI and dev (commands to be aware) complexity. Also this is a new pattern that is not used in this codebase.

What do you think of simply maintaining the list in the GraphQL schema (which is a pattern we use elsewhere), using it for the mutations and maybe (this part is optional if not feasible) see if we can check (through a unit test) that the set of all python tools for all agent is present in this enum list ?

mrivar:
> Wow, you went further than I expected, I didn't envision generating the GQL enums based on the backend tools list !
> 
> The thing that is bothering me with this approach is that we don't reference the enum in the mutation input/output and are still relying on strings. And it also adds CI and dev (commands to be aware) complexity. Also this is a new pattern that is not used in this codebase.
> 
> What do you think of simply maintaining the list in the GraphQL schema (which is a pattern we use elsewhere), using it for the mutations and maybe (this part is optional if not feasible) see if we can check (through a unit test) that the set of all python tools for all agent is present in this enum list ?

Yeah that's fair enough, I guess I wanted to make it extra cool but it's a bit of over-engineering 😂

About having the enum referenced in the mutation input/output, I'm not 100% sure. If we rename /remove a tool, any historical conversation that used that tool will fail GraphQL, so it adds a trap that we might not even realise about. What do you think?

yolanfery:
> About having the enum referenced in the mutation input/output, I'm not 100% sure. If we rename /remove a tool, any historical conversation that used that tool will fail GraphQL, so it adds a trap that we might not even realise about. What do you think?

I think it is exactly what we want : to force us to either keep the tool for historical reasons or to migrate it but not having a string (generic) representation of it that will also change but will be unoticed and dis-aligned with the new list of tools

We will realise there is an issue if we have a unit test comparing the GQL list with the tools list

mrivar:
> > About having the enum referenced in the mutation input/output, I'm not 100% sure. If we rename /remove a tool, any historical conversation that used that tool will fail GraphQL, so it adds a trap that we might not even realise about. What do you think?
> 
> I think it is exactly what we want : to force us to either keep the tool for historical reasons or to migrate it but not having a string (generic) representation of it that will also change but will be unoticed and dis-aligned with the new list of tools
> 
> We will realise there is an issue if we have a unit test comparing the GQL list with the tools list

Well, the issue I see is that the unit test does make it harder to change, but not impossible (it just adds a second change to do). Especially with the rising use of AI agents/tools, it might even just change it on the side without the unit test triggering a failure, and then we would have the bug in old conversations.

One solution we can have is the enum AssistantToolName as a new optional parameter, and keep toolName as a string. In the unfortunate case of an unknown tool (old/renamed tool) we still read through it without the GQL layer failing silently.
We would have an extra layer of safety there, even if it's a case that should never happen

yolanfery (approved):
Ok, I come to realize that it can a matter of preference : I prefer having the GQL enum totally aligned with the database options even if this means maintaining older values (and at some point deciding to clean them up with a DB migration and GQL breaking change). But you propose to keep historical values intact, leaning towards convenient string tool name. 

I prefer the intermediate solution that you propose by combining the two over the string approach only and I think we spent enough time on this already, so I am good with this ☺️

I also like that we have not added the dev/CI complexity 
Well done and thanks for refactoring this as promised

PR #1887, opened 2026-06-24
fix: sub tab for web app api access permissions (HEXA-1722)

yolanfery:
Create a separate tab for web api access permissions


frontend/src/core/components/forms/Checkbox/Checkbox.tsx
```diff
@@ -22,14 +33,18 @@ const Checkbox = (props: CheckboxProps) => {
           id={id ?? name}
           name={name}
           type="checkbox"
```
DimitriKwihangana, line 35: I do think toggles would a good choice for this functionality. For UX perspective I would prefer toggle in giving access.

frontend/src/webapps/features/WebappApiAccess/WebappApiAccess.tsx
```diff
@@ -0,0 +1,230 @@
...
+          "rounded-md border p-3",
+          section.isEdited
+            ? "border-blue-300 bg-blue-50"
+            : "border-gray-200 bg-gray-100",
+        )}
+      >
+        <p className="text-xs font-medium text-gray-500 mb-2">
+          {t("Allowed operations")}
+        </p>
+        <div className="space-y-2">
```
DimitriKwihangana: Also For me, I would prefer to group the map for example. having a title of dataset. and below we add its read and write and we do the same on pipeline. and we put the toggles on the right side. uploading a picture of what I think from claude ofc 😀.

Image

DimitriKwihangana (changes requested):
It  looks very good @yolanfery ! I liked the idea of adding a new tab. I only added a few suggestions let me know what you think.

frontend/src/webapps/features/WebappApiAccess/WebappApiAccess.tsx
yolanfery: Thanks a lot for your review and suggestion ☺️

I really like your idea and followed what you proposed

Screenshot 2026-06-29 at 16 53 04

DimitriKwihangana (approved):
This looks very cool!

PR #1888, opened 2026-06-26
feat: Bring AI settings to the organization level (HEXA-1721)

bramj:
All AI configuration is managed at the organization level. This means that every organization has their AI assistant in one of these states:

- **managed**: We manage and choose the models that best work in our experience. This is the default for all SaaS subscription workspaces
- **bring your own key/model**: The org does their own config, with the technical limitation that they use the same model for all agents
- **disabled**: AI features are fully disabled

Full details: https://bluesquare.atlassian.net/browse/HEXA-1721

## Changes

- Migrate AI Assistant settings from the user level to the organization level, so AI configuration and access is managed per-organization rather than per-user
  - Frontend: removed from the user settings, moved to the organization settings
- Add AI usage limits and budgeting: `monthly_ai_budget` on organization subscriptions, enforced on the backend and visualized in the organization settings UI
- Take enabled/disabled + limits state into account on frontend:
  - Hide "Create/Edit with AI" when AI is disabled
  - Disable buttons with explanation when org's AI budget is consumed
- Wire up Vertex AI for managed OpenHEXA instances
- Remove the `assistant` feature flag

## How/what to test

- Configure the ENV vars (see `.env.dist` changes) to turn your local instance into a "managed" instance
- Verify that user settings no longer have AI config
- Try enabled/disable on organization settings to see how AI features become enabled/disabled
- Use Django admin to set a very low budget to see if it's correctly enforced
- Make sure the Google Vertex hookup works
- Pretend your local instance is a localhosting, confirm you can configure an Anthropic key to use AI features

### Settings
image

image

### Limits reached

image

### Django admin

image


backend/hexa/assistant/model_builder.py
```diff
@@ -1,76 +1,130 @@
...
+            # We control the model for managed orgs; ignore any stored value.
+            model = MANAGED_DEFAULT_MODEL
+        else:
+            model = ai_settings.model
+        self._model_api_name = get_api_name(ai_settings.provider, model)
 
     @classmethod
     def from_conversation(cls, conversation: Conversation) -> "AiModelBuilder":
-        ai_settings: AiSettings = conversation.user.ai_settings_safe
+        organization = conversation.workspace.organization
```
yolanfery, line 101: Organization can be null for localhosting

yolanfery (approved):
Exciting to see us moving forward with fully managed agent 🤩

Reviewed and tested, not much to say (except that it's great), well done ! 

Just a minor issue with null orgs and on the admin view, the API key shouldn't be mandatory for managed 

Image

backend/hexa/assistant/model_builder.py
bramj: Good catch! We should fix that actually, I feel it's not the first time this bites me.
I'll create a ticket for that.

yolanfery: Yes I agree, we might want to have a migration script creating an org if none exist and assign all workspaces to it and remove nullable on the org attribute

yolanfery: But how to decide about the name of the org is not obvious 😅

bramj: Lol for sure :sweat_smile: fyi https://bluesquare.atlassian.net/browse/HEXA-1729

bramj:
Note to self: is there a default setting for the new budget column?

bramj:
@yolanfery few small improvements:

- Be more defensive about the case where `workspace.organization` is `None`
- Django admin model + api key validation when "managed"
- Don't break if Console doesn't provide monthly_ai_budget (default to 0 if the Console doesn't provide the value)

Changes: https://github.com/BLSQ/openhexa-app/compare/071ff82c0db1a409ad0e54efaddd8df3c6a810a7..05b92b4e55cf31e4a6d89e6bfda511c773f1b3bd

yolanfery (approved):
👌

PR #1889, opened 2026-06-26
feat: Data Studio frontend - 1st iteration (HEXA-1669)

mrivar:
Data Studio, added route, page, editor, results, run button, max rows, export csv

## Changes

- Backend: Added calculation of query duration to show in frontend
- Created Data Studio page with its UI
- Shows tables and table columns
- Can query
- Can export output into CSV

Some notes:
- Queries showed in DOM (in visible UI) capped to 500 rows. If we show too many it can start being sluggish and buggy. On export everything is added tho. Future development can include some sort of lazy-loading of rows to end up showing all of it more efficiently.
- Table columns are lazy-loaded to avoid N+1 queries in very big databases (drawback of this being that we cannot search for columns atm - can also be proposed for future development)

TODO in future developments:
- Saved queries
- AI generation
- search inlcuding table columns?
- lazy-loading or improved loading of result rows?

## Screenshots / screencast

image


frontend/src/workspaces/layouts/WorkspaceLayout/Sidebar.tsx
```diff
@@ -8,6 +8,7 @@ import {
   ChevronRightIcon,
   ChevronUpIcon,
   CircleStackIcon,
+  CommandLineIcon,
```
bramj: I think something more data-related would be better for this

e.g. "table-cells" (https://heroicons.com/)

backend/hexa/databases/utils.py
```diff
@@ -146,13 +148,15 @@ def execute_database_query(
...
             fetched = cursor.fetchmany(max_rows + 1) if cursor.description else []
+            duration_ms = round((time.perf_counter() - started_at) * 1000)
         truncated = len(fetched) > max_rows
         rows = json.loads(json.dumps(fetched[:max_rows], cls=ResultJSONEncoder))
         return {
             "columns": columns,
             "rows": rows,
             "row_count": len(rows),
             "truncated": truncated,
+            "duration_ms": duration_ms,
```
bramj, line 177: @DimitriKwihangana FYI I think you also need this functionality for the audit log, just to keep in mind

frontend/src/core/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,61 +1,157 @@
...
+// Design-system syntax palette: pink keywords, emerald strings, amber numbers.
+// Shared default for the embedded look; split into its own prop if a consumer
+// ever needs the embedded layout with different colors.
+const embeddedHighlightStyle = HighlightStyle.define([
+  { tag: tags.keyword, color: "#db2777", fontWeight: "500" },
+  { tag: [tags.string, tags.special(tags.string)], color: "#047857" },
+  { tag: [tags.number, tags.bool, tags.null], color: "#b45309" },
+  { tag: tags.comment, color: "#9ca3af", fontStyle: "italic" },
+  { tag: [tags.operator, tags.punctuation], color: "#6b7280" },
+  { tag: [tags.function(tags.variableName), tags.labelName], color: "#7c3aed" },
```
bramj: Beautiful! I wonder if we have these colours somewhere to reuse them and make sure we always have the same "OpenHEXA pink" etc?

frontend/src/core/components/CodeEditor/CodeEditor.tsx
```diff
@@ -1,61 +1,157 @@
...
+  ".cm-scroller": { minHeight: "100%" },
+  ".cm-content": { minHeight: "100%" },
+  ".cm-gutters": {
+    backgroundColor: "transparent",
+    border: "none",
+    color: "#d1d5db",
+  },
+  ".cm-lineNumbers .cm-gutterElement": { padding: "0 1rem 0 1.25rem" },
+  ".cm-activeLineGutter": { backgroundColor: "transparent", color: "#9ca3af" },
+  "&.cm-focused .cm-activeLine": { backgroundColor: "rgba(21, 93, 251, 0.03)" },
```
bramj: Not a huge fan of the focus line, but not against it either. I do have a very small glitch where I don't see the right border:

Image

backend/hexa/databases/graphql/schema.graphql
```diff
@@ -82,6 +82,11 @@ type ExecuteSQLResult {
   Whether the result was truncated because it exceeded the maximum number of rows.
   """
   truncated: Boolean
+
+  """
+  The server-side execution time of the query, in milliseconds.
+  """
+  durationMs: Int
```
bramj, line 92: Very cool 👍

backend/hexa/databases/utils.py
```diff
@@ -146,13 +148,15 @@ def execute_database_query(
             )
             # Fetch one extra row to detect (without returning) that more exist.
             fetched = cursor.fetchmany(max_rows + 1) if cursor.description else []
+            duration_ms = round((time.perf_counter() - started_at) * 1000)
```
bramj, line 169: 👍

frontend/src/pages/workspaces/[workspaceSlug]/data-studio/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+              href={`/workspaces/${encodeURIComponent(
+                workspace.slug,
+              )}/data-studio`}
+            >
+              {t("Data Studio")}
+            </Breadcrumbs.Part>
+          </Breadcrumbs>
+        }
+      >
+        {/* Full-bleed IDE: fill the viewport below the fixed 4rem workspace header. */}
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/data-studio/index.tsx:47]: That's great, we should do the same thing for pipeline and webapp code views 👍

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
```diff
@@ -0,0 +1,181 @@
...
+    if (!result?.success) {
+      return;
+    }
+    const csv = buildCsv(result.columns ?? [], result.rows ?? []);
+    downloadCsv("query-results.csv", csv);
+  };
+
+  const onKeyDown = (event: KeyboardEvent<HTMLDivElement>) => {
+    if ((event.metaKey || event.ctrlKey) && event.key === "Enter") {
+      event.preventDefault();
```
bramj: Not sure why, but for me (in Firefox) it doesn't prevent the line break. The ctrl + enter works great, and I really like it, but it also add a newline.

frontend/src/core/components/SplitButton/SplitButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+              </MenuItem>
+            );
+          })}
+        </MenuItems>
+      </Menu>
+    </div>
+  );
+};
+
+export default SplitButton;
```
bramj: I realize we should have updated the mockups with this, but in de end we decided to only allow 1 executed query at a time, so the "Run all" option no longer makes sense.

Sorry about this, but I think we can simplify and launch with just a "Run" button.

To go a bit further: the UX I think would be really great is that when I hit `CTRL + Enter`, it executes:
- If there's a selection: the selected query
- If there's no selection: the query where my cursor is currently positioned

This would allow me to write several queries in the file + execute them easily, without being bitten by the "only query at a time" error from the backend.
WDYT?

frontend/src/workspaces/features/DataStudioEditor/DataStudioResults.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+type DataStudioResultsProps = {
+  loading: boolean;
+  result?: ExecuteSqlResult;
+};
+
+// The grid is a preview: rendering tens of thousands of rows as live DOM is slow,
+// and users scan rather than scroll the full set. The complete result stays
+// available through CSV export.
+const MAX_DISPLAYED_ROWS = 500;
```
bramj, line 29: I like this approach, perhaps we won't need to do pagination at all because of this. And 500 still render smoothly for me. Good call!

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
```diff
@@ -0,0 +1,181 @@
...
+  const { t } = useTranslation();
+  const [query, setQuery] = useState("");
+  const [maxRows, setMaxRows] = useState(MAX_ROWS_OPTIONS[0]);
+  const editorRef = useRef<CodeEditorHandle>(null);
+
+  // Results are large, ad-hoc, and never read from the cache elsewhere; skip
+  // normalisation so big result sets are not retained for the page lifetime.
+  const [execute, { data, loading }] = useExecuteWorkspaceSqlLazyQuery({
+    fetchPolicy: "no-cache",
+  });
```
bramj: Maybe we should add some graceful error handling here for real errors that could come from the backend, or networks issues?

frontend/src/workspaces/features/DataStudioEditor/csv.ts
```diff
@@ -0,0 +1,40 @@
+import { downloadBlob } from "core/helpers/files";
+import { stringifyCellValue } from "./format";
+
+// Leading characters that spreadsheet apps interpret as the start of a formula.
+// Prefixing these values with a single quote neutralises CSV injection.
+const FORMULA_PREFIX = /^[=+\-@\t\r]/;
+
+const escapeCsvField = (raw: unknown): string => {
+  const text = stringifyCellValue(raw);
+  // Numbers are safe and must stay intact (e.g. "-5" should not become "'-5").
```
bramj, line 10: Not sure how this works, but I do get `'-5.5` for example for negative decimal numbers.

bramj (changes requested):
@mrivar This is really awesome! 😁  Looking super polished already on v1, very nice! 👏 

Only minor comments actually, the whole thing works really well and the UX feels really nice.

Just a few minor "general" comments:
- WDYT autofocus on the "Query" pane when you arrive on the page?
- Small indentation issue on EXPLAIN queries (see screenshots below)
- For the CSV export: This is fine for a first version, but I'm sure at some point we'll need to support downloading the full table, regardless of the row limit set. Not sure yet what that will look like, so let's postpone that for now.

Example EXPLAIN query:
Image

Same query in DBeaver:
Image

mcrimi:
Hi @mrivar - this is looking suuuuper cool, congratulations.

Sharing this problem with the tutorial database (`covid_data`) which predates this change, but that this change makes more easy to see:

When you create a workspace with tutorial content:
CleanShot 2026-07-06 at 10 32 37

You cannot read the tutorial table:
CleanShot 2026-07-06 at 10 19 48

 
## What seems to be the probem
When the workspace was created, we set up the read-only user and ran GRANT SELECT ON ALL TABLES but at that moment the `covid_data` table did not existed yet.

Then sample data is loaded (`demo.sql`), which created `covid_data`. That load runs as the admin DB user, not the workspace user. Ownership of `covid_data` was transferred to the workspace user, but no SELECT grant was given to the read-only user.


UPDATE: Manually granted the permission and able to test the feature.

mcrimi:
This is great. I would only suggest left value alignment instead of right, to prevent cases where selecting only a few or one columns looks weird:

CleanShot 2026-07-06 at 10 49 55

Checked GCP SQL Studio and that seems to be the pattern there:

CleanShot 2026-07-06 at 10 52 03

bramj:
@mrivar Well spotted on the bug @mcrimi ! 

I think this issue is sufficiently out of scope of this PR, we can tackle it in a different ticket: https://bluesquare.atlassian.net/browse/HEXA-1734

bramj:
@mrivar FYI this is how I think the CSV export should behave: https://bluesquare.atlassian.net/browse/HEXA-1712

We can absolutely go with your current implementation and tackle this ticket later.

frontend/src/workspaces/layouts/WorkspaceLayout/Sidebar.tsx
mrivar: Not sure about what do I think about this (no strong opinion), but I asked Claude out of curiosity and it pointed out this: 

> Context that matters: Data Studio sits right between Database (CircleStackIcon) and Datasets (Square2StackIcon) in the sidebar. So whatever icon we pick has to be distinguishable from two data-grid-ish neighbors.
>  The trade-off:
>  - CommandLineIcon (>_) reads as "terminal / write commands." It actually fits the authoring side of Data Studio well — you write SQL queries. Downside: it can feel developer-y/intimidating, and doesn't hint at the tabular results.
>  - TableCellsIcon (grid) reads as "spreadsheet / tabular data." It emphasizes the output. Downside: it's visually and conceptually close to the adjacent Database and Datasets items — three data-container icons in a row risks looking repetitive/confusing.

I don't really have a strong opinion on this tbh, but it's true that if the icons feel too similar it is a bit harder to distinguish them, wdyt?

bramj: Good points. The "too developer-y" is why I feel the `CommandLineIcon` is not a great fit, but no strong opinion on what it _should_ be.

Any thoughts on this @mcrimi ?

mcrimi: @bramj @mrivar - I also don't have strong feelings about it. Let's go for `TableCellsIcon`.

mrivar: Perfect, thanks for the feedback!

frontend/src/core/components/CodeEditor/CodeEditor.tsx
mrivar: Yes! Just added the brand colors (pink in this case) in globals.css as the single source of truth (can be used in tailwind and non-tailwind contexts). As a follow-up (in a separate PR/task) we can migrate the existing pink-* classes across the app to bg-brand so they all resolve there too, but for now this works for this context.

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
mrivar: good one! adding it

frontend/src/workspaces/features/DataStudioEditor/csv.ts
mrivar: good catch! that is indeed a bug, I'll check for other possible problems too

mrivar:
@bramj @mcrimi thanks for the feedback! The Data Studio is becoming quite good for a first iteration, although I'm afraid the PR grew too big by now 😅 I'll try to keep it into smaller ones next time 😂

Some things changed since last time based on your feedback:
- Changed the Data Studio icon to `TableCellIcon`
- Added autofocus on arriving to Data Studio
- Improved focus border
- Added PostgreSQL dialect highlights (keywords such as explain, analyze, etc)
- Refactored brand theme colors
- Removed `Run all` and `Run selection` in favor of a single `Run` button that already runs selections if there's any
- Improved visibility of Ctrl+Enter feature
- Better backend/network error handling

Bugs:
- Fixed indentation bug of `explain` queries / query plans
- Fixed Ctrl+Enter adding a new line
- Fixed csv export of negative decimal numbers
- Fixed right alignment on number columns

Side quest:
- tables schema browser - removed the 100 table cap and lazy loading of columns in favor of a more efficient 1-query load without any lazy loading

bramj (approved):
@mrivar Sorry for the wait on the review

Honestly, I've just played with it again and it is just sublime. Love it ❤️ , great work! 👏

PR #1890, opened 2026-06-26
fix: Pipeline AI new proposal was not pending

bramj:
When getting a new pipeline version proposal from the agent, the proposal was not marked as "pending".
This means that if you refresh the page, the proposal would be gone. The changes were in the conversation history however so the agent would still see them.


mrivar (approved):
Good catch!
Also really nice use of the `_on_tool_result` function

PR #1891, opened 2026-06-26
chore: Improve the loading states on the agent chat panel

bramj:
There were several minor UI issue remaining with the edit pipeline agent:
- Sometimes text would stop mid-sentence without any loading indication
- Loading states were missing up until tool uses

## Changes

- Always indicate a busy state while the stream is still ongoing, with a debounce to avoid "flickering"
- Added a "thinking" indicator to make UX a bit nicer
- Fix in `WordDrain` that would swallow the last word until a tool completed
- Small fix: Only fire Typewriter effect on new conversation (The effect also triggered on the first message of an existing conversation.)
- Small UX change: Remove disable of the chat input to keep autofocus during back and forth chat: The sending is still blocked during agent streaming, but this way you don't have to click back to the input on every message. (I also think it's useful to be able to start typing your next message while you're waiting for the agent to finish.)

## Screenshots / screencast

[ai-thinking2-2026-06-26_17.02.19.webm](https://github.com/user-attachments/assets/3f57d1d6-af48-4626-8ba0-34b3db3a25d1)


frontend/src/assistant/features/ChatPane/index.tsx
```diff
@@ -277,11 +320,32 @@ export default function ChatPane({
...
+    const container = scrollContainerRef.current;
+    if (container && isPinnedToBottom.current) {
+      container.scrollTop = container.scrollHeight;
     }
-  }, [messages.length, pendingUserMessage, streamingText]);
+  }, []);
+
+  useEffect(() => {
+    followBottom();
+  });
```
mrivar, line 311: useEffect with no dependency array runs on every render. Minor, not a big efficiency drawback but worth noting and checking if it can be improved or it is aboslutely needed.

mrivar (approved):
Great! The agent keeps looking more and more polished, I like that 😄

Just 2 small comments:
- Most of the PR is prettier formatter which is a bit of noise, minor but it might pollute git blame and makes the PR a bit harder to read and analyse
- ChatPane keeps getting growing immensely, it'd be worth it to take some time to see if it can be refactored so it doesn't become a "super component" with too much responsibility (doesn't have to be in this PR, but we should keep it in mind)

bramj:
> Most of the PR is prettier formatter which is a bit of noise, minor but it might pollute git blame and makes the PR a bit harder to read and analyse

You're right, sorry about that. I've just rebased and cleaned up actually, removing the prettier formatting.

> ChatPane keeps getting growing immensely, it'd be worth it to take some time to see if it can be refactored so it doesn't become a "super component" with too much responsibility (doesn't have to be in this PR, but we should keep it in mind)

Indeed, the component was already huge and it keeps growing. I really want to get this out of the door now, but let's refactor this one the next time we're working it.

PR #1892, opened 2026-06-29
feat: expose web app's underlying git server to pull push directly (HEXA-1542)

yolanfery:
Expose in the web app's UI a link to the underlying git server.

I decided to go with GCM with a Gitea API imitation instead of a CLI command since this allows to offload the browser client sign in to GCM which is pre installed on Windows with Git.

## Changes

- Mimic Gitea's OAuth so that users using GCM are brought to OH's UI for auth
- A 1-year token (because of a refresh issue in GCM https://github.com/git-ecosystem/git-credential-manager/issues/1408) is generated
- A `nginx` proxy is configured to ask OH if the token is authorized or not to query `Forgejo` based on org/workspace membership
- Enable branch protection so that history can't be re-written

## How to test
- Set the new env vars : 
    ```
    GIT_PROXY_USERNAME=openhexa-proxy                                                                                                                                                                                                                                                                                 
    GIT_PROXY_PASSWORD=openhexa-proxy
   ```
- Run the migrations
- Run the command `docker compose run --rm app manage sync_git_repositories`
- Run the app with the git proxy profile `docker compose --profile git_proxy up`
- Follow instruction to clone a repo and do operations on it 

## Screenshots / screencast

<img width="2227" height="1069" alt="Screenshot 2026-06-26 at 13 51 01" src="https://github.com/user-attachments/assets/63751653-2ad5-4078-913a-d4430251f459" />

<img width="774" height="222" alt="Screenshot 2026-06-29 at 14 49 32" src="https://github.com/user-attachments/assets/8e826e44-9944-4846-8f55-d33dadc0c49f" />


backend/config/settings/base.py
```diff
@@ -732,19 +732,44 @@ def _load_oidc_providers() -> list[dict]:
 GIT_SERVER_URL = os.environ.get("GIT_SERVER_URL", "http://forgejo:3000")
 GIT_SERVER_ADMIN_USERNAME = os.environ.get("GIT_SERVER_ADMIN_USERNAME")
 GIT_SERVER_ADMIN_PASSWORD = os.environ.get("GIT_SERVER_ADMIN_PASSWORD")
+GIT_PROXY_USERNAME = os.environ.get("GIT_PROXY_USERNAME", "openhexa-proxy")
```
yolanfery: Using a separated user for the proxy to reduce blast radius if there is an issue with the authorization endpoint

frontend/src/webapps/features/GitClonePopover/GitClonePopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    href="https://github.com/git-ecosystem/git-credential-manager/releases"
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="text-blue-600 hover:underline"
+                  >
+                    {t("releases page")}
+                  </a>
+                  {t(", then run:")}
+                </p>
+                <CommandBox command="sudo dpkg -i gcm-linux_amd64.*.deb && git-credential-manager configure" />
```
yolanfery: Not a Linux user, not sure what's the standard/easy way of installing this kind of tools

backend/config/settings/base.py
```diff
@@ -732,19 +732,44 @@ def _load_oidc_providers() -> list[dict]:
 GIT_SERVER_URL = os.environ.get("GIT_SERVER_URL", "http://forgejo:3000")
 GIT_SERVER_ADMIN_USERNAME = os.environ.get("GIT_SERVER_ADMIN_USERNAME")
 GIT_SERVER_ADMIN_PASSWORD = os.environ.get("GIT_SERVER_ADMIN_PASSWORD")
+GIT_PROXY_USERNAME = os.environ.get("GIT_PROXY_USERNAME")
+GIT_PROXY_PASSWORD = os.environ.get("GIT_PROXY_PASSWORD")
+GIT_PUBLIC_URL = os.environ.get("GIT_PUBLIC_URL", "http://localhost:3200")
```
bramj, line 748: Does this mean we'll need to use a different URL than `git.openhexa.org` on production?

backend/config/settings/base.py
```diff
@@ -732,19 +732,44 @@ def _load_oidc_providers() -> list[dict]:
...
+
+# Public OAuth2 clients used by git credential helpers. The Gitea client id is
+# the universal one baked into Git Credential Manager's Gitea provider: when the
+# proxy advertises `WWW-Authenticate: Basic realm="Gitea"`, GCM auto-discovers
+# OpenHEXA and signs in with no client configuration. The generic client serves
+# manual setups (git-credential-oauth). Both Applications are created in a
+# hexa.git data migration.
+GITEA_OAUTH_CLIENT_ID = "e90ee53c-94e2-48ac-9358-a874fb9e0662"
+GIT_OAUTH_CLIENT_ID = os.environ.get("GIT_OAUTH_CLIENT_ID", "openhexa-git")
+GIT_OAUTH_CLIENT_IDS = {GITEA_OAUTH_CLIENT_ID, GIT_OAUTH_CLIENT_ID}
```
bramj: So just to be clear: GCM offers support for Gitea by settings `Basic realm="Gitea"`?

What do you think about keeping naming on our end to `Forgejo` and maybe add a comment on the link to `Gitea` to avoid confusion there?

backend/hexa/oauth/views.py
```diff
@@ -148,21 +160,28 @@ class OAuthAuthorizeView(AuthorizationView):
...
+            scope = request.POST.get("scope") or request.GET.get("scope") or ""
+        tools = mcp_tools() if MCP_SCOPE in scope.split() else []
         html = render_to_string(
             "oauth2_provider/authorized.html",
             {"redirect_uri": redirect_to, "tools": tools},
         )
         return HttpResponse(html)
+
+
+def gitea_authorize(request: HttpRequest) -> HttpResponse:
```
bramj: Is there a good reason why we don't name it as `forgejo` internally?

docker-compose.yaml
```diff
@@ -192,6 +194,28 @@ services:
     networks:
       - openhexa
 
+  # Local equivalent of the production git proxy
+  git_proxy:
+    image: nginx:alpine
+    profiles:
+      - git_proxy
```
bramj: This is such a light-weight addition and I feel it'll just bite us when developing locally, what do you think about just launching it by default?

forgejo/entrypoint.sh
```diff
@@ -23,4 +23,18 @@ else
     echo "Admin user '$GIT_SERVER_ADMIN_USERNAME' already exists."
 fi
 
+# Non-admin service account the git proxy forwards git traffic as. It is added
+# as a write collaborator per repo, so branch protection applies to its pushes.
```
bramj, line 27: I feel this is a balanced choice between security and manageability 👍

docker/git-proxy/git-proxy.conf.template
```diff
@@ -0,0 +1,50 @@
...
+        proxy_pass http://$app_upstream$request_uri;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+
+    # Everything else is git: authorize against OpenHEXA, then forward to Forgejo.
+    location / {
+        auth_request /_git_authorize;
```
bramj, line 20: This is so cool, perfect for our needs

backend/hexa/git/mixins.py
```diff
@@ -31,6 +32,11 @@ def create_repo(self, *, files: list[dict] | None = None, user: User) -> str:
             self.client.create_org_repository(
                 self.git_org.slug, self.repository, auto_init=not files
             )
+            self.client.protect_branch(self.git_org.slug, self.repository)
+            if settings.GIT_PROXY_USERNAME:
+                self.client.add_collaborator(
+                    self.git_org.slug, self.repository, settings.GIT_PROXY_USERNAME
+                )
```
bramj, line 39: Not sure why we should check for `GIT_PROXY_USERNAME` here? Feels unnecessarily defensive.

backend/hexa/git/forgejo.py
```diff
@@ -71,6 +77,29 @@ def create_org_repository(
...
+        response = self._request(
+            "POST",
+            f"/repos/{org_slug}/{repo_name}/branch_protections",
+            json={
+                "rule_name": branch,
+                "enable_push": True,
+                "block_admin_merge_override": True,
+            },
+        )
+        return response.json()
```
bramj, line 92: Good call!

backend/hexa/git/tests/test_commands.py
```diff
@@ -0,0 +1,95 @@
...
+from hexa.core.test import TestCase
+from hexa.git.forgejo import ForgejoAPIError
+from hexa.user_management.models import User
+from hexa.webapps.models import GitWebapp, Webapp
+from hexa.workspaces.models import Workspace
+
+PROXY_USER = "openhexa-proxy"
+
+
+class SyncGitRepositoriesCommandTest(TestCase):
```
bramj, line 15: Thorough!

backend/hexa/git/views.py
```diff
@@ -0,0 +1,143 @@
...
+    except GitWebapp.DoesNotExist:
+        return HttpResponse(status=403)
+
+    if webapp.git_org.slug != org:
+        return HttpResponse(status=403)
+
+    if operation is Operation.WRITE and not user.has_perm(
+        "webapps.update_webapp", webapp
+    ):
+        return HttpResponse(status=403)
```
bramj, line 141: All this is so cool 😮

frontend/src/webapps/features/GitClonePopover/GitClonePopover.tsx
bramj: I think this makes sense, I'll double-check on my desktop tomorrow just to be sure

bramj (approved):
I'm very impressed by this @yolanfery , nice work! 👏 

It's a massive feature, that I'm sure will be greatly appreciated by our users. The implementation is solid, only minor comments.

This is actually exactly as I hoped it would work: no need to keep users and roles synced with Forgejo, and have OH core handle the permissions.
I'm really happy to see that this was possible in the end! 😁 I feel like despite the LOC count, the general workings are fairly easy to reason about.

backend/config/settings/base.py
yolanfery: I din't think about this, good catch. 

I am proposing to configure the infra to point `git.openhexa.org` to the new proxy (this will be the domain used by users to clone/pull/push) and create a new record (maybe `forgejo.openhexa.org` ?) to point at our forgejo instance that we keep using internally but still accessible if we need to perform admin actions on it. What do you think ?

backend/config/settings/base.py
yolanfery: Yes GCM is supporting Gitea. Using Forgejo naming would be more misleading imo

backend/hexa/oauth/views.py
yolanfery: To make it clear that we are mimicking gitea and not implementing for Forgejo

But maybe it is not clear enough 😅 maybe I should try some renames like `authorize_like_gitea_would`,...

What do you think ?

docker-compose.yaml
yolanfery: Makes sense, changed

backend/hexa/git/mixins.py
yolanfery: To not block localhosting with this, I think we should keep it

backend/config/settings/base.py
bramj: @yolanfery Sounds good! :+1:

backend/hexa/git/mixins.py
bramj: Fair enough

backend/hexa/oauth/views.py
bramj: @yolanfery Hm I disagree actually.

Sure we're mimicking Gitea towards GCM, but we're definitely implementing for Forgejo.
So for us I think `forgejo_authorize` is more correct and easier to understand/maintain towards the future.

For me the only reference to Gitea that really makes sense is `Basic realm="Gitea"`, to satisfy GCM. Apart from that I think it would just add confusion to the code base.

What do you think, happy to discuss over a call!

bramj:
> I think this makes sense, I'll double-check on my desktop tomorrow just to be sure

@yolanfery I had some issues on Linux and took the liberty of adding a commit with a bit more install instructions.

backend/hexa/oauth/views.py
yolanfery: I still don't agree for the same reason but it's not a big deal anyway, I did the rename as you suggested and will move forward with the infra

yolanfery:
> > I think this makes sense, I'll double-check on my desktop tomorrow just to be sure
> 
> @yolanfery I had some issues on Linux and took the liberty of adding a commit with a bit more install instructions.

Thanks a lot

PR #1893, opened 2026-06-29
feat: mcp tools get dbschema and dbtable schema

DimitriKwihangana:
Adds two new MCP tools for inspecting a workspace's database schema:

## Changes
get_db_schema: lists all tables with their names and row counts
get_db_table_schem :returns column names and PostgreSQL data types for a specific table

### DIFF
--- a/backend/hexa/databases/tests/helpers.py
+++ b/backend/hexa/databases/tests/helpers.py
@@ -1,16 +1,23 @@
+from psycopg2 import sql
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
 from hexa.databases.utils import get_workspace_database_connection
 
 
-def seed_demo_table(workspace, rows):
-    """Create a `demo` table on the workspace database using the read-write role."""
+def seed_demo_table(workspace, rows, table_name="demo"):
+    """Create a demo table on the workspace database using the read-write role."""
     conn = get_workspace_database_connection(workspace)
     conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+    table = sql.Identifier(table_name)
     try:
         with conn.cursor() as cursor:
-            cursor.execute("DROP TABLE IF EXISTS demo;")
-            cursor.execute("CREATE TABLE demo (id int, label text);")
-            cursor.executemany("INSERT INTO demo (id, label) VALUES (%s, %s);", rows)
+            cursor.execute(sql.SQL("DROP TABLE IF EXISTS {};").format(table))
+            cursor.execute(
+                sql.SQL("CREATE TABLE {} (id int, label text);").format(table)
+            )
+            cursor.executemany(
+                sql.SQL("INSERT INTO {} (id, label) VALUES (%s, %s);").format(table),
+                rows,
+            )
     finally:
         conn.close()
--- a/backend/hexa/mcp/tests/test_databases.py
+++ b/backend/hexa/mcp/tests/test_databases.py
@@ -1,0 +1,112 @@
+from unittest.mock import patch
+
+from hexa.mcp.tools.databases import get_db_schema, get_db_table_schema
+
+from .testutils import MCPTestCase
+
+TABLES = [
+    {"name": "patients", "count": 1000, "workspace": object()},
+    {"name": "visits", "count": 5000, "workspace": object()},
+]
+
+PATIENTS_DEF = {
+    "name": "patients",
+    "count": 1000,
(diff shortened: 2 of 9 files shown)


mrivar (changes requested):
Good to see expansion on the mcp tools!
Very nice testing overall as well. However, I wonder if we should use the `execute_graphql` layer as the other mcp tools use for consistency.
This would also be benifitial as:
- It would allow to utilise code that is already there (we wouldn't have to maintain the same logic in 2 different places)
- Things like pagination are already implemented in the code, so we would skip having to implement it here


Also as a side quest, as you'd be going through this part of the code; currently our pagination works as follows:

```python
def resolve_database_tables(workspace, info, page=1, per_page=15, **kwargs):
      tables = get_database_definition(workspace=workspace)   # ← ALL tables, ALL COUNT(*) here
      return result_page(tables, page=page, per_page=per_page) # ← in-memory slice afterwards
```

`get_database_definition` runs SELECT COUNT(*) for every table in the workspace, then throws away all but one page.
It would be nice to explore an improvement in this resolver so we don't load the whole data and then slice it, but get directly the data we really want from the db (would be a relevant performance improvement for workspaces with a big amount of tables and an interesting challenge!)

backend/hexa/mcp/tests/test_databases.py
```diff
@@ -0,0 +1,110 @@
...
+        {"name": "dob", "type": "date"},
+    ],
+}
+
+
+class GetDbSchemaTest(MCPTestCase):
+    def test_get_db_schema(self):
+        with patch(
+            "hexa.mcp.tools.databases.get_database_definition", return_value=TABLES
+        ):
```
yolanfery: I think we can do the tests without patching : https://github.com/BLSQ/openhexa-app/blob/0f44a9710bc33c0e69bd9c854e896b1da8a0465d/backend/hexa/databases/tests/test_utils.py#L306-L325

yolanfery (changes requested):
Totally agreeing with @mrivar comments, it's actually the same comment as in https://github.com/BLSQ/openhexa-app/pull/1874#discussion_r3492587587

Let's try as much as possible to use the GQL layer first to reduce logic duplication, it will not always be possible but in this case it probably is

DimitriKwihangana:
> Good to see expansion on the mcp tools! Very nice testing overall as well. However, I wonder if we should use the `execute_graphql` layer as the other mcp tools use for consistency. This would also be benifitial as:
> 
> * It would allow to utilise code that is already there (we wouldn't have to maintain the same logic in 2 different places)
> * Things like pagination are already implemented in the code, so we would skip having to implement it here
> 
> Also as a side quest, as you'd be going through this part of the code; currently our pagination works as follows:
> 
> ```python
> def resolve_database_tables(workspace, info, page=1, per_page=15, **kwargs):
>       tables = get_database_definition(workspace=workspace)   # ← ALL tables, ALL COUNT(*) here
>       return result_page(tables, page=page, per_page=per_page) # ← in-memory slice afterwards
> ```
> 
> `get_database_definition` runs SELECT COUNT(*) for every table in the workspace, then throws away all but one page. It would be nice to explore an improvement in this resolver so we don't load the whole data and then slice it, but get directly the data we really want from the db (would be a relevant performance improvement for workspaces with a big amount of tables and an interesting challenge!)

Thanks for the comments @mrivar. I did try to refactor with the execute_graphql. On the side quest you mentioned is really good as well. Only that my first instinct was to do this in a separate ticket, but since I was already in this part of the code I took the liberty of including it here: database.tables now paginates in SQL. The MCP benefits since it uses the same resolver

backend/hexa/databases/tests/helpers.py
```diff
@@ -1,16 +1,23 @@
+from psycopg2 import sql
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
 from hexa.databases.utils import get_workspace_database_connection
 
 
-def seed_demo_table(workspace, rows):
-    """Create a `demo` table on the workspace database using the read-write role."""
+def seed_demo_table(workspace, rows, table_name="demo"):
```
yolanfery, line 7: Good addition :)

yolanfery (approved):

mrivar (approved):
This is great, thanks for the effort! 😁
Just as a comment, at some point in the future we might want to have the page structure as a dataclass for consistency (instead of the current dict structure that is in different parts of the code, i.e. the output of `core.graphql.result_page`), but that's out of the scope of this PR.

PR #1897, opened 2026-07-01
fix: mcp should not read binary webapps files (HEXA-1707)

yolanfery:
Through GQL, skip reading binary files for the MCP tool


DimitriKwihangana (approved):
This looks good!! I think It will help alot on the editing webapps witht the assistant

PR #1899, opened 2026-07-03
feat: s3 storage bucket configuration

mrivar:
Added S3 storage bucket configuration (CORS, lifecycle and versioning)

## Changes

- Added `WORKSPACE_BUCKET_VERSIONING_ENABLED` env variable (was already in codebase but not in env file). True by default
- Align `enable_versioning` config with gcp.py
- Added CORS config (`*` as in gcp)
- Set bucket versioning
- Set bucket objects lifecycle, mimicking GCP config:
  - `STANDARD_IA` @ 30 days
  - `GLACIER_IR` @ 90 days
  - `DEEP_ARCHIVE` @ 365 days
- Added bucket labels for consistency
- Added config tests

### DIFF
--- a/backend/hexa/files/backends/s3.py
+++ b/backend/hexa/files/backends/s3.py
@@ -148,13 +148,15 @@ class S3Storage(Storage):
         """Apply a post-creation bucket configuration step.
 
         S3-compatible stores (e.g. MinIO) do not implement all bucket
-        configuration APIs; skip those instead of failing bucket creation.
-        Any other error propagates, like a failed bucket.patch() on GCP.
+        configuration APIs; skip those (NotImplemented/MethodNotAllowed)
+        instead of failing bucket creation. Any other error propagates,
+        like a failed bucket.patch() on GCP.
         """
         try:
             step()
         except ClientError as e:
             if e.response["Error"]["Code"] in ("NotImplemented", "MethodNotAllowed"):
+                # Swallowed on purpose (methods not supported on S3-compatible stores)
                 sentry_sdk.capture_exception(e)
             else:
                 raise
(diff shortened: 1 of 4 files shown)


backend/hexa/files/backends/s3.py
```diff
@@ -134,6 +138,119 @@ def create_bucket(self, bucket_name: str, *args, **kwargs) -> str:
...
+                        "ExposeHeaders": ["ETag", "Content-Type", "Content-Length"],
+                        "MaxAgeSeconds": 3600,
+                    }
+                ]
+            },
+        )
+
+    def _set_bucket_versioning(self, bucket_name: str) -> None:
+        if not self.enable_versioning:
+            return
```
yolanfery, line 185: [Minor] I would probably check before calling

backend/hexa/files/backends/s3.py
```diff
@@ -123,7 +128,6 @@ def create_bucket(self, bucket_name: str, *args, **kwargs) -> str:
                     Bucket=bucket_name,
                     CreateBucketConfiguration={"LocationConstraint": self.region},
```
yolanfery, line 129: There is no way we can create the bucket with the correct options without having to modify it afterwards ?

yolanfery (approved):

backend/hexa/files/backends/s3.py
mrivar: Apparently, the create-then-configure pattern is the standard approach for S3. I also found it odd tbh

yolanfery: Ok, thanks

PR #1900, opened 2026-07-07
feat: add DatabaseQueryLog model for auditing SQL queries (HEXA-1668)

DimitriKwihangana:
Log all SQL queries executed against workspace databases via the executeSQL GraphQL field, for auditing and to enable a future query history component

## Changes

- New DatabaseQueryLog model + initial migration: workspace, user, query, status (SUCCESS / ERROR / REJECTED / DENIED), SQLSTATE result code, timestamp, duration, row count, origin.
- executeSQL resolver writes one log row on every exit path; logging is failure-safe (a failed audit write never breaks the query).
-  optional origin argument (API | DATA_STUDIO, defaults to API) so Data Studio queries can be distinguished later.
- Logs are permission-scoped via filter_for_user() (workspace members only).
- Composite indexes (workspace, -created_at) and (user, -created_at) to keep history-page queries fast as the table grows.


backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
     except QueryCanceled as e:
+        _log_executed_query(
+            request,
+            workspace,
+            query,
+            origin,
+            DatabaseQueryLog.Status.ERROR,
+            result_code=e.pgcode,
+            error_message=str(e).strip(),
+            duration_ms=int((time.monotonic() - started_at) * 1000),
```
mrivar: We can extract this into a tiny helper.
Actually if you merge from main there is a [4th instance of the same calculation](https://github.com/BLSQ/openhexa-app/pull/1889/changes#diff-30be73a24581d8b381d14463504dded900ace917bad3f0d66f30a02bbb870317R169), so we can change it to use the helper as well (might be worth exploring if there's a difference between `time.monotonic()` and `time.perf_counter()`)

backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
+        request,
+        workspace,
+        query,
+        origin,
+        DatabaseQueryLog.Status.SUCCESS,
+        result_code="00000",
+        duration_ms=int((time.monotonic() - started_at) * 1000),
+        row_count=result["row_count"],
+        truncated=result["truncated"],
+    )
```
mrivar: As we always end up doing `_log_executed_query`, this part can be improved by avoiding repeating the same (or similar) type of code by having a finally that absorbs the `_log_executed_query`. E.g. Something like:

```python
  status, log_fields = None, {}
  started_at = time.monotonic()
  try:
      ...
      result = execute_database_query(...)
      status = DatabaseQueryLog.Status.SUCCESS
      log_fields = {"result_code": "00000", "row_count": result["row_count"],
                    "truncated": result["truncated"], "duration_ms": _elapsed(started_at)}
      return {"success": True, "errors": [], **result}
  except MultipleStatementsError as e:
      status, log_fields = DatabaseQueryLog.Status.REJECTED, {"error_message": str(e)}
      return {...}
      ...
  finally:
      if status:
          _log_executed_query(request, workspace, query, origin, status, **log_fields)
```

backend/hexa/databases/models.py
```diff
@@ -0,0 +1,63 @@
...
+        "user_management.User",
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="database_query_logs",
+    )
+    query = models.TextField()
+    status = models.CharField(max_length=10, choices=Status.choices)
+    # SQLSTATE error code (https://www.postgresql.org/docs/current/errcodes-appendix.html);
+    # "00000" on success, null when the query never reached the database
+    result_code = models.CharField(max_length=5, null=True, blank=True)
```
mrivar: Well documented, didn't know about that! It would be nice to extract this "00000" success result error into a constant for better maintainability

backend/hexa/databases/models.py
```diff
@@ -0,0 +1,63 @@
...
+        # The query reached the database but failed (invalid SQL, timeout, ...)
+        ERROR = "ERROR"
+        # The user was not allowed to run queries against the workspace database
+        DENIED = "DENIED"
+        # The query was rejected before reaching the database (e.g. multiple statements)
+        REJECTED = "REJECTED"
+
+    class Origin(models.TextChoices):
+        API = "API"
+        DATA_STUDIO = "DATA_STUDIO"
```
mrivar, line 36: Right now `ExecuteSQLOrigin` and `DatabaseQueryLog.Origin` share the same values for contract purposes, which is needed, but it would be nice if we find a way so that they are always closely related. It would give us better maintainability (if a value is added/changed in one of the two, the contract breaks. But if they're closely coupled, changing a value in one them changes it in both).
You can get some inspiration in the codebase with `FileEncoding` and `file_encoding_enum`.

backend/hexa/databases/graphql/schema.graphql
```diff
@@ -16,7 +16,23 @@ type Database {
   # 'executeSQL' runs a read-only SQL query against the workspace database and returns the result.
   # 'maxRows' caps the number of returned rows; it defaults to 50 when omitted and is itself
   # capped to a server-side hard limit.
-  executeSQL(query: String!, maxRows: Int): ExecuteSQLResult!
+  # 'origin' identifies where the query comes from for auditing purposes; it defaults to API.
+  executeSQL(query: String!, maxRows: Int, origin: ExecuteSQLOrigin): ExecuteSQLResult!
```
mrivar, line 23: Given we already merged https://github.com/BLSQ/openhexa-app/pull/1889/, it is worth it to update the frontend to already use `ExecuteSQLOrigin`, to make sure we start logging the 'DATA_STUDIO' queries correctly from the beginning (instead of being tagged as API by default)

backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
+    query: str,
+    origin: str,
+    status: str,
+    **fields,
+):
+    # Auditing must never break query execution, hence the broad except.
+    try:
+        DatabaseQueryLog.objects.create(
+            workspace=workspace,
+            user=request.user if isinstance(request.user, User) else None,
```
mrivar: Minor: Currently if a request user doesn't have a proper `User` instance, we ignore and save None, losing track of it (right now this is not reachable, so it is not a real risk, but it might happen if me change it in the future unknowingly).
We can mitigate this by detecting the user that don't have a direct user attached (PipelineRunUser / DAGRunUser) and getting their `real_user`.
It's more of a safety net than anything and an easy quick win, so I would vote to add it

mrivar (changes requested):
Very solid PR, with very extensive testing!
Just some code comments, mainly on improving maintainability and DRY / applying Solid principles, it just needs some fiddling to be prod ready 😄

@bramj Given that this new table is going to grow wih every query we do unbounded, with full query text (TextField, uncapped), it might be that it quickly becomes one of our biggest tables. I'd say it is worth to create a ticket to investigate pruning/retention strategies or TTL/archival/partitioning (out of scope from this task)

We might also want to look into whether queries sometimes have personal data and if so how to treat it/clean it (also out of scope from this task)

backend/hexa/databases/graphql/schema.graphql
```diff
@@ -16,7 +16,23 @@ type Database {
...
+}
+
+"""
+The origin of a SQL query, recorded for auditing and to build a per-user query history.
+"""
+enum ExecuteSQLOrigin {
+  """
+  The query was submitted directly through the API.
+  """
+  API
```
yolanfery: Not sure `API` is a great name, maybe we just need a `OTHER`

backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
+            query=query,
+            origin=origin,
+            status=status,
+            **fields,
+        )
+    except Exception:
+        logger.exception(
+            "Failed to log database query for workspace %s", workspace.slug
+        )
```
yolanfery: If we keep this : Should we add more details about the error and/or the context maybe ?

backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
         result = execute_database_query(workspace, query, **max_rows_kwarg)
     except MultipleStatementsError as e:
+        _log_executed_query(
+            request,
+            workspace,
+            query,
+            origin,
+            DatabaseQueryLog.Status.REJECTED,
+            error_message=str(e),
+        )
```
yolanfery: Those calls should probably live at a lower level for instance in `execute_database_query`

backend/hexa/databases/schema.py
```diff
@@ -62,34 +68,108 @@ def resolve_database_credentials(workspace: Workspace, info, **kwargs):
...
+    # Auditing must never break query execution, hence the broad except.
+    try:
+        DatabaseQueryLog.objects.create(
+            workspace=workspace,
+            user=request.user if isinstance(request.user, User) else None,
+            query=query,
+            origin=origin,
+            status=status,
+            **fields,
+        )
```
yolanfery: Since we store those logs in the database (and not a distant server), there is no reason to assume this can fail. So I would assume that if this raises that there is a bug and we shouldn't catch it

yolanfery (changes requested):

backend/hexa/databases/schema.py
DimitriKwihangana: Good catch! I will extract this into small helper which will be reusable

backend/hexa/databases/schema.py
DimitriKwihangana: You mean to also add user id and origin of the error. That's helpful

yolanfery: You can include the exception, this is a pattern we use int he codebase https://github.com/BLSQ/openhexa-app/blob/120f169d4f86d4341e9ede238d26d0c4b82a2af9/backend/hexa/core/admin.py#L70-L76

backend/hexa/databases/schema.py
DimitriKwihangana: About the coverage I think I would keep it at the resolver level because the DENIED case never reaches the 

`execute_database_query` and the function has no access to the request/user/origin that the log entry needs. Since the resolver is the only caller today, the API boundary covers everything. I've added a not in docstring so that a second caller wont miss it. let me know what you think.

yolanfery: The new version looks already better but I think we should enclose the logic in one function to avoid repeating ourself and avoid dealing with logging logic in the resolver layer : 
we can have a wrapper of `execute_database_query` that accepts the correct args (the one required for the query execution and the one for the logging), catch exceptions, logs accordingly and raise the exception to the resolver.

yolanfery: The goal is to maintain a single point of entry for executing SQL from the resolver layer. It's easier to test, maintain and not forget to call the log entry creation if we add a new exception type handling to the resolver

DimitriKwihangana: Good point! I did implement that. Let me know if it aligns.

yolanfery (approved):
Looking great, good job !

mrivar (approved):
Love it! Looking good 😄

PR #1905, opened 2026-07-08
feat: Add Django admin SQL dashboards (HEXA-1738)

bramj:
Simple, pragmatic SQL dashboards served straight from Django.

https://bluesquare.atlassian.net/browse/HEXA-1738

Notes:
- We use a read-only DB user + tight statement_timeout for security and performance reasons
- Dashboards edit rights will belong only to the OH superadmin roles, viewer rights will be granted to specific persons
- This may feel weird, but it’s actually just a tailored view-only version of the Django admin

New ENV vars to add to your `.env`:
```env
# Read-only user for Django dashboards
DATABASE_USER_READ_ONLY=hexa-app
DATABASE_PASSWORD_READ_ONLY=hexa-app
```

## Changes

- Add django-sql-dashboard to Django requirements, wire it up with a read-only Postgres role
- Small fix to only require OH permission system for `hexa` Django apps
- Add a custom Django admin view for "Groups": Groups are a part of the standard Django permission system and
django-sql-dashboard relies on it. They were deliberately hidden in our admin for obvious reasons, now we
bring them back in a simplified fashion with a clear disclaimer about what it's for. A bit more code than I liked, but this allows us to give a specific user access to a specific dashboard.

## How/what to test

Create a dashboard via the Django admin and play around with the view/edit permissions. Attempt to access it when logged in as a different user to see if the permissions are respected.

## Screenshots / screencast

It's pretty cool:

### DIFF
--- a/backend/hexa/user_management/admin.py
+++ b/backend/hexa/user_management/admin.py
@@ -2,6 +2,7 @@ from datetime import timedelta
 
 from django import forms
 from django.contrib import admin, messages
+from django.contrib.admin.widgets import FilteredSelectMultiple
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.forms import PasswordResetForm
 from django.contrib.auth.forms import UserCreationForm as BaseUserCreationForm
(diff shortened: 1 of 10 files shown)


backend/hexa/user_management/backends.py
```diff
@@ -39,9 +42,7 @@ def has_perm(
 
         permission_module = self._get_permission_module(app_label)
         if permission_module is None:
-            raise ValueError(
-                f'Could not find a permission module for the "{app_label}" app'
-            )
+            return False  # Not an OpenHEXA app, let other backends handle it
```
mrivar, line 45: nice one!
We could add some tests for this so we don't regress to it in the future

backend/hexa/user_management/admin.py
```diff
@@ -32,10 +33,47 @@
...
+@admin.register(Group)
+class GroupAdmin(admin.ModelAdmin):
+    form = GroupForm
+    list_display = ("name", "members_count")
+    search_fields = ("name",)
+
+    def members_count(self, obj):
+        return obj.user_set.count()
+
+    members_count.short_description = "Members Count"
```
mrivar: Minor proposal:

With the current implementation, members_count is calculated per row (N+1 query), which is not really a problem in here, as this is always going to be a small table only for superusers; so it will never be a slow query anyway.
However, making it a single query (see the code below) allows us to sort the table by members count as well.
Minor improvement but it is a quick and easy win, but I'll leave it to your judgement.

```python
    def get_queryset(self, request):
        return super().get_queryset(request).annotate(_members_count=Count("user", distinct=True))

    @admin.display(description="Members Count", ordering="_members_count")
    def members_count(self, obj):
        return obj._members_count
```

mrivar (changes requested):
Very nice to have this! I tested a couple of dashboards and it works very nicely 😄

I only have a request for tests for the new perm check for non-hexa apps, and a minor comment that you cna take or leave

backend/hexa/user_management/admin.py
bramj: This will indeed always be a small table, but it's a good practice :+1:

mrivar (approved):
nice testing! LGTM

PR #1907, opened 2026-07-08
chore: Enable GZip compression on Django endpoints

bramj:
A suggestion from @madewulf :

This commit enables gzip compression on everything served from Django.

This is huge, it impacts:
- `/graphql/`
- `/mcp`
- Served web apps
- the Django admin

This means we can expect a 3–10x reduction on size of text-based payloads on all API calls and served web apps.

### DIFF
--- a/backend/config/settings/base.py
+++ b/backend/config/settings/base.py
@@ -350,9 +350,11 @@ INSTALLED_APPS = [
 
 MIDDLEWARE = [
     "django.middleware.security.SecurityMiddleware",
-    # Enable gzip compression on everything served from Django. This impacts
-    # /graphql/, /mcp, all web apps and the Django admin.
-    "django.middleware.gzip.GZipMiddleware",
+    # Enable gzip compression on everything served from Django except SSE
+    # streams. This impacts /graphql/, /mcp, all web apps and the Django admin.
+    # Note: SSE issues appears resolved in Django 6.0, to check when we upgrade:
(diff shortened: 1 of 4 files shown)


yolanfery (approved):
Great catch @madewulf  🤯

Curious if Sentry will handle it seamlessly, let's see

bramj:
Note: Important addition to the PR (last commit)

```
Adapt the GZip middleware to exclude the streaming endpoints.

On streaming endpoints, the Django `GZipMiddleware` compresses chunks as
individual files, rather than as a continuous stream.
As a result, only the first chunk is decompressible, and browsers
drop the rest of the response.

There appears a fix for this in Django 6, but for now, we'll just
deactivate it for streaming endpoints.

More info:
https://code.djangoproject.com/ticket/36656
```

yolanfery:
> Note: Important addition to the PR (last commit)
> 
> ```
> Adapt the GZip middleware to exclude the streaming endpoints.
> 
> On streaming endpoints, the Django `GZipMiddleware` compresses chunks as
> individual files, rather than as a continuous stream.
> As a result, only the first chunk is decompressible, and browsers
> drop the rest of the response.
> 
> There appears a fix for this in Django 6, but for now, we'll just
> deactivate it for streaming endpoints.
> 
> More info:
> https://code.djangoproject.com/ticket/36656
> ```

Good catch and I agree with the approach

yolanfery (approved):

PR #1909, opened 2026-07-10
fix: get static webapp file int cannot represent non integer value (HEXA-1739)

yolanfery:
Union of types were not exposed correctly. Solve the issue globally.

Note : this is leaving out more complex type system like Unions of Unions, but let's keep the logic simple until we need more complex types

## Changes

- Fix the type system in the protocol
- Unit tests

### DIFF
--- a/backend/hexa/mcp/protocol.py
+++ b/backend/hexa/mcp/protocol.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import logging
+import typing
 
 from hexa.mcp.models import ToolCall
 
@@ -18,6 +19,14 @@ def tool(func):
     return func
 
 
+def _base_type(annotation):
+    """Unwrap Optional/Union annotations (e.g. `int | None`) to the underlying type."""
+    args = [a for a in typing.get_args(annotation) if a is not type(None)]
+    if len(args) == 1:
+        return args[0]
+    return annotation
+
+
 def _get_tool_schema(func):
     sig = inspect.signature(func)
     properties = {}
@@ -27,7 +36,9 @@ def _get_tool_schema(func):
     for name, param in sig.parameters.items():
         if name == "user":
             continue
-        properties[name] = {"type": type_map.get(param.annotation, "string")}
+        properties[name] = {
+            "type": type_map.get(_base_type(param.annotation), "string")
+        }
         if param.default is inspect.Parameter.empty:
             required.append(name)
 
--- a/backend/hexa/mcp/protocol.py
+++ b/backend/hexa/mcp/protocol.py
@@ -1,6 +1,8 @@
+import enum
 import inspect
 import json
 import logging
+import types
 import typing
 
 from hexa.mcp.models import ToolCall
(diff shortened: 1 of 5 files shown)


backend/hexa/mcp/protocol.py
```diff
@@ -18,6 +19,14 @@ def tool(func):
     return func
 
 
+def _base_type(annotation):
+    """Unwrap Optional/Union annotations (e.g. `int | None`) to the underlying type."""
+    args = [a for a in typing.get_args(annotation) if a is not type(None)]
+    if len(args) == 1:
+        return args[0]
+    return annotation
```
mrivar, line 33: Nice one!! However I think I found a change in the behaviour of list types. It is not a problem atm as we don't have any param that has a type list, but it is worth consciously deciding which behaviour we want. If we want to keep the old one we just have to add a guard for non-Union types.
These are the changed behaviours:

  
| annotation        | OLD (main) | NEW (PR) | changed? |
|-------------------|------------|----------|----------|
| list[str]         | string     | string   | no       |
| list[int]         | string     | **integer**  | ✅ yes    |
| list[float]       | string     | **number**   | ✅ yes    |
| list[bool]        | string     | **boolean**  | ✅ yes    |
| list[str] \| None | string     | string   | no       |
| dict[str,int]     | string     | string   | no     |

mrivar (changes requested):
Nice fix! I like that the mcp is getting more polished 😄
Mainly one lil change in behaviour worth noting, and it would be nice adding tests regarding it.

On a nice polish follow-up, although this is out of the scope of this PR and can be discusssed, we can improve the `_get_tool_schema` by adding an `enum` type; such that for values coming fom TextChoices or IntEnum etc, the LLM already knows it's restricted values and is less likely to suggest a non-valid one (that would then fail against the graphQL layer and restart the process). Claude proposed this function:

```python
def _property_schema(annotation):
    base = _base_type(annotation)
    if isinstance(base, type) and issubclass(base, enum.Enum):
        values = [member.value for member in base]
        value_type = type(values[0]) if values else str
        return {"type": TYPE_MAP.get(value_type, "string"), "enum": values}
     return {"type": TYPE_MAP.get(base, "string")}
```

It would end up with:
```python
PipelineFunctionalType | None  -> {"type": "string", "enum": ["extraction","transformation","loading","computation"]}
Priority (IntEnum)             -> {"type": "integer", "enum": [1, 2]}
```

I leave it up to you if you think if it's worth it or if we can skip it :)

backend/hexa/mcp/protocol.py
yolanfery: Yes `list[]` were not correctly supported, thanks, I fixed and test it to make it more consistent

yolanfery:
> On a nice polish follow-up, although this is out of the scope of this PR and can be discusssed, we can improve the `_get_tool_schema` by adding an `enum` type; such that for values coming fom TextChoices or IntEnum etc, the LLM already knows it's restricted values and is less likely to suggest a non-valid one (that would then fail against the graphQL layer and restart the process). Claude proposed this function:
> 
> ```python
> def _property_schema(annotation):
>     base = _base_type(annotation)
>     if isinstance(base, type) and issubclass(base, enum.Enum):
>         values = [member.value for member in base]
>         value_type = type(values[0]) if values else str
>         return {"type": TYPE_MAP.get(value_type, "string"), "enum": values}
>      return {"type": TYPE_MAP.get(base, "string")}
> ```
> 
> It would end up with:
> 
> ```python
> PipelineFunctionalType | None  -> {"type": "string", "enum": ["extraction","transformation","loading","computation"]}
> Priority (IntEnum)             -> {"type": "integer", "enum": [1, 2]}
> ```
> 
> I leave it up to you if you think if it's worth it or if we can skip it :)

Thanks, it's a great addition. I thought it was out of scope but we have some tools that could already benefit from it (PipelineFunctionalType, webapp's allow operations,...) so I decided to follow your recommendation ☺️

Added and tested

PR #1910, opened 2026-07-10
feat: added banner on sso error (HEXA-1606)

mrivar:
Added a banner when there is an SSO error, so it doesn't fail silently which would be annoying for users


bramj (approved):
Nice and thorough! 👍

PR #1914, opened 2026-07-15
fix: workspaces should always be created inside an organization (HEXA-1729)

yolanfery:
We have not enforced workspace to have an organization but it's now time to do it to simplify the codebase.

## Changes

- Change the model to enforce organization for each workspace
- Migration (and a unit test) to assign orphan workspace to a default org created during the migration
- Adapt the BE and FE with the new mandatory constraint
- Update unit tests by introducing a new `create_workspace` wrapper to add an organization flawlessly if not provided


backend/hexa/user_management/models.py
```diff
@@ -332,6 +330,11 @@ class Meta:
     objects = OrganizationManager.from_queryset(OrganizationQuerySet)()
     all_objects = IncludeSoftDeletedManager.from_queryset(OrganizationQuerySet)()
 
+    def save(self, *args, **kwargs):
+        if not self.slug:
+            self.slug = create_organization_slug(self.name)
+        super().save(*args, **kwargs)
```
mrivar, line 336: nice refactor!

backend/hexa/workspaces/models.py
```diff
@@ -292,10 +290,8 @@ def db_ro_url(self):
         return f"postgresql://{self.db_ro_username}:{self.db_ro_password}@{self.db_host}:{self.db_port}/{self.db_name}"
 
     @property
-    def current_subscription(self) -> OrganizationSubscription | None:
-        if self.organization:
-            return self.organization.current_subscription
-        return None
+    def current_subscription(self) -> OrganizationSubscription:
+        return self.organization.current_subscription
```
mrivar: I think the `OrganizationSubscription` can still be none, can't it?

backend/hexa/workspaces/schema/types.py
mrivar: Not sure if we still need this, I think it has been removed from all places in the FE, wdyt?
Also on a side note, if we decide not to remove, we should rethink how to do it, as rn it would send organization as None which would raise a `TypeError`

frontend/src/pages/workspaces/index.tsx
```diff
@@ -97,21 +93,7 @@ const WorkspacesHome = (props: WorkspacesHomeProps) => {
...
             </Link>
           )}
-          {me.permissions.createWorkspace && (
-            <button
-              onClick={() => setDialogOpen(true)}
-              className="mt-6 inline-flex items-center rounded-md bg-blue-600 px-4 py-2 text-sm font-semibold text-white shadow-sm hover:bg-blue-500"
-            >
-              {t("Create a workspace")}
-            </button>
-          )}
```
mrivar, line 107: Given that we've removed this, how is an organization created in a, let's say, new environment or self-hosted instance? Not debating the change, just wondering what the expected route is

backend/hexa/workspaces/tests/testutils.py
```diff
@@ -0,0 +1,25 @@
...
+    through ``create_if_has_perm`` (which also provisions the database/bucket, so
+    callers usually patch those); omit it for a plain ``create``.
+    """
+    if organization is None:
+        organization = Organization.objects.create(name=f"{name} Organization")
+    if principal is not None:
+        return Workspace.objects.create_if_has_perm(
+            principal, name=name, organization=organization, **kwargs
+        )
+    return Workspace.objects.create(name=name, organization=organization, **kwargs)
```
mrivar, line 25: This function is repeated 3 times, I guess we can keep one of them and reuse it given they do practically the same

mrivar (dismissed):
Amazing PR! Really big and important change, mostly some minor comments, but it is practically perfect

backend/hexa/workspaces/models.py
yolanfery: Good catch, thanks, fixed

backend/hexa/workspaces/schema/types.py
yolanfery: Yes we can remove it from the schema, this PR is a breaking change already soit makes sense to do it now, thanks

backend/hexa/workspaces/tests/testutils.py
yolanfery: Improved, thanks

frontend/src/pages/workspaces/index.tsx
yolanfery: Superusers, could use the admin interface. But I now added the button in the UI and resolver to create organizations form the UI as superusers, its a small addition that makes it more natural

PR #1915, opened 2026-07-16
fix: generate token should work for org admin/superusers (HEXA-1747)

yolanfery:
Right now org admin and super users can't generate a token. Because there is no explicit workspace membership.

## Changes

- Object oriented approach by introducing new Token objects
- Sign some Token based on if the explicit membership exists
- The middleware parses the token using the method in the objects
- Add unit tests to cover the new abstraction and some edge case in this flow


backend/hexa/workspaces/authentication.py
```diff
@@ -0,0 +1,103 @@
...
+            .filter(id=payload.get("workspace_id"))
+            .first()
+        )
+        return cls(user, workspace) if workspace else None
+
+
+_TOKEN_TYPES_BY_PAYLOAD = {
+    dict: IdentityToken,
+    str: MembershipToken,
+}
```
mrivar: this is clever, but I feel we could improve it by adding some sort of identifier in the token, in case this expands in the future we don't have to do magic not to break it, wdyt?

backend/hexa/workspaces/authentication.py
```diff
@@ -0,0 +1,103 @@
...
+    def payload(self) -> str | dict:
+        """The object to sign for this token."""
+
+    @classmethod
+    @abc.abstractmethod
+    def from_payload(cls, payload: str | dict) -> "WorkspaceToken | None":
+        """Rebuild a token from a signed payload, or ``None`` if it is not ours."""
+
+    def sign(self) -> str:
+        return Signer().sign_object(self.payload())
```
mrivar, line 32: Given that now we give out an identity token with superadmin powers, it would be sensible to add some expiry time using TimestampSigner and max_age. We could keep a None max_age for MembershipToken's tho, to maintain the current functionality and add it only to Identity tokens.

backend/hexa/workspaces/middlewares.py
mrivar, line 11: Addressing this FIXME is more relevant after this change. Not necessarily the scope of this PR but it would be nice to create a follow-up task to address it soonish

backend/hexa/workspaces/authentication.py
```diff
@@ -0,0 +1,103 @@
...
+    def authenticate(cls, raw_token: str) -> "WorkspaceToken | None":
+        """Resolve a signed token string, or ``None`` if it is invalid or revoked."""
+        try:
+            payload = Signer().unsign_object(raw_token)
+        except (UnicodeDecodeError, binascii.Error, BadSignature):
+            return None
+
+        # The payload type identifies the token: dict -> identity, str -> membership.
+        token_type = _TOKEN_TYPES_BY_PAYLOAD.get(type(payload))
+        return token_type.from_payload(payload) if token_type else None
```
mrivar: Currently we don't check is a user `is_active`. It would be a "new logic", but I feel it would be the correct security implementation. We would avoid deactivated-but-not-deleted users to still be able to authenticate. wdyt?

mrivar (dismissed):
Nice one!
I have a few security/design suggestions, and 2 follow-ups that we can just throw into separate Jira tasks.
Mainly:
1.  `_TOKEN_TYPES_BY_PAYLOAD`: we can explore a better logical separations in case we have more token implementations in the future, unless you really think it's not worth it
2. Exploring expiry times for IdentityToken's given that they're overpowered tokens now, and might become dangerous if leaked
3. `is_active`: revoking access to deactivated accounts

As follow-ups:
1. Addressing the FIXME in the authentication middleware, so the tokens only have access to the given workspaces. More relevant now with the IdentityToken's.
2. `workspaces/views.py:78-88`: Divergent startegy for the same problem. In the notebook/credentials path we already had an strategy for this, by creating a WorkspaceMembership on the fly for admins without WorkspaceMembership. Might be worth it to refactor it to use the same logic as here.

yolanfery:
Thanks for the review @mrivar 

I agree with most of the comments, it needs re-work, I will improve it after my holidays

backend/hexa/workspaces/authentication.py
yolanfery: Improved it a bit, thanks

backend/hexa/workspaces/authentication.py
yolanfery: Added a 24H expiry for the IdentityToken, good idea until we fix the FIXME

backend/hexa/workspaces/middlewares.py
yolanfery: Added a ticket https://bluesquare.atlassian.net/browse/HEXA-1775

No idea at this point how we could implement this without doing massive changes to `filter_for_user` and permission checks. Maybe we have no way around this and just need to take the leap

backend/hexa/workspaces/authentication.py
yolanfery: Accounted for this

yolanfery:
> Nice one! I have a few security/design suggestions, and 2 follow-ups that we can just throw into separate Jira tasks. Mainly:
> 
> 1. `_TOKEN_TYPES_BY_PAYLOAD`: we can explore a better logical separations in case we have more token implementations in the future, unless you really think it's not worth it
> 2. Exploring expiry times for IdentityToken's given that they're overpowered tokens now, and might become dangerous if leaked
> 3. `is_active`: revoking access to deactivated accounts
> 
> As follow-ups:
> 
> 1. Addressing the FIXME in the authentication middleware, so the tokens only have access to the given workspaces. More relevant now with the IdentityToken's.
> 2. `workspaces/views.py:78-88`: Divergent startegy for the same problem. In the notebook/credentials path we already had an strategy for this, by creating a WorkspaceMembership on the fly for admins without WorkspaceMembership. Might be worth it to refactor it to use the same logic as here.

Addressed the 5 topics, looking better now, thanks

Will do more extensive E2E tests in demo

PR #1916, opened 2026-07-17
fix: read only db user has no select grant on tutorial covid data table (HEXA-1734)

yolanfery:
We were not granting permissions on the tutorial table to the read-only role. 

Longer term fix by using the RW role to seed the database

### DIFF
--- a/backend/hexa/workspaces/migrations/0061_grant_select_on_covid_data_to_ro_role.py
+++ b/backend/hexa/workspaces/migrations/0061_grant_select_on_covid_data_to_ro_role.py
@@ -5,19 +5,14 @@ from psycopg2 import sql
 from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
 
 
-def get_admin_connection(db_name):
-    """Get a connection to a specific database using the admin role."""
-    host = settings.WORKSPACES_DATABASE_HOST
-    port = settings.WORKSPACES_DATABASE_PORT
-    admin_role = settings.WORKSPACES_DATABASE_ROLE
-    admin_password = settings.WORKSPACES_DATABASE_PASSWORD
-
+def get_rw_connection(workspace):
+    """Connect to a workspace database as its read-write role."""
     conn = psycopg2.connect(
-        host=host,
-        port=port,
-        dbname=db_name,
-        user=admin_role,
-        password=admin_password,
+        host=settings.WORKSPACES_DATABASE_HOST,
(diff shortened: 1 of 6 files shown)


backend/hexa/workspaces/migrations/0061_grant_select_on_covid_data_to_ro_role.py
```diff
@@ -0,0 +1,71 @@
...
+    """Connect to a workspace database as its read-write role."""
+    conn = psycopg2.connect(
+        host=settings.WORKSPACES_DATABASE_HOST,
+        port=settings.WORKSPACES_DATABASE_PORT,
+        dbname=workspace.db_name,
+        user=workspace.db_name,
+        password=workspace.db_password,
+    )
+    conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+    return conn
```
mrivar, line 18: as in migration 0060, should we use `get_admin_connection` instead, in order not to silently skip workspace where the RW role doesn't own covid_data? wdyt?

backend/hexa/workspaces/migrations/0061_grant_select_on_covid_data_to_ro_role.py
```diff
@@ -0,0 +1,71 @@
...
+
+def grant_select_on_covid_data(apps, schema_editor):
+    """
+    Back-fill the read-only role's SELECT grant on the tutorial covid_data table.
+    """
+    Workspace = apps.get_model("workspaces", "Workspace")
+
+    granted = 0
+    skipped = 0
+    for workspace in Workspace.objects.filter(db_ro_password__isnull=False):
```
mrivar: Minor: as db_ro_password is already `null=False`, shouldn't we just use `.all()`? or is it for readability/extra safeguarding?

mrivar (commented):
Nice fix! Minor comments only, nothing blocking

bramj:
@mrivar Could you take over this PR to avoid it staying open for too long? Feel free to add the improvements you suggested if you'd like and then we can merge.

backend/hexa/workspaces/migrations/0061_grant_select_on_covid_data_to_ro_role.py
mrivar: Rejected and kept `get_rw_connection`. Admin doesn't ensure being the owner of the covid_table

yolanfery:
Thanks @mrivar for improving the logging and merging

PR #1918, opened 2026-07-20
feat: data studio csv download (HEXA-1712)

mrivar:
Improve the data studio CSV export to support full query data download

## Changes

- Implemented 2 download paths:
  - **Happy path**: Already implemented path we had. Frontend gets the data returned by the query and transforms it into a CSV, to download directly. Fast and cheap (saves going to backend and DB again for the data). Only used for queries in which we know we have all the data.
  - **Bigger queries path**: When queries are bigger (we know this as that information comes) we go to the backend and DB to execute the query again without truncating the data, and then stream the file directly to the disk. Implemented some extra helpers to show an error if something happens before the download.
- Added a harness test to make sure both paths generate the exact same CSV files for the same data, for consistency.

### Possible follow-up
~~The main issue with the bigger queries path is that mid-stream errors are not shown, so if the stream fails in the middle of it, the user gets a truncated csv and no download error in the UI. This can be mitigated by 2 different follow-ups to consider:~~
- ~~**Change from streaming to buffer-then-stream**. This is the usual practice for huge files (eg how MEGA lets you donwload GB size files, or Google Drive). The main issue is that it leaves the user waiting while the backend prepares the file. To consider, but we have to take into account timeouts and first analyse how big the files downloaded are (if they're usually small it might not be worth it). Downside is that it holds memory space while creating the file, which might be a problem for big queries.~~
- ~~**Gzip compression**: If we don't go for the buffer-then-stream approach, then we can use gzip (.zip/.csv.gz) so if the file is truncated, the user notices~~

EDIT: We already have the gzip benefit for free thanks to `SSEAwareGZipMiddleware`, so we might not need a follow-up at all

- Duplicate column names in a result (SELECT a.id, b.id …) — RealDictCursor collapses same-named keys, so the CSV would repeat one value. This is pre-existing (affects the interactive path identically)

## How/what to test

- Generate queries (big and small) and export them


backend/hexa/core/tests/fixtures/csv_cell_vectors.json
DimitriKwihangana, line 1: Nice pattern! First time seeing this. Very clean

backend/hexa/data_studio/views.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+        # A successful attachment never navigates the iframe the frontend posts into,
+        # so the page can't see the download start. Signal it with a short-lived,
+        # JS-readable cookie whose *name* carries the caller's token (per-token, so
+        # concurrent downloads don't clobber each other). Signals "began", not
+        # "completed". See the app README and frontend downloadQueryCsv.ts.
+        download_token = request.POST.get("download_token")
+        if download_token and _DOWNLOAD_TOKEN_RE.fullmatch(download_token):
+            response.set_cookie(
```
DimitriKwihangana, line 198: Minor: I think adding the Secure=True would be a good idea to avoid unencrypted connection. Not mentioning there is a risk but it is already used in the app.

backend/hexa/data_studio/views.py
```diff
@@ -0,0 +1,160 @@
...
+_DOWNLOAD_TOKEN_RE = re.compile(r"[A-Za-z0-9-]{1,64}")
+
+# Per-process admission control (settings.DATA_STUDIO_EXPORT_MAX_CONCURRENCY): each
+# export holds a read-only connection open for the whole download, so an unbounded
+# number would pile up connections and threads in one worker. Excess callers get a 429
+# immediately rather than queueing behind a backlog. See the app README.
+_EXPORT_SLOTS = threading.BoundedSemaphore(settings.DATA_STUDIO_EXPORT_MAX_CONCURRENCY)
+
+
+def _next_projected_batch(batch_iter, columns):
```
DimitriKwihangana, line 50: Nice

backend/config/urls.py
```diff
@@ -49,6 +49,7 @@
     path("pipelines/", include("hexa.pipelines.urls", namespace="pipelines")),
     path("workspaces/", include("hexa.workspaces.urls", namespace="workspaces")),
     path("files/", include("hexa.files.urls", namespace="files")),
+    path("databases/", include("hexa.databases.urls", namespace="databases")),
```
bramj: I think this should belong to the `data_studio`

backend/hexa/core/tests/fixtures/csv_cell_vectors.json
```diff
@@ -0,0 +1,42 @@
+{
+  "_comment": "Canonical CSV serialisation contract shared by the backend (hexa.core.csv) and the frontend (DataStudioEditor/csv.ts). `vectors` pins how a single cell serialises (post-guard, post-quote); `rowVectors` pins whole-record shape (columns + rows -> full CSV, no BOM), which cell vectors cannot cover because each is wrapped in a fixed two-column row. Both export paths are asserted against this same file so the client path (small results) and server path (large results) cannot drift. Only JSON-representable values live here: type-specific branches that never cross the GraphQL wire (Python bytes/Decimal/datetime) are covered by backend-only tests. Numbers beyond the JS safe range are represented as strings, mirroring how NUMERIC/DECIMAL/bigint actually reach the frontend.",
```
bramj: This is great! 👏

backend/hexa/core/tests/test_csv.py
```diff
@@ -1,49 +1,193 @@
...
+        # as "" (not a bare line), and the frontend buildCsv now matches that, so
+        # a single-column result is byte-identical across both export paths.
+        for vector in CSV_ROW_VECTORS:
+            with self.subTest(vector=vector["description"]):
+                writer = csv.writer(Echo())
+                columns = vector["columns"]
+                output = _csv_line(writer, columns)
+                for row in vector["rows"]:
+                    output += _csv_line(writer, [row[column] for column in columns])
+                self.assertEqual(vector["expected"], output)
```
bramj: Awesome tests 👍

backend/hexa/core/csv.py
```diff
@@ -1,39 +1,214 @@
...
+from contextlib import aclosing
+from decimal import Decimal
 
-from django.db.models import QuerySet
-from django.http import HttpResponse
+from django.http import StreamingHttpResponse
 
+# Excel decodes a UTF-8 CSV as the local ANSI codepage unless it sees a BOM,
+# which mangles accented characters. Prepend one so non-ASCII data survives.
+UTF8_BOM = "\ufeff"
```
bramj, line 14: TIL: BOM is not just Bill Of Materials 😅

backend/hexa/core/middlewares.py
```diff
@@ -23,6 +23,14 @@ class SSEAwareGZipMiddleware(GZipMiddleware):
...
         if response.get("Content-Type", "").startswith("text/event-stream"):
             return response
+        # An async streaming response is gzipped one gzip member per chunk
+        # (django/middleware/gzip.py compresses each chunk via compress_string),
+        # which browsers can't reliably stream-decode — the same failure this
+        # class already avoids for SSE, fixed upstream in Django 6.0. The Data
+        # Studio CSV export is served as an async stream, so leave every async
+        # stream uncompressed until the upgrade.
+        if getattr(response, "is_async", False) and response.streaming:
+            return response
```
bramj, line 33: 👍

DimitriKwihangana (changes requested):
Great implementation @mrivar . The logic looks good only the small change of adding a path of "/database/". in file `frontend/server/index.mjs`

backend/hexa/databases/README.md
```diff
@@ -0,0 +1,143 @@
...
+  `DOWNLOAD_QUERY_IDLE_TIMEOUT_MS` aborts the transaction only if the client stops
+  consuming mid-stream — by which point the user's own download has visibly stalled
+  anyway, so a truncated file is not a silent surprise.
+
+That leaves genuine infra events as the realistic causes: the workspace DB connection
+dropping (restart, failover, network blip) or the web worker being killed mid-stream
+(deploy, OOM, or a scale-down outlasting gunicorn's graceful-shutdown window). These
+are infrequent and usually visible through other signals (deploy notices, error
+rates), not only through a short CSV.
+
```
bramj: This section feels too verbose, WDYT?

backend/hexa/databases/utils.py
```diff
@@ -193,6 +193,80 @@ def execute_database_query(
...
 
+# The full result set makes the interactive per-request timeout far too short; this
+# generous ceiling still guards a runaway scan from pinning a connection.
+DOWNLOAD_QUERY_TIMEOUT_MS = 5 * 60 * 1000
+# statement_timeout bounds a single FETCH; between batches the transaction sits idle
+# while the client consumes the previous one. This caps that idle time so a stalled or
+# vanished client can't pin the connection (and block VACUUM) indefinitely.
+DOWNLOAD_QUERY_IDLE_TIMEOUT_MS = 5 * 60 * 1000
+# Server-side cursor batch size, bounding peak memory regardless of result size.
+DOWNLOAD_QUERY_BATCH_SIZE = 2_000
```
bramj, line 176: 👍

backend/hexa/data_studio/views.py
```diff
@@ -0,0 +1,160 @@
...
+# The token is echoed back inside a Set-Cookie *name*, so restrict it to characters
+# safe there. A token failing this is ignored (the frontend then times out rather
+# than misreading another download's signal).
+_DOWNLOAD_TOKEN_RE = re.compile(r"[A-Za-z0-9-]{1,64}")
+
+# Per-process admission control (settings.DATA_STUDIO_EXPORT_MAX_CONCURRENCY): each
+# export holds a read-only connection open for the whole download, so an unbounded
+# number would pile up connections and threads in one worker. Excess callers get a 429
+# immediately rather than queueing behind a backlog. See the app README.
+_EXPORT_SLOTS = threading.BoundedSemaphore(settings.DATA_STUDIO_EXPORT_MAX_CONCURRENCY)
```
bramj, line 47: I feel like this is bit much for a v1, but that sure is future-proof 🙂

backend/hexa/databases/utils.py
```diff
@@ -193,6 +193,80 @@ def execute_database_query(
...
+            )
+            setup_cursor.execute(
+                sql.SQL(
+                    "SET LOCAL idle_in_transaction_session_timeout = {timeout};"
+                ).format(timeout=sql.Literal(idle_timeout_ms))
+            )
+        cursor = conn.cursor(
+            name="data_studio_csv_download", cursor_factory=RealDictCursor
+        )
+        cursor.itersize = batch_size
```
bramj: Is this used? You set the batch_size explicitly later on line 249.

bramj (changes requested):
I just downloaded 200K records in a matter of seconds on my local machine, cool! This is working very smooth 👍 

Two points:
- Like we discussed this morning, I think the backend endpoints should be moved to the `data_studio` Django app.
- My bigger issue with the approach: I think the double-path support adds too much complexity for a relatively low benefit.

It adds complexity such as:
- `_format_float_like_js` in `csv.py`
- the shared fixtures and tests that ensure parity (plus github wiring)
- the frontend logic in `csv.ts` (and its test file)
- ...

I think the "fast" path does not add enough value and I'd argue it's better for maintainability and code understanding to only use the async path. For small files, this will be very fast as well.

backend/hexa/core/csv.py
mrivar: hahah I didn't even know Bill Of Materials 😂

backend/hexa/databases/utils.py
mrivar: good call! not needed indeed

mrivar:
@bramj @DimitriKwihangana 
Thanks for the reviews! I finally did:
- Refactored to data_studio in this PR already. Only kept in hexa/databases/ the `stream_database_query` function, as it is the only "pure" database logic and fits better there than in the data_studio
- Removed the fast path in favor of the stream backend path (means more db calls and backend calls, but should be fine unless exporting csvs becomes a performance issue)
- Added secure cookie and data_studio/ path in express server

.github/workflows/frontend-ci.yml
```diff
@@ -5,11 +5,19 @@ on:
...
+      - 'backend/hexa/data_studio/tests/fixtures/download_contract.json'
   push:
     branches: [main]
     paths:
       - 'frontend/**'
       - 'backend/**/schema.graphql'
+      # Shared fixtures read cross-tier by the frontend suite (csv.parity.test.ts /
+      # downloadQueryCsv.test.ts), so a change to them must re-run frontend CI too.
+      - 'backend/hexa/core/tests/fixtures/csv_cell_vectors.json'
+      - 'backend/hexa/data_studio/tests/fixtures/download_contract.json'
```
bramj, line 18: I think these shared fixtures can be removed now, no?

backend/hexa/core/csv.py
```diff
@@ -1,39 +1,165 @@
...
+        finally:
+            super().close()
+
+
+def async_streaming_csv_response(
+    *,
+    header: typing.Sequence[typing.Any],
+    row_batches: typing.AsyncIterator[typing.Iterable[typing.Sequence[typing.Any]]],
+    filename: str,
+    with_bom: bool = True,
```
bramj: Is there a reason we want to make this configurable? I don't think it's used with `False`

backend/hexa/data_studio/tests/fixtures/download_contract.json
```diff
@@ -0,0 +1,8 @@
+{
+  "_comment": "Cross-tier contract for the Data Studio CSV download handshake, shared by the backend view (hexa/data_studio/views.py) and the frontend (DataStudioEditor/downloadQueryCsv.ts). Both sides assert their real implementation against this single file (backend: test_views.py; frontend: downloadQueryCsv.test.ts) so a rename on one tier fails a test instead of silently breaking the handshake. `cookiePrefix` is the critical one: a successful attachment is handed straight to the browser's download manager without firing the iframe's load event, so the ONLY 'download began' signal is the backend setting a cookie named `${cookiePrefix}${token}` and the frontend polling for that exact name. A silent mismatch there would not error — the download would appear to hang until the frontend's timeout. `fields` are the POST field names of the same request; a mismatch there fails loudly (HTTP 400) but is bound here too since it is part of the same contract.",
+  "cookiePrefix": "csvDownloadToken-",
+  "fields": {
+    "query": "query",
+    "token": "download_token"
+  }
+}
```
bramj, line 8: Still needed?

frontend/src/workspaces/features/DataStudioEditor/downloadQueryCsv.ts
```diff
@@ -0,0 +1,132 @@
...
+// iframe and the page untouched; a failed response lands in the iframe instead
+// of navigating the whole app away.
+const DOWNLOAD_FRAME_NAME = "data-studio-csv-download-frame";
+
+// The backend signals "download began" by setting a short-lived cookie whose NAME is
+// this prefix plus our per-download token — our only positive signal, since a
+// successful download never fires the iframe's load event. A per-token name lets
+// concurrent downloads each wait on their own signal without clobbering one another.
+// Part of the cross-tier download contract (backend tests/fixtures/download_contract.json),
+// asserted on both sides so a rename can't silently break the handshake.
```
bramj, line 14: Still needed?

frontend/src/workspaces/features/DataStudioEditor/downloadQueryCsv.test.ts
```diff
@@ -0,0 +1,175 @@
...
+  it("rejects when the download never starts (timeout)", async () => {
+    (getCookie as jest.Mock).mockReturnValue(undefined);
+    const promise = downloadQueryCsv("ws-1", "SELECT 1");
+    const assertion = expect(promise).rejects.toThrow("timed out");
+
+    jest.advanceTimersByTime(10 * 60 * 1000 + 1000);
+    await assertion;
+  });
+
+  describe("stays in sync with the shared cross-tier contract", () => {
```
bramj, line 158: Still needed?

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
```diff
@@ -128,13 +128,29 @@ const DataStudioEditor = ({
                   ))}
                 </select>
               </label>
+              {/* The server re-run can take a while; reassure the user right
+                  where they clicked. */}
+              {exporting && (
+                <span className="text-xs text-gray-400">
+                  {t("This may take a while")}
```
bramj, line 233: Nice, didn't notice this earlier 👍

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.test.tsx
```diff
@@ -230,15 +235,47 @@ describe("DataStudioEditor", () => {
     });
   });
 
-  it("exports the current result to CSV", async () => {
-    mockQueryState = successState();
+  it("exports the last run query from the server, not the current editor contents", async () => {
+    // A complete (not truncated) result still streams from the server — there is
+    // no longer a client-side path.
```
bramj: This feels overkill?

backend/hexa/databases/utils.py
```diff
@@ -193,85 +190,78 @@ def execute_database_query(
...
+# while the client consumes the previous one. This caps that idle time so a stalled or
+# vanished client can't pin the connection (and block VACUUM) indefinitely.
+DOWNLOAD_QUERY_IDLE_TIMEOUT_MS = 5 * 60 * 1000
+# Server-side cursor batch size, bounding peak memory regardless of result size.
+DOWNLOAD_QUERY_BATCH_SIZE = 2_000
 
 
-def run_and_log_database_query(
-    request: HttpRequest,
+def stream_database_query(
```
bramj, line 179: I just realized that we don't do audit logging of the query on this code path. Since here we're doing queries without limits on potentially very large tables, I think it's useful to add it here.
Would also be nice to somehow differentiate, perhaps we can set the `origin` to "Data Studio CSV export" or something like that?
WDYT?

bramj (changes requested):
Sorry to send this back again, but I think we're almost there! 💪

.github/workflows/frontend-ci.yml
mrivar: `csv_cell_vectors.json` can be removed, but `download_contract.json` is still useful to ensure there's not a silent cookie-name mismatch

backend/hexa/core/csv.py
mrivar: True, it's dead code, removing it

backend/hexa/data_studio/tests/fixtures/download_contract.json
mrivar: As mentioned above, still useful to ensure there's not a silent cookie-name mismatch

backend/hexa/databases/utils.py
mrivar: True, adding the audit logs

bramj (approved):
@mrivar Thorough job on the logging of the streamed queries! 👍  I didn't think about how involved it would be, but it makes absolute sense to do this async so we don't lose track of a stream that failed mid-way.

Apologies for how long this one took, okay to merge for me! 🚀

DimitriKwihangana (approved):

PR #1919, opened 2026-07-20
feat: agent that generate sql queries(HEXA-1697)

DimitriKwihangana:
Adds a new assistant agent that turns a natural-language request into a validated SQL query for the workspace database, with a generate entry.

## Changes

- GenerateSqlAgent (new generate_sql_agent.py), registered instruction set GENERATE_SQL.
- get_full_database_definition() — schema (tables + columns, no row counts) for instruction inlining.
- validate_query()  EXPLAIN-based validation. for validation purporses
- execute_sql MCP tool.
- Generate Sql query dialog

## Screenshots / screencast
image


backend/hexa/databases/utils.py
```diff
@@ -275,6 +275,30 @@ def run_and_log_database_query(
...
+        with conn.cursor() as cursor:
+            cursor.execute(
+                sql.SQL("SET LOCAL statement_timeout = {timeout};").format(
+                    timeout=sql.Literal(timeout_ms)
+                )
+            )
+            cursor.execute("EXPLAIN " + query)
+    finally:
+        if conn:
+            conn.close()
```
bramj, line 303: I think this is a great idea for a deterministic validation 👍 

@mcrimi Making evals for this SQL agent will actually be one of the easier ones, could be a good start!

backend/hexa/assistant/agents/generate_sql_agent.py
```diff
@@ -0,0 +1,119 @@
...
+                "## Workspace database schema\n"
+                "The workspace database contains no tables."
+            )
+        schema = _render_schema(tables)
+        if len(schema) > _SCHEMA_MAX_CHARS:
+            names = "\n".join(f"- {table['name']}" for table in tables)
+            return (
+                "## Workspace database schema\n"
+                "The schema is too large to show in full, so only table names are "
+                "listed. Use the `execute_sql` tool on `information_schema.columns` "
```
bramj: Why not the `get_db_schema` and `get_db_table_schema` tools?

backend/hexa/assistant/instructions.py
```diff
@@ -98,12 +99,25 @@ class InstructionSet(TextChoices):
...
 """
 
+_GENERATE_SQL = """
+# Your task
+You translate the user's natural-language request into a single PostgreSQL query against their workspace database.
+- The database schema is provided in your context below. When it only lists table names (large database) or when you need more detail, use the `execute_sql` tool to inspect `information_schema` or peek at sample rows.
+- Use the `execute_sql` tool to test your query and make sure it returns the expected result before answering. The workspace slug is injected automatically — do not pass it.
+- Your final answer MUST be the SQL statement and nothing else:
+  - A single read-only statement (`SELECT`, or `WITH ... SELECT`).
+  - No markdown fences, no commentary, no explanations before or after.
```
bramj: I suppose we can allow it to add SQL comments to clarify it's work? When it's a complex query I think it's useful to have some comments, for example:

```sql
-- This returns the monthly trend of indicators...
WITH monthly AS (
    SELECT
        disease,
        DATE_TRUNC('month', report_date) AS month,
        AVG(incidence_rate) AS avg_incidence_rate
    FROM disease_indicators
    WHERE report_date >= '2025-01-01' AND report_date < '2026-01-01'
    GROUP BY disease, DATE_TRUNC('month', report_date)
),
trend AS (
    SELECT
...
```

backend/hexa/databases/utils.py
```diff
@@ -429,6 +453,30 @@ def get_database_definition_page(
...
+    conn = None
+    try:
+        conn = get_workspace_database_connection(workspace)
+        with conn.cursor(cursor_factory=RealDictCursor) as cursor:
+            # LIMIT NULL means "no limit" in PostgreSQL, which lets us reuse
+            # the paginated query to fetch the full listing.
+            params = {"ignore_tables": IGNORE_TABLES, "limit": None, "offset": 0}
+            tables = _fetch_tables_with_columns(cursor, workspace, params)
+        for table in tables:
+            del table["count"]
```
bramj: Not sure what happens here, are the table counts calculated and then removed before returning the result?

On a more general note: I feel like we're getting more and more ways to do the general task of "getting workspace DB metadata", might be worth checking if there's a refactoring possible.

backend/hexa/mcp/graphql/databases.graphql
```diff
@@ -28,3 +28,20 @@ query GetDatabaseTable($workspaceSlug: String!, $tableName: String!) {
...
+        errorMessage
+        columns
+        rows
+        rowCount
+        truncated
+        durationMs
+      }
+    }
+  }
+}
```
bramj: Not sure why this query shows up here, we already had this no?

backend/hexa/mcp/tools/databases.py
```diff
@@ -25,6 +25,22 @@ def get_db_schema(user, workspace_slug: str, page: int = 1, per_page: int = 15)
...
+        user,
+        "ExecuteSQL",
+        {"workspaceSlug": workspace_slug, "query": query, "maxRows": max_rows},
+    )
+    if "errors" in data:
+        return data
+    workspace = data.get("workspace")
+    if workspace is None:
+        return {"error": "Workspace not found"}
+    return workspace["database"]["executeSQL"]
```
bramj: So this is exactly what we discussed about our responsible use of AI. This is a very powerful tool, and it will probably make results from the AI agent a lot better, but for now we don't want to allow agents access to the data itself.

frontend/src/pages/workspaces/[workspaceSlug]/data-studio/index.tsx
```diff
@@ -46,7 +50,12 @@ const WorkspaceDataStudioPage: NextPageWithLayout = (props: Props) => {
       >
         {/* Full-bleed IDE: fill the viewport below the fixed 4rem workspace header. */}
         <div className="h-[calc(100vh-4rem)] p-4">
-          <DataStudioEditor workspaceSlug={workspace.slug} />
+          <DataStudioEditor
+            workspaceSlug={workspace.slug}
+            aiEnabled={aiEnabled}
+            aiBudgetLimitReached={aiBudgetLimitReached}
+            monthlyLimitExceeded={monthlyLimitExceeded}
+          />
```
Comment (bramj) [inline frontend/src/pages/workspaces/[workspaceSlug]/data-studio/index.tsx:42]: 👍

frontend/src/workspaces/features/DataStudioEditor/GenerateSqlBar/GenerateSqlBar.tsx
```diff
@@ -0,0 +1,128 @@
...
+              </button>
+            </div>
+          )}
+        </form>
+      )}
+    </div>
+  );
+};
+
+export default GenerateSqlBar;
```
bramj, line 128: This looks really nice 👍

backend/hexa/databases/utils.py
```diff
@@ -275,6 +275,30 @@ def run_and_log_database_query(
...
+    statement is submitted and ``psycopg2.Error`` when the query is invalid.
+    """
+    ensure_single_statement(query)
+    conn = None
+    try:
+        conn = get_workspace_database_ro_connection(workspace)
+        with conn.cursor() as cursor:
+            cursor.execute(
+                sql.SQL("SET LOCAL statement_timeout = {timeout};").format(
+                    timeout=sql.Literal(timeout_ms)
```
bramj: The timeout here can actually just be hardcoded, there's no reason we'll ever want to override it.

backend/hexa/assistant/agents/generate_sql_agent.py
```diff
@@ -0,0 +1,119 @@
...
+    """Turns a natural-language request into a validated PostgreSQL query.
+
+    The final answer is enforced to be a single read-only SQL statement: an
+    output validator EXPLAINs it against the workspace database (read-only
+    role) and raises ModelRetry with the database error so the model can
+    correct itself.
+    """
+
+    instruction_set = InstructionSet.GENERATE_SQL
+    tools = [execute_sql]
```
bramj: I know it's more limiting, but we actually only want to give the agent access to the _schema_, not to the data itself. So we should use existing tools `get_db_schema` and `get_db_table_schema` here.

bramj (changes requested):
This is so cool @DimitriKwihangana ! I think **this** is a very valuable feature for our data manager users: it allows them to ask questions like "what are the trends on my maternal deaths over the last 3 years per province?".

We'll definitely want to expand on this and do things like propose examples so users can really understand how useful this is.

Mostly minor comments but one big one: for now we don't want to expose any data from the database to the LLM, so this agent should use only schema and table inspection tools, rather than actually executing queries.

backend/hexa/databases/utils.py
DimitriKwihangana: I think we do need a refactor here. These looked different at first  some needed pagination, others were just listing tables  but they've ended up very similar. I'd  make `get_database_definition_page` the single source of truth and turn the others into thin wrappers over it. **NB**: that is not implemented in this PR.

backend/hexa/databases/utils.py
```diff
@@ -428,6 +457,33 @@ def get_database_definition_page(
             conn.close()
 
 
+def get_full_database_definition(workspace: Workspace) -> List[Dict]:
```
bramj, line 460: Is this similar to what we use for the DB schema in the sidebar + the autocompletion? Out of scope for this PR, but I feel we can take a moment to refactor this `utils.py` file to make sure we maintain a minimal set of methods.

bramj (approved):
Looking good @DimitriKwihangana, nice work ! 👏

PR #1928, opened 2026-07-22
chore: Add Django admin view for saved queries

bramj:
Minor importance, but will come in handy at some point.

image


backend/hexa/data_studio/migrations/0003_alter_savedquery_options.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("data_studio", "0001_initial"),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name="savedquery",
+            options={
+                "ordering": ["-updated_at"],
```
mrivar, line 15: I think ordering is already handled in migration 0001

mrivar (commented):
Nice one! I love that N+1 queries is handled 🤪

One more producty question: do we want to keep some values read-only (such as `created by` or `content`), or is it fine being editable? More food for thought than anything, just worrying about editing a functioning query by mistake and annoying someone while at it 😂

bramj:
> One more producty question: do we want to keep some values read-only (such as created by or content), or is it fine being editable?

@mrivar It's quite okay to not worry about this actually. We're very restrictive with access to the Django admin on prod, and so people that use it are supposed to know what they're doing. And it's actually a feature to be able to "correct" data when the need arises (it rarely does).

backend/hexa/data_studio/migrations/0003_alter_savedquery_options.py
bramj: Late reply: it is indeed! But Django adds the full thing because I added the `verbose_name_plural`, this migration is just the result of `./manage.py makemigrations`. I've always found that a bit weird as well, but 🤷‍♂️

PR #1929, opened 2026-07-22
chore: Improve assistant conversation Django admin

bramj:
To be able to better view and debug expensive conversations, this commit adds:

- Add a collapsible section on the conversation page with an estimated context-size token count and a copy-JSON button to copy the whole payload. This will allow us to figure out why context and cost exploded.
- Add cost/token warning highlighting: orange above $2 / 200k tokens, red above $10 / 500k tokens. So we can easily spot potentially problematic conversations.
- Show per-message input tokens, output tokens, and cost in the `MessageInline` on the conversation page.
- Add date_hierarchy, to easily search in time.

### Conversations list
image

### Conversation details
image

### DIFF
--- a/backend/hexa/assistant/admin.py
+++ b/backend/hexa/assistant/admin.py
@@ -1,3 +1,5 @@
+import json
+
 from django.contrib import admin
 from django.templatetags.static import static
 from django.urls import reverse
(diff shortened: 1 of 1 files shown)


backend/hexa/assistant/admin.py
```diff
@@ -6,6 +8,47 @@
...
+def _format_with_warning(value, formatter, tiers):
+    if value is None:
+        return "—"
+    text = formatter(value)
+    for threshold, style in tiers:
+        if value >= threshold:
+            return format_html("<span style='{}'>{}</span>", style, text)
+    return text
+
+
```
DimitriKwihangana, line 44: I like the warning mechanisms! Otherwise we would find it hard to monitor!

DimitriKwihangana (approved):
Veery Nice! I believe this will help us to manage the costs.

PR #1930, opened 2026-07-22
chore: Bring back BYOK option on organization when the instance is managed (HEXA-1753)

bramj:
On managed instances (`ASSISTANT_MANAGED=true`), the organization AI Assistant settings previously only exposed the **Enabled** toggle, forcing the provider to "Managed". 

This adds the ability for admins to opt out of the managed provider and bring their own Anthropic API key.

Jira: https://bluesquare.atlassian.net/browse/HEXA-175

## How/what to test

Navigate to the organization settings, confirm that the Anthropic options is now available.

## Screenshots

image
image

### DIFF
--- a/frontend/src/organizations/features/OrganizationAiSettings/OrganizationAiSettings.tsx
+++ b/frontend/src/organizations/features/OrganizationAiSettings/OrganizationAiSettings.tsx
@@ -17,6 +17,11 @@ const MODELS: Record<string, string[]> = {
   [AiProvider.Anthropic]: ["", AiModel.Opus, AiModel.Sonnet, AiModel.Haiku],
 };
 
+const usesManagedProvider = (value: unknown) => value === AiProvider.Managed;
+
+const showByok = (values: { enableAI?: unknown; provider?: unknown }) =>
+  Boolean(values.enableAI) && !usesManagedProvider(values.provider);
+
 type OrganizationAiSettingsProps = {
   organization: Organization_OrganizationFragment;
 };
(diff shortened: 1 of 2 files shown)


frontend/src/organizations/features/OrganizationAiSettings/OrganizationAiSettings.tsx
```diff
@@ -105,14 +106,10 @@ const OrganizationAiSettings = ({
           options={modelOptions}
           getOptionLabel={getModelLabel}
           visible={(_, __, values) =>
-            Boolean(values.enableAI) &&
-            !isManagedInstance &&
-            !usesManagedProvider(values.provider)
+            Boolean(values.enableAI) && !usesManagedProvider(values.provider)
```
mrivar: Minor: we could have a constant `const showByok = (v) => Boolean(v.enableAI) && !usesManagedProvider(v.provider);` to avoid repetition of this value

frontend/src/organizations/features/OrganizationAiSettings/OrganizationAiSettings.tsx
mrivar, line 117: Test gap: Somehow related, it would be nice to have a test that ensures that when not modifying the api key, it doesn't send "••••••" (which it doesn't, but it would be nice to close the test gap and avoid future problems)

mrivar: Minor: Was already here before, so not urgent, but it would be nice to have the provider have only one source of truth (atm it has this `useState` and the `form` instance, they keep synced due to `onChange` in the provider form). So exploring how to get` rid of `useState` in favor of the form value would be nice for future maintainability.

mrivar (commented):
Nice having this back!
Nothing really blocking, just some code improvements suggestions, you can take it or leave it :)
The only thing I'd push more for is filling the test gap for coverage.

frontend/src/organizations/features/OrganizationAiSettings/OrganizationAiSettings.tsx
bramj: Tried to improve on this, but it required some added complexity that I didn't like, going to keep it as is for now.
I added a commit for your other comments.

PR #1931, opened 2026-07-22
feat: add mixpanel autocapture and session replay (HEXA-1679)

mcrimi:
## Summary

Adds browser-side Mixpanel instrumentation to the frontend: pageview autocapture and session replay. Session replay is scoped to the onboarding experience, and both features are gated by the current user’s `analyticsEnabled` consent.

### Wiring

* Adds a new `core/helpers/mixpanel.ts` helper for:
    * SDK initialization
    * user identification
    * session replay gating
* Calls the helper from `_app.tsx`:
    * initializes and identifies on mount
    * re-evaluates the replay gate on every route change
* Exposes `MIXPANEL_TOKEN` and `MIXPANEL_API_HOST` to the browser via `window.__ENV__` runtime config.
* Keeps `DISABLE_ANALYTICS=true` as a frontend kill switch.
* Disables the SDK entirely when `MIXPANEL_TOKEN` is empty, so self-hosted deployments must explicitly configure one.
* Normalizes the API host to a full URL. Unlike the Python SDK, the browser SDK treats a bare hostname as relative and posts to the app’s own origin.

### Consent

* Exposes `analyticsEnabled` on the GraphQL `User` type using an existing per-user setting.
* Adds a snake_case fallback resolver with no backend logic changes.
* Fetches `analyticsEnabled` through `GetUser`.
* Syncs browser SDK state on load:
    * opted-in users are identified
    * opted-out users are opted out
* Sends the `$opt_in` event only when the user changes consent, not on every page load.

### Autocapture — Pageviews Only

* Enables `pageview: "url-with-path"`.
* Re-fires pageviews on route changes.
* Ignores query strings and hashes.

### Onboarding Session Replay

* Starts recording only when all of the following are true:
    * the user has `analyticsEnabled`
    * the account is less than 7 days old
    * the browser session is one of the user’s first 3 sessions
    * the current route is not sensitive - excludes auth, pipeline code editor, notebooks, connections.
* Masks all text with `record_mask_text: true`.
* Relies on the SDK default behavior for always-masked inputs.
* Limits recordings to a maximum length of 10 minutes.

## Follow up work

- Custom product events (onboarding milestones, etc.) - follow-up PR.
- Self-service consent: `analyticsEnabled` is currently only editable via the Django admin. I think it's important to enable users to set their preference via Account Management - I'll be creating a follow-up PR for that.

## Testing

Tested locally using the Mixpanel `OpenHEXA Dev` Project. Checked that autocapture events are reaching the project and that session replays were captured for a recently onboarded user:

https://github.com/user-attachments/assets/f9464160-9b9f-40c7-8fd7-25afd089cb49

https://github.com/user-attachments/assets/2e62ea9b-7ffc-4ffb-856c-c7b2d6e5572b


frontend/src/core/helpers/mixpanel.ts
```diff
@@ -0,0 +1,159 @@
...
+const SESSION_MARKER_KEY_PREFIX = "hexa_replay_session_counted";
+
+// Routes where we never want to capture a session replay. The pipeline code
+// editor and notebooks can render user-supplied credentials or PHI; auth flows
+// contain passwords and MFA codes; account settings handle 2FA secrets.
+const SENSITIVE_ROUTE_PATTERNS: RegExp[] = [
+  /^\/login/,
+  /^\/register/,
+  /^\/forgot-password/,
+  /^\/account/,
```
bramj: These paths seem incorrect:
```suggestion
  /^\/signup/,
  // Covers /auth/password_reset and /auth/reset/<uid>/<token>.
  /^\/auth\//,
  /^\/user\/account/,
```

frontend/src/core/helpers/mixpanel.ts
```diff
@@ -0,0 +1,159 @@
...
+      capture_text_content: false,
+    },
+    // Session replay disabled by default; opt-in per session via applyReplayGate.
+    record_sessions_percent: 0,
+    record_idle_timeout_ms: 600_000,
+    record_min_ms: 2_000,
+    // Mask all text and inputs by default; defense in depth on top of the
+    // sensitive-route exclusions below.
+    record_mask_text_selector: "*",
+    record_block_selector: "[data-mp-no-record]",
```
bramj: Should we also add the default `img, video, audio` back here?

https://docs.mixpanel.com/docs/session-replay/session-replay-privacy-controls#common-gtm-privacy-configurations

bramj (changes requested):
Just a small doubt on some of the sensitive paths.

frontend/src/core/helpers/mixpanel.ts
mcrimi: I had wrongly assumed that those were conserved by default - adding them explicitly. Thanks!

mcrimi:
Hi @bramj - thanks for the review (and sorry for not checking the paths, my bad). I've addressed your comments.

bramj (approved):
Looking good to me! 👍

PR #1932, opened 2026-07-22
fix: open hexa banner should not overlap webapp page(HEXA-1751)

DimitriKwihangana:
This adds a spacer <div> matching the banner's height, injected right before the banner, and gives the banner an explicit height (2.25rem, replacing the old padding-based sizing) so the two stay in sync. The webapp's content now gets pushed up instead of being covered.

## How/what to test

Create a web app that doesn't have padding and margin bottom. the banner will be beside the content


bramj (approved):
Great!

mrivar (approved):
nice!

mrivar (approved):

PR #1933, opened 2026-07-22
refactor: Generalise DatabaseQueryLog  and move it in datastudio django app(HEXA-1752)

DimitriKwihangana:
Moves the SQL query audit log out of `databases` and into `data_studio`, renaming it `QueryLog`, so it's no longer tied to Postgres-specific naming as Data Studio prepares to support other source types (Parquet, CSV, SQLite, ...).

## Changes

- Moved `DatabaseQueryLog` → `data_studio.QueryLog` (renamed model + queryset), alongside the existing `SavedQuery` model.
- Added a `target` field plus indexes on `status`, `origin`, and `target`, so future source types can be logged and filtered without a new migration.
- Folded the model creation directly into `data_studio`'s initial migration and drop `databases`' initial migration nothing has been deployed yet, so no data migration is needed.
- Updated `databases/utils.py` and `databases/schema.py` to log against `QueryLog`; 
- Add a read-only `QueryLogAdmin` in `data_studio/admin.py`.
- Moved the corresponding tests.

Note:`executeSQL` resolver and its permissions stay in `databases` its refactoring is not implemented in this PR


bramj:
Note for other people testing on localhost:

```bash
# undo edited migrations on main branch
./manage.py migrate databases zero
./manage.py migrate data_studio zero

# then switch to this branch and run migrations
./manage.py migrate
```

bramj (approved):
Looking great @DimitriKwihangana ! 👍

PR #1935, opened 2026-07-23
feat: Support the copying of template pipelines to a remote server (HEXA-1757)

bramj:
This add a script that copies all template pipelines from an OH instance (by default our prod SaaS instance) to an OH instance of your choice.

Since template pipelines are linked to workspaces, it creates a new workspace "Template pipelines" on the target instance, and then continues to add all template pipelines and their versions to this workspace.

Re-running the script will check all pipelines, create new ones where they're missing, and add new versions if there are missing versions on the target.

Two entrypoints again: a CLI command and a Django admin command.

Example usage for CLI (`--source-url` defaults to production):

```bash
# copy from prod to demo
./manage.py copy_templates \
	--source-token 'my-prod-token' \
	--target-url https://api.demo.openhexa.org/graphql/ \
	--target-token 'my-demo-token' \
	--target-organization 002f2f74-7cdb-452c-8ef5-28cc27c04fbe

# copy from prod to your localhost
./manage.py copy_templates \
        --source-token 'my-prod-token' \
        --target-url http://app:8000/graphql/ \
	--target-token 'my-localhost-token' \
        --target-organization a6dfd796-7d2a-4857-9638-e35d88243132
```

### How to test

**Context:** The context for this is to be able to easily set up template pipelines on localhosting instances. We do not want to invest a lot of time into this. So testing should only be functional, no need to read every line of code.

#### Admin interface
image

#### Example CLI output
image

### DIFF
--- a/backend/hexa/workspace_copier/admin.py
+++ b/backend/hexa/workspace_copier/admin.py
@@ -11,10 +11,14 @@ from django.contrib import admin, messages
 from django.core.exceptions import PermissionDenied
 from django.template.response import TemplateResponse
 
-from hexa.workspace_copier.forms import CopyWorkspaceForm
+from hexa.workspace_copier.forms import CopyTemplatesForm, CopyWorkspaceForm
 from hexa.workspace_copier.progress import BufferReporter
-from hexa.workspace_copier.results import format_summary
-from hexa.workspace_copier.service import CredentialError, run_copy
+from hexa.workspace_copier.results import format_summary, format_templates_summary
+from hexa.workspace_copier.service import (
+    CredentialError,
+    run_copy,
+    run_template_copy,
+)
 from hexa.workspace_copier.transport import GraphQLError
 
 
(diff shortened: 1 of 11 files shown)


backend/hexa/workspace_copier/tests/test_service.py
DimitriKwihangana, line 1: I think the tests here mostly cover run_template_copy credential checks mostly.Do we have tests that covers the logic that is in `templates.py`?

backend/hexa/workspace_copier/admin.py
```diff
@@ -62,3 +66,46 @@ def copy_workspace_view(request):
...
+
+    summary = None
+    log = None
+    if request.method == "POST":
+        form = CopyTemplatesForm(request.POST)
+        if form.is_valid():
+            data = form.cleaned_data
+            reporter = BufferReporter()
+            try:
+                result = run_template_copy(
```
DimitriKwihangana, line 83: I don't have an alternative but I think with the delay. large number of templates will take a lot of time or get a timeout error even if the copy is still running.

DimitriKwihangana (approved):
LGTM! only minor comments

backend/hexa/workspace_copier/admin.py
bramj: Indeed, it takes a little while, but for now we're talking ~2 minutes. Since this is a one-shot during setup of a locally hosted instance, for now this is more than acceptable. Note that on a second run it will only copy over new versions, which is a lot faster.

backend/hexa/workspace_copier/tests/test_service.py
bramj: I've been deliberately light on tests here, because I think the current direction is not great on that front. There's a ticket in the backlog to improve on this: https://bluesquare.atlassian.net/browse/HEXA-1767

PR #1937, opened 2026-07-23
feat: Make workspace copy script idempotent + file copy improvements (HEXA-1755)

bramj:
Several improvements to the "copy file" part of the workspace copier, as well a general improvement to be able to "resume" a copy.

## Changes

- Make copy_workspace command idempotent: Useful if a copy breaks somewhere in the middle. Add a flag `--target-workspace-slug`, if set:
  - Copy into an existing workspace, early return if the slug doesn't exist.
  - Copy files (they'll be overwritten), skip pipelines and connections that already exist.
- Improved file copying:
  - Added a counter
  - Skip existing files when copying into an existing workspace
  - Skip several directories such as `cache` to save time

## How/what to test

You can run the script via the CLI or Django admin, instructions are in the `backend/hexa/workspace_copier/README.md`.

## Screenshots / screencast

Red box marks the new part:
image

### DIFF
--- a/backend/hexa/workspace_copier/orchestrator.py
+++ b/backend/hexa/workspace_copier/orchestrator.py
@@ -52,7 +52,6 @@ def copy_workspace(
     *,
     resources: set[str] | None = None,
     options: CopyOptions = CopyOptions(),
-    skip_existing: bool = False,
 ) -> CopyResult:
     """Copy a workspace from ``source`` to ``target``.
 
@@ -62,17 +61,8 @@ def copy_workspace(
     to discard it. ``options`` carries the run-wide switches (see
     :class:`~hexa.workspace_copier.options.CopyOptions`); every copier receives
     them and reads only what concerns it.
-    ``skip_existing`` skips files already on the target with
-    the same key and size (the idempotent re-run flow).
     """
     selected = _resolve_selection(WORKSPACE_COPIERS, resources)
-    if skip_existing:
-        # Only the files copier has per-run configuration; swap in a configured
-        # instance rather than mutating the shared registry singleton.
-        selected = [
-            FilesCopier(skip_existing=True) if isinstance(c, FilesCopier) else c
-            for c in selected
-        ]
     selected_names = {c.name for c in selected}
     result = CopyResult()
     for copier in selected:
--- a/backend/hexa/workspace_copier/orchestrator.py
+++ b/backend/hexa/workspace_copier/orchestrator.py
@@ -52,6 +52,7 @@ def copy_workspace(
     *,
     resources: set[str] | None = None,
     options: CopyOptions = CopyOptions(),
+    skip_existing: bool = False,
 ) -> CopyResult:
     """Copy a workspace from ``source`` to ``target``.
 
@@ -61,8 +62,17 @@ def copy_workspace(
     to discard it. ``options`` carries the run-wide switches (see
     :class:`~hexa.workspace_copier.options.CopyOptions`); every copier receives
     them and reads only what concerns it.
+    ``skip_existing`` skips files already on the target with
+    the same key and size (the idempotent re-run flow).
     """
     selected = _resolve_selection(WORKSPACE_COPIERS, resources)
+    if skip_existing:
+        # Only the files copier has per-run configuration; swap in a configured
+        # instance rather than mutating the shared registry singleton.
+        selected = [
+            FilesCopier(skip_existing=True) if isinstance(c, FilesCopier) else c
+            for c in selected
+        ]
     selected_names = {c.name for c in selected}
     result = CopyResult()
     for copier in selected:
--- a/backend/hexa/workspace_copier/resources/files.py
+++ b/backend/hexa/workspace_copier/resources/files.py
@@ -231,10 +231,11 @@ class FilesCopier(ResourceCopier):
                 f"   could not list target files ({exc}) — copying everything"
             )
             return {}
-        reporter.info(
-            f"   target already has {len(existing)} file(s); "
-            "files matching by key and size will be skipped"
-        )
+        if existing:
+            reporter.info(
+                f"   target already has {len(existing)} file(s); "
+                "files matching by key and size will be skipped"
+            )
         return existing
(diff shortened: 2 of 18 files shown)


backend/hexa/workspace_copier/service.py
```diff
@@ -148,8 +185,15 @@ def run_copy(
...
         target_organization_id=target_organization_id,
         target_workspace_name=target_workspace_name,
+        target_workspace_slug=target_workspace_slug,
+    )
+    return copy_workspace(
+        source,
+        target,
+        reporter,
+        resources=resources,
+        skip_existing=bool(target_workspace_slug),
```
mrivar: We have a double source of truth for when to skip existing files, here and in 
```python
# workspace.py:39
  if target.slug:
      self._use_existing_target(target, result, reporter)
```

I'd vote to instead have a property in the Endpoint class where we get that info.

backend/hexa/workspace_copier/resources/files.py
```diff
@@ -20,6 +20,16 @@
 
 OBJECTS_PAGE_SIZE = 100
 
+# Directory names whose contents are never worth copying (editor/runtime
+# scratch dirs). Matched against any segment of an object key, so a nested
+# ``notebooks/.ipynb_checkpoints/foo.ipynb`` is skipped too.
+SKIPPED_DIRECTORIES = frozenset({".ipynb_checkpoints", "cache", ".cache"})
```
mrivar: Plain `cache/` might be risky. A file called `data/cache/reference-tables.csv` would disappear without any notice. WDYT?
It would be handy as well to have these skipped directories documented somewhere in the admin page

backend/hexa/workspace_copier/orchestrator.py
```diff
@@ -49,15 +49,24 @@ def copy_workspace(
...
+    the same key and size (the idempotent re-run flow).
     """
     selected = _resolve_selection(WORKSPACE_COPIERS, resources)
+    if skip_existing:
+        # Only the files copier has per-run configuration; swap in a configured
+        # instance rather than mutating the shared registry singleton.
+        selected = [
+            FilesCopier(skip_existing=True) if isinstance(c, FilesCopier) else c
+            for c in selected
+        ]
```
mrivar: This logic feels a bit awkward. If we already do the endpoint property to avoid the double source of truth, maybe it's okay to just check it in the `copy` function, wdyt?

A more "clean" way to do this is to have all ResourceCopiers have a constructor with a class input CopyOptions, but I'd say that would be overcomplicated for this case; maybe useful if we add more "options" in the future

backend/hexa/workspace_copier/service.py
```diff
@@ -66,6 +69,32 @@ def _build_target(
...
+    copier skips creation and the downstream copiers make the rest of the run
+    idempotent by skipping resources that already exist.
+    """
+    if url:
+        client = build_client(url, token, label="target")
+        if client.workspace(slug=workspace_slug) is None:
+            raise GraphQLError(
+                f"target workspace '{workspace_slug}' not found — create it first "
+                "or omit --target-workspace-slug to create a new workspace."
+            )
```
mrivar: would be nice to add a test to check this logic

backend/hexa/workspace_copier/resources/workspace.py
```diff
@@ -48,6 +53,29 @@ def copy(
...
+        resources that already exist on the target.
+        """
+        ws = self._read_target(target)
+        result.workspace_slug = target.slug
+        result.workspace_name = ws.name
+        reporter.info(f"   using existing workspace {ws.name!r} (slug '{target.slug}')")
+
+    def _read_target(self, target: Endpoint) -> Any:
+        if target.is_remote:
+            tgt_ws = target.client.workspace(slug=target.slug)
```
mrivar: We fetch the workspace twice (here and in service.py:86). Might be worth it to stash it in Endpoint.workspace, given we already have that param

backend/hexa/workspace_copier/resources/workspace.py
```diff
@@ -35,6 +35,11 @@ def copy(
         reporter: ProgressReporter,
     ) -> None:
         src_ws = self._read_source(source)
```
mrivar: if we move this after the `if target.slug` we avoid fetching the source for nothing

mrivar (changes requested):
Very nice improvement!! I hadn't check this feature yet so it's cool to try it and dive into the code. Mainly some correctness and maintainance comments, as well as the test gap which seems quite relevant.

Apart from it, Claude suggests that there might be a timeout risk due to the growing amount of logs. Haven't tested but I think it's worth mentioning or creating a task to investigate:

> The admin path is the real ops risk. admin.py runs the whole copy synchronously inside the HTTP request, with BufferReporter holding every line in memory and dumping it into one `<pre>`. This PR adds a full extra target-bucket listing before any copying (pushing time-to-first-byte up) plus ~1 line per file. A 20k-file re-run means a 20k-line response from a request that was already at gateway-timeout risk. Pre-existing, but amplified — and progress.py's docstring already anticipates the fix ("when we run the copier in an async job"). Worth stating in the PR that the admin path is CLI-scale-only until then.

backend/hexa/workspace_copier/orchestrator.py
bramj: You're very right, and funnily enough, `CopyOptions` is exactly what @yolanfery added :)
I've refactored mine to add it to this new dataclass, the result is way better 👍

backend/hexa/workspace_copier/resources/workspace.py
bramj: Great point :+1: fixing

backend/hexa/workspace_copier/orchestrator.py
bramj: @mrivar on second thought, I got rid of the `skip_existing` entirely, it is not standard behavior. That makes sense and is way simpler.

mrivar (approved):
Looking much better!!

PR #1938, opened 2026-07-24
feat: add schema-aware sql autocomplete to Data Studio editor (HEXA-1714)

DimitriKwihangana:
The Data Studio SQL editor now autocompletes the workspace's tables and columns. Schema introspection is cached server-side to keep it cheap.

## Changes

- CodeEditor: new `sqlSchema`/`sqlDefaultTable` props feeding CodeMirror's sql() schema option.
- DataStudioEditor: loads the schema (via Apollo, deduped with the schema browser) and passes it to the editor.
- Backend: cache schema pages with a self-invalidating TTL (new WORKSPACE_DATABASE_SCHEMA_CACHE_TTL, default 60s); bump on delete_table.

## How/what to test
Write a query you'll be suggested tables as shown in screenshot

## Screenshots / screencast

Screenshot 2026-07-27 at 5 46 47 PM


backend/hexa/databases/utils.py
```diff
@@ -307,6 +308,38 @@ def get_database_definition(workspace: Workspace):
...
+
+def _bump_schema_cache_version(workspace: Workspace) -> None:
+    """Invalidate every cached schema page for this workspace at once.
+
+    Cheaper than tracking/deleting each (page, per_page, with_columns,
+    with_counts) cache key individually, and works on any cache backend
+    (LocMemCache has no key-pattern deletion).
+    """
+    key = _SCHEMA_CACHE_VERSION_KEY.format(workspace.id)
+    cache.set(key, _get_schema_cache_version(workspace) + 1)
```
mrivar: I really like the implementation of the cache and the cache key and what not, it's very smart. However, because we only have in-memory cache but 3 different workers, when bumping the schema cache it will only be updated in one of the workers, leaving the other 2 outdated.

It's not a big deal, but I think it's worth a mention/comment.
I'm not sure if there's a way to mitigate it, using some sharing memory cache in-between processes, but it might be worth some investigation if we want to go down that road. We could use a postgre cache but I honestly don't think is worth it for this. And a more complicated solution such as Redis is too much for this too and it would include infra work. 

I leave it to you if you want to look into it but I wouldn't waste much time in this.

mrivar: This cache.set should add the timeout=None in order not to have the Django default one, which would clear the cache after 600s

backend/hexa/databases/tests/test_utils.py
```diff
@@ -164,6 +166,13 @@ def setUpTestData(cls):
             countries=[],
         )
 
+    def setUp(self):
+        # The schema cache (LocMemCache) is a process-global store that isn't
+        # rolled back with the per-test DB transaction, so tests that hit the
+        # same (workspace, page, per_page, with_columns, with_counts) key
+        # would otherwise silently read another test's cached result.
+        cache.clear()
```
mrivar: We can also turn off caching in `config/settings/test.py:`
```python
CACHES = {"default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}}
```
caching off by default in tests (what you want 99% of the time) — then @override_settings a LocMemCache on GetDatabaseDefinitionPageCacheTest. Would avoid remembering having to add these cache.clear() setups into other tests in the future

frontend/src/core/components/CodeEditor/CodeEditor.tsx
```diff
@@ -44,6 +44,10 @@ type CodeEditorProps = {
    */
   shortcuts?: CodeEditorShortcut[];
   className?: string;
+  /** Table/column metadata for `lang="sql"` autocomplete (`sql()`'s `schema` option). */
+  sqlSchema?: SQLNamespace;
+  /** Table assumed for bare column references when `lang="sql"`, e.g. in a single-table editor. */
+  sqlDefaultTable?: string;
```
mrivar: Can we use this somehow? Otherwise, it's dead code, not sure if it's worth it to keep it

backend/hexa/databases/utils.py
```diff
@@ -427,6 +466,9 @@ def get_database_definition_page(
         if conn:
             conn.close()
 
+    cache.set(cache_key, result, settings.WORKSPACE_DATABASE_SCHEMA_CACHE_TTL)
```
mrivar: In the result we're caching, under items (via _fetch_tables_with_columns or _fetch_tables), we store the workspace object which contains the db passwords. If they are changed while cached and the user goes to the page, they would get a 500 error.

To avoid this we coudl try not to cache the workspace object and rehydrate with the latest workspace data. Worth investigating it.

mrivar (changes requested):
Great work 😁! I really like the cache implementation and how it works, I just think it needs a bit more analysis on how to handle cache mismatches to avoid having random erros on the user side. As the famous quote says:
> There are only two hard things in Computer Science: cache invalidation and naming things

frontend/src/core/components/CodeEditor/CodeEditor.tsx
bramj: How does this work? It doesn't feel like something we can use for our use case: https://github.com/codemirror/lang-sql#user-content-sqlconfig.defaulttable

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
```diff
@@ -53,6 +55,28 @@ const DataStudioEditor = ({
   const { run, retry, result, loading, error, canExport } =
     useDataStudioQuery(workspaceSlug);
 
+  // Same query DataStudioSchemaBrowser runs to populate its table tree.
+  // Apollo dedupes identical in-flight queries and serves matching variables
+  // from its normalized cache afterwards, so this doesn't add a network
+  // request — it's how the editor gets at the schema it needs for autocomplete.
+  const { data: schemaData } = useWorkspaceDataStudioSchemaQuery({
+    variables: { workspaceSlug },
+  });
```
bramj, line 81: Nice how we can reuse the query from the table tree 👍

bramj (changes requested):
Really great that you could reuse the payload that serves the Data Studio sidebar with the DB schema! 👍 

With respect to the caching: is won't really work well and even if it did, I would still argue against it.

The reason we don't use the Django cache (`django.core.cache`) is because in it's current implementation it is a `LocMemCache` (https://docs.djangoproject.com/en/6.0/topics/cache/#local-memory-caching) which, as @mrivar pointed out, will live only inside the current process. On prod we serve the backend on at least 3 kubernetes pods with each running several processes. The only way to do this correctly would be to add a tool like Redis to our infrastructure.

But even if it worked fine, this would be a good example where the functionality added does not justify the amount of complexity. This will make our codebase grow and become harder to reason about, for limited gain.
The first question is: do we even need caching at all? I'd say we probably don't since a schema inspection is not expensive as long as we don't add table counts to the results (which we don't).
Caching and cache key invalidation is notoriously difficult and to be avoided until absolutely necessary. If we see that in the future this becomes a problem, we'll first try to simplify the underlying query before even thinking about caching.

DimitriKwihangana:
> Really great that you could reuse the payload that serves the Data Studio sidebar with the DB schema! 👍
> 
> With respect to the caching: is won't really work well and even if it did, I would still argue against it.
> 
> The reason we don't use the Django cache (`django.core.cache`) is because in it's current implementation it is a `LocMemCache` (https://docs.djangoproject.com/en/6.0/topics/cache/#local-memory-caching) which, as @mrivar pointed out, will live only inside the current process. On prod we serve the backend on at least 3 kubernetes pods with each running several processes. The only way to do this correctly would be to add a tool like Redis to our infrastructure.
> 
> But even if it worked fine, this would be a good example where the functionality added does not justify the amount of complexity. This will make our codebase grow and become harder to reason about, for limited gain. The first question is: do we even need caching at all? I'd say we probably don't since a schema inspection is not expensive as long as we don't add table counts to the results (which we don't). Caching and cache key invalidation is notoriously difficult and to be avoided until absolutely necessary. If we see that in the future this becomes a problem, we'll first try to simplify the underlying query before even thinking about caching.

Thanks Bram! I agree. going to remove whole chaching layer

DimitriKwihangana:
> Great work 😁! I really like the cache implementation and how it works, I just think it needs a bit more analysis on how to handle cache mismatches to avoid having random erros on the user side. As the famous quote says:
> 
> > There are only two hard things in Computer Science: cache invalidation and naming things

Good points @mrivar , as discussed in the daily, we are removing the caching layer to remove unneeded complexity

bramj (approved):
Nice and simple, love it! ❤️

mrivar (approved):
Sorry for the late review! I love the solution, super simple and easy 😁

PR #1963, opened 2026-07-28
fix: Split QueryLog migration into separate file

bramj:
The initial migration was already deployed on demo, prod and potential local instances. This means that the creation of the `data_studio_querylog` table is skipped on these instances. This fixes that.

Note: You might need to redo migrations on your localhost.


DimitriKwihangana (approved):
I had to do this locally too.

PR #1968, opened 2026-07-29
fix: saved query dialog flickering (HEXA-1762)

mrivar:
Fix saved query modal to flicker. The reason it happened is because the modal would mount just as it was opened, resulting in flickering the first time as it is still "creating itself" while opening

## Changes

- Mount modal on page load, keep hidden until opened

## How/what to test

Open the saved query, ensure it's not flickering as in the video in [HEXA-1762](https://bluesquare.atlassian.net/browse/HEXA-1762)

[HEXA-1762]: https://bluesquare.atlassian.net/browse/HEXA-1762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/core/components/Dialog/Dialog.tsx
+++ b/frontend/src/core/components/Dialog/Dialog.tsx
@@ -16,6 +16,13 @@ import {
 } from "react";
 
 type DialogProps = {
+  // Always render this component and drive visibility from here — never guard it
+  // with `{condition && <Dialog …>}`. Headless UI wraps the dialog in a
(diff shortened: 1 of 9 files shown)


bramj (approved):
Great! 👍 

Always a bit shocking to me how complex this sort of thing is in Next.js 😅

DimitriKwihangana (approved):

PR #1969, opened 2026-07-29
feat:format SQL queries from the data studio editor(HEXA-1713)

DimitriKwihangana:
Adds SQL query formatting to Data Studio : a Format button plus **Cmd/Ctrl+F**, using `sql-formatter` with the PostgreSQL dialect.
Cmd/Ctrl+F is CodeMirror's find binding, so find moves to **Cmd/Ctrl+Shift+F** in this editor only. Shout if you'd rather keep find where it is.

## Changes

- New `formatSql.ts`  wraps `sql-formatter`, leaves unparseable SQL untouched.
- `CodeEditor`: `replaceAll()` on the handle (keeps the rewrite undoable), and relocates the search panel when a caller claims `Mod-f`.
## How/what to test

Write a random sql query:
- Button, Cmd/Ctrl+F and Shift+Alt+F all reformat it; Cmd/Ctrl+Z undoes it.

## Screenshots / screencast
image


frontend/src/workspaces/features/DataStudioEditor/formatSql.test.ts
```diff
@@ -0,0 +1,35 @@
+import { formatSql } from "./formatSql";
+
+describe("formatSql", () => {
```
yolanfery, line 3: I like those small test sanity-checking if the library is (still) working as intended 👍

frontend/src/core/components/CodeEditor/CodeEditor.test.tsx
```diff
@@ -86,4 +87,110 @@ describe("CodeEditor shortcuts", () => {
...
+    );
+    const content = await getContentDOM(container);
+
+    act(() => ref.current?.replaceAll("SELECT\n  1"));
+    // Formatting must not be a one-way door: the pre-format text is still one
+    // undo away.
+    dispatchKeyDown(content, { key: "z", ctrlKey: true });
+
+    expect(content.textContent).toEqual("select 1");
+  });
```
yolanfery, line 194: Super nice.

yolanfery: Can we improve a little bit the test to ensure that after `act(() => ref.current?.replaceAll("SELECT\n  1"));` the text is not "select 1" ? Just a small sanity check before checking the result of undoing

yolanfery (commented):
Well done 👏 I like the code and tests

I am not agreeing with the shortcuts choices however, especially since there is no way for the users to know the new binding of the search shortcut

The search shortcut is a very common one and I assume is used way more than a formatting shortcut so I would keep the search shortcut standard and find a new one (or a variant) for the formatting and show it on the screen

Open to discuss it of course as always

DimitriKwihangana:
> Well done 👏 I like the code and tests
> 
> I am not agreeing with the shortcuts choices however, especially since there is no way for the users to know the new binding of the search shortcut
> 
> The search shortcut is a very common one and I assume is used way more than a formatting shortcut so I would keep the search shortcut standard and find a new one (or a variant) for the formatting and show it on the screen
> 
> Open to discuss it of course as always

I do agree I recently used the search one my self. Let me look around on what we could use for formatting

yolanfery (commented):
Sorry for the misdirection but it has been decided in https://github.com/BLSQ/openhexa-app/pull/1976#discussion_r3726762455 to not include the shortcut visuals on the buttons, for now, so I guess we should do the same for the format button 

Also, a bit out-of-scope of this but I feel like the order of the buttons is... arbitrary ? 

Image

At the very least I would move the Generate with AI button at the far right since it's a bit different than formatting, saving,running queries, maybe you could check with Claude other placements and improve it if possible ?

frontend/package.json
```diff
@@ -39,6 +39,7 @@
     "@codemirror/lang-xml": "^6.1.0",
     "@codemirror/lang-yaml": "^6.1.2",
     "@codemirror/merge": "^6.12.1",
+    "@codemirror/search": "^6.7.1",
```
mrivar: It seems like this dependency is dead code, although I'm not 100% sure. Can we remove it?

mrivar (commented):
Super cool, good job 👏! Happy to see how the Data Studio is improving rapidly!

I agree with @yolanfery comment that the buttons ordering is becoming random, which is probably the effect of growing this rapidly 😅 we can do the improvement Yolan mentioned but reordering all of them is out of the scope of the task. Maybe we can create a separate task to improve on that, or do you have any ideas on it @mcrimi?

frontend/src/workspaces/features/DataStudioEditor/formatSql.ts
```diff
@@ -0,0 +1,22 @@
...
+} as const;
+
+/**
+ * Pretty-print a SQL query. A query that cannot be parsed — which is the normal
+ * state of a half-typed one — is returned untouched rather than mangled, so
+ * formatting is never destructive.
+ */
+export const formatSql = (sql: string): string => {
+  try {
+    return format(sql, OPTIONS);
```
mrivar: 1 little improvement we can have here: `import { format } from "sql-formatter"` imports all ~22 dialects into the bundle. If we change this part of the code for only the postgresql dialect we use, we can avoid importing some unused data from our bundle:

```js
import { formatDialect, postgresql } from "sql-formatter";
formatDialect(sql, { dialect: postgresql, keywordCase: "upper", tabWidth: 2 });
```

yolanfery (approved):

PR #1971, opened 2026-07-29
fix: Timestamp formatting and small Django SQL dashboard improvements

bramj:
- **Django SLQ dashboard**
  - Tweak a few settings: Let's just hardcode, no need to add ENV vars
  - Allow serving of Django dashboard on frontend domain: Right now a dashboard will only work on e.g. https://api.openhexa.org/dashboard/some-metrics/ This change will make it also serve on: https://app.openhexa.org/dashboard/some-metrics/
- **Pipeline run timestamps:** There were some discrepencies between PipelineRun "executed_at" and the
actual messages during the execution. The former was correctly shown in the user's timezone, the latter would display in UTC.

**Before:**
image

**After:**
image

### DIFF
--- a/backend/hexa/pipelines/models.py
+++ b/backend/hexa/pipelines/models.py
@@ -3,7 +3,6 @@ import secrets
 import time
 import typing
 import uuid
-from datetime import datetime
 
 from croniter import croniter
 from django.apps import apps
@@ -932,7 +931,7 @@ class PipelineRun(Base, WithStatus):
             {
                 "priority": priority if priority else "INFO",
                 "message": message,
-                "timestamp": datetime.utcnow().isoformat(),
+                "timestamp": timezone.now().isoformat(),
             }
         )
         self.save()
@@ -946,7 +945,7 @@ class PipelineRun(Base, WithStatus):
                 "uri": uri,
                 "name": name,
                 "type": output_type,
-                "timestamp": datetime.utcnow().isoformat(),
+                "timestamp": timezone.now().isoformat(),
             }
         )
         self.save()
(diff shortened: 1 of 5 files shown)


frontend/server/index.mjs
```diff
@@ -22,6 +22,7 @@ const API_PATHS = [
   "/mcp/",
   "/oauth/",
   "/.well-known/",
+  "/dashboard/", // django-sql-dashboard
```
DimitriKwihangana, line 25: Nice!

backend/hexa/pipelines/models.py
```diff
@@ -932,7 +931,7 @@ def log_message(self, priority: str, message: str):
             {
                 "priority": priority if priority else "INFO",
                 "message": message,
-                "timestamp": datetime.utcnow().isoformat(),
+                "timestamp": timezone.now().isoformat(),
```
DimitriKwihangana, line 934: This is the consistent one!

DimitriKwihangana (approved):
Nice!! The implementation looks good but I was expecting to see the dashboard since I'm an admin on prod 😂. Not sure why but it is irrelvant.

bramj:
> Nice!! The implementation looks good but I was expecting to see the dashboard since I'm an admin on prod 😂. Not sure why but it is irrelvant.

The link is actually not a real link  😅 it was just an example to show how we can use the `app` subdomain instead of `api` (should have been clearer about that, feels safer since we're on a public repo)

bramj:
BTW at some point I would like to change the US formatting (default from `en`) to more granular e.g. `en-BE` or `en-GB` depending on the user's browser settings. But that can be for another day.

PR #1972, opened 2026-07-29
feat: Saved query navigation guard (HEXA-1749)

mrivar:
Added a navigation guard in the query editor, for saved queries that have been changed in the UI without being saved.

## Changes

- Improve the existing navigation guard that was already used in the pipelines code FileEditor `useNavigationWarning`
- `useNavigationWarning` in the saved queries editor


frontend/src/workspaces/features/DataStudioEditor/useSavedQueryEditor.ts
```diff
@@ -35,6 +39,29 @@ export const useSavedQueryEditor = ({
...
+    isSaved: Boolean(savedQuery),
+    hasContent: Boolean(content.trim()),
+    canUpdate,
+    canCreate,
+  });
+  const { navigateWithoutWarning } = useNavigationWarning({
+    enabled: Boolean(savedQuery) && isDirty && canPersist,
+    message: savedQuery
+      ? t(
+          'You have unsaved changes to "{{name}}". If you leave this page, they will be lost.',
```
DimitriKwihangana, line 59: I know it is the same implementation in pipelines and webapp. for having browser popups instead of app's dialog not against it but I think with our dialog it would be more cute :) WYT?

Image

DimitriKwihangana (approved):
LGTM!

frontend/src/workspaces/features/DataStudioEditor/SaveQueryButton.tsx
```diff
@@ -118,12 +124,12 @@ const SaveQueryButton = ({
...
   }
 
-  const hasUnsavedEdits = isDirty && hasContent && !saving;
+  const hasUnsavedEdits = isDirty && canSave;
   return (
     <div className="flex items-center gap-1">
       <button
         onClick={onSave}
-        disabled={!isDirty || !hasContent || saving}
+        disabled={!isDirty || !canSave}
```
yolanfery: [Very minor] `!hasUnsavedEdits` is more idiomatic

yolanfery (approved):
Cool overall improvement

frontend/src/workspaces/features/DataStudioEditor/useSavedQueryEditor.ts
mrivar: Yeah I thought about that too, but we have other guards in other parts of the page and we show the native warning as well, so I decided to keep this one for consistency.
It is a good idea for a future improvement tho! but I would then add it for all the guards 😁

frontend/src/workspaces/features/DataStudioEditor/SaveQueryButton.tsx
mrivar: good catch! added

PR #1976, opened 2026-07-30
feat: save query shortcut ctrl+s (HEXA-1749)

mrivar:
Add the option to save queries using ctrl/cmd+s shortcut in the data studio editor

## Changes

- Added hook `useSaveShortcut.ts` to listen to keydown ctrl/cmd+s
- Added `useSaveShortcut` to Data Studio editor
- Added tooltips in the save query buttons for visibility


frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
```diff
@@ -60,12 +61,16 @@ const DataStudioEditor = ({
...
+    canCreate,
   });
 
+  useSaveShortcut(editor.commit);
+
   const runShortcutLabel = isMac ? "⌘+Enter" : "Ctrl+Enter";
   // Compact form for the in-button pill: the return glyph reads cleanly next to
   // ⌘ on macOS; other platforms keep the spelled-out modifier.
   const runShortcutBadge = isMac ? "⌘↵" : "Ctrl+Enter";
+  const saveShortcutLabel = isMac ? "⌘S" : "Ctrl+S";
```
yolanfery: [Minor] This can be encapsulated in the `SaveQueryButton` component

yolanfery: Image

Also I don't see the shortcut icon like I do for the run button and search bar

yolanfery (changes requested):

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
yolanfery: > Image
> Also I don't see the shortcut icon like I do for the run button and search bar

I realize now that it's a label so it appears on hover

Not sure if we should show it though and I actually have the same question for the formatting button (https://github.com/BLSQ/openhexa-app/pull/1969) 

Not sure why we would show it for the run button and not others. WDYT ? @mrivar @DimitriKwihangana

mrivar: Yeah I don't know either. I just felt it was getting too cramped, but we could absolutely add it. I do feel the run shortcut label is more useful as it would be used more tho

mrivar: Moved the shortcut to the SaveQueryButton component, thanks!
What do you think we should do about showing explicitely the shortcut? Should we have it there as well or keep as it is for now?

yolanfery: Following your call, I didn't try it myself so I don't know how it would look, no worries ☺️

yolanfery (approved):

frontend/src/workspaces/features/DataStudioEditor/DataStudioEditor.tsx
mrivar: Sure, as it's a visual thing it can always be added in a min if we feel the need for it :)

PR #1979, opened 2026-08-04
feat: saved query can only be retrieved by workspace

mrivar:
Saved query retrieved by id + **workspace**, keeping it consistent with other data classes queries of the codebase

## Changes

- Made workspace slug mandatory to retrieve a saved query


backend/hexa/data_studio/graphql/schema.graphql
```diff
@@ -55,8 +55,8 @@ extend type WorkspacePermissions {
 }
 
 extend type Query {
-  "Retrieves a saved query by its id. When a workspace slug is given, the lookup is scoped to that workspace."
-  savedQuery(id: ID!, workspaceSlug: String): SavedQuery
+  "Retrieves a saved query by workspace slug and id."
+  savedQuery(workspaceSlug: String!, id: ID!): SavedQuery
```
yolanfery: This pattern seems odd to me. 

If you know the ID (which is unique), you don't need to provide the slug. And if you don't the the id you want to browse what's available in the workspace so you just have : 

```
savedQuery(id: ID!): SavedQuery
savedQueries(workspaceSlug: String!): SavedQuery
```

Does this make sense ?

yolanfery: A bit like https://github.com/BLSQ/openhexa-app/blob/a64a1ea64d5cc451c9b7e6bd0e116175ccbd99d7/backend/hexa/pipelines/graphql/schema.graphql#L814-L823

yolanfery (changes requested):

backend/hexa/data_studio/graphql/schema.graphql
mrivar: yeah I think you're right, I misinterpreted the task. I'm going to rethink about it and go on a different direction.

yolanfery (approved):

PR #1985, opened 2026-08-04
feat: sql editor render widgets when columns adhere to specific format

DimitriKwihangana:
This PR draws a chart as an output when specified in the query

## Changes

- `chart.ts `de tectChart resolves bar/line/pie from the returned column set
- `ResultsChart.tsx`  bars anchored at zero, SVG line over an ordinal x axis, pie folding past the sixth slice into "Other". Caps at 30 bars / 1000 points.
- `docs/en/sql-widgets.md`, registered in hexa/mcp/docs.py so the assistant can serve it.

## How/what to test
- Run a two-column query using one of the conventions and confirm the Chart tab appears with the Table tab showing the same rows
-  Add a query in the agent specify the format you would like the output in

## Screenshots / screencast
Screenshot 2026-08-04 at 3 50 04 PM
Screenshot 2026-08-04 at 3 51 11 PM
Screenshot 2026-08-04 at 3 50 41 PM


frontend/src/workspaces/features/DataStudioEditor/chart.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        [{ pie_label: "a", pie_quantity: 1 }],
+      ),
+    ).toBe("pie");
+  });
+
+  it("does not depend on the order the columns are returned in", () => {
+    expect(detectChart(["bar_quantity", "bar_label"], barRows)).toBe("bar");
+  });
+
+  it("falls back to the table when an extra column is returned", () => {
```
yolanfery: Do we actually want that ? It might make sense to just take the relevant columns and ignore the rest, this could be useful to extract more data than needed/debug but still show the graph

frontend/src/workspaces/features/DataStudioEditor/ResultsChart.tsx
```diff
@@ -0,0 +1,325 @@
...
+          className="fill-gray-400 text-[11px]"
+        >
+          {points[index].label}
+        </text>
+      ))}
+    </svg>
+  );
+};
+
+const Pie = ({ points }: { points: Point[] }) => {
```
yolanfery, line 364: We can probably improve a bit the Pie chart : there are blank parts between the parts, it's not continuous 

Image

Also, hovering the parts should give the label and numeric value (percentage) and "hihglight" the part, I feel like it's a common pattern

docs/en/sql-widgets.md
```diff
@@ -0,0 +1,65 @@
+<div class="hero-section">
+  <h1><i class="fas fa-chart-simple" style="margin-right: 0.5rem;"></i>SQL Chart Widgets</h1>
+</div>
+
+Data Studio draws a chart instead of a table when a query returns a specific set of column names: aliasing columns to `bar_label` and `bar_quantity` renders a bar chart, turning a saved query into a small dashboard.
```
yolanfery, line 5: We should probably have a link to this documentation from the datastudio somewhere

yolanfery (changes requested):
Crazy feature 🤯

Just a few details to fix/change, well done

frontend/src/workspaces/layouts/DataStudioLayout/DataStudioLayout.tsx
```diff
@@ -41,6 +41,16 @@ const DataStudioLayout = ({
       workspace={workspace}
       withMarginBottom={false}
       headerActions={headerActions}
+      helpLinks={[
+        {
+          label: t("SQL chart widgets"),
+          href: "https://docs.openhexa.com/sql-widgets/",
+        },
```
yolanfery, line 51: This is not very visible to users (since they are used to this button and forgot about it), so I am afraid user will not self discover this new sweet feature 

If you find a better way to show it somewhere else (like a small tooltip icon in the query result card), feel free to add it, if not, no worries it's minor

yolanfery (approved):
Looking sharp now, well done !!

frontend/src/workspaces/layouts/DataStudioLayout/DataStudioLayout.tsx
DimitriKwihangana: Good Idea! let me see how it turns out

PR #1986, opened 2026-08-04
feat: saved query visibility (HEXA-1749)

mrivar:
Added visibility for saved queries

## Changes

- Added visibility value and migration (PRIVATE or WORKSPACE, PRIVATE by default. Previously existing queries as WORKSPACE in order not to disrupt previous behaviour)
- New permission to change query visibility (only authors can change it)
- Updated permission to update queries (if a query is private, only authors can update it)
- Added frontend (show visibility in saved queries table, and add the value to be changed in data studio editor and dialog)

## Follow-up
In a new PR, we will add a mechanism on how to handle queries when a user is deleted (nullify on delete has a footgun as private queries of deleted users would not be accesible, and a simple delete-cascade doesn't suffice as we don't want public queries to be deleted when a user is deleted - could affect webapps connected to saved queries etc).


frontend/src/workspaces/features/SavedQueries/savedQueryVisibility.ts
```diff
@@ -0,0 +1,53 @@
...
+    [t],
+  );
+};
+
+export const useSavedQueryVisibilityOption = (
+  visibility: SavedQueryVisibility,
+): VisibilityOption => {
+  const options = useSavedQueryVisibilityOptions();
+  // Falls back to private rather than crashing if the backend ever grows a value
+  // this build does not know about: the safe reading of an unknown visibility.
```
yolanfery: But we get no sentry error then ?

yolanfery (approved):
Tested and reviewed, looks perfect !

Minor : we might not need the optional label there since there is already a preselected option
Image

frontend/src/workspaces/features/SavedQueries/savedQueryVisibility.ts
mrivar: we wouldn't, but we can change that

yolanfery: I would change it yes then

mrivar: Another option is to make it so that a not known value breaks the build at compile-time, thus it would never reach production. I think that is better than firing the sentry error and noticing afterwards, wdyt?

yolanfery: Sure, sounds even better

mrivar: Tested (in github actions, second to last commit), it fails when adding a new visibility. Merging now

mrivar: Also, removed the Optional label as well, thanks!

PR #1987, opened 2026-08-05
fix: archiving workspaces should prevent scheduled pipelines to run (HEXA-1772)

yolanfery:
Ensure that pipelines of archived workspaces are not scheduled for running


DimitriKwihangana (approved):
Nice fix!! @yolanfery  I like these fixes you are doing that were under the hood. :)

PR #1988, opened 2026-08-05
fix: archiving a workspace should rotate db user passwords (HEXA-1773)

yolanfery:
To prevent access to the databases, for now we will just rotate credentials

### DIFF
--- a/backend/hexa/workspaces/models.py
+++ b/backend/hexa/workspaces/models.py
@@ -324,6 +324,8 @@ class Workspace(Base):
             raise PermissionDenied
         if not principal.has_perm("workspaces.archive_workspace", self):
             raise PermissionDenied
+        self.generate_new_database_password(principal=principal)
+        self.generate_new_database_ro_password(principal=principal)
         self.archived = True
         self.archived_at = timezone.now()
         self.save()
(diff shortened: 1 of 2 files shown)


backend/hexa/workspaces/models.py
```diff
@@ -324,6 +324,8 @@ def archive_if_has_perm(self, *, principal: User):
             raise PermissionDenied
         if not principal.has_perm("workspaces.archive_workspace", self):
             raise PermissionDenied
+        self.generate_new_database_password(principal=principal)
+        self.generate_new_database_ro_password(principal=principal)
```
DimitriKwihangana, line 330: Nice!

DimitriKwihangana (dismissed):
It looks very good! But I think we can implement the same thing on deleting a workspace since it is the most uscase that is happening(we are simply archiving the workspace too). And the password are untouched. WDYT?

yolanfery:
> It looks very good! But I think we can implement the same thing on deleting a workspace since it is the most uscase that is happening(we are simply archiving the workspace too). And the password are untouched. WDYT?


Thanks @DimitriKwihangana, great idea, even though we will mostly go for archiving, I think it makes sense to have the same actions there as well, for now

PR #1989, opened 2026-08-06
Add tags to workspace by admins(Hexa-1726)

DimitriKwihangana:
Organization admins and owners can now assign free tags to workspaces and filter theworkspace list by them

## Changes

- `Workspace.tags` M2M onto the existing `tags.Tag` model 

## How/what to test
- Add a tag to the workspace and try to filter


backend/hexa/user_management/schema/types.py
```diff
@@ -313,6 +313,19 @@ def resolve_workspaces(organization: Organization, info, **kwargs):
...
 
+@organization_object.field("workspaceTags")
+def resolve_organization_workspace_tags(organization: Organization, info, **kwargs):
+    request = info.context["request"]
+    return list(
+        Tag.objects.filter(
+            workspaces__in=organization.filter_workspaces_for_user(user=request.user)
+        )
+        .distinct()
+        .order_by("name")
```
yolanfery: Maybe we should have a default model order by name ? Instead of always doing it in the resolver ?

backend/hexa/workspaces/graphql/schema.graphql
```diff
@@ -697,6 +738,7 @@ enum IssueWorkspaceTokenError {
 extend type Mutation {
     createWorkspace(input: CreateWorkspaceInput!): CreateWorkspaceResult! @loginRequired
     updateWorkspace(input: UpdateWorkspaceInput!): UpdateWorkspaceResult! @loginRequired
+    setWorkspaceTags(input: SetWorkspaceTagsInput!): SetWorkspaceTagsResult! @loginRequired
```
yolanfery: Why not having this as part of the update Workspace mutation ?

yolanfery (changes requested):
Code looks good, just a minor comment and a doubt about creating a new mutation for setting tags instead of reusing the update mutation

Also saw some issues in the UI :
- The search does not seem to work with tags, can you verify ? 
- The dropdown in the dialog is causing overflows, it should probably on top of the dialog and not modify the dialog itself 
       - Image
- When there are tags in one of the cards, the buttons of other cards are misplaced
Image

mcrimi:
Hi @DimitriKwihangana - Excited abour this!  I was wondering if the 'Search workspaces' would take into considerations the tags. I think it would be important it does.  cc @bramj @yolanfery

backend/hexa/workspaces/graphql/schema.graphql
DimitriKwihangana: This is a permissions thing. updateWorkspace is open to EDITORs (and workspace ADMINs) too, whereas tagging is deliberately scoped to organization owners and admins only. Let me know what you think if we should allow workspace editors to tag

yolanfery: I think I am fine with limiting to org admins/owners. Still, we can, in the resolver of `updateWorkspace ` set tags if the user has the permission and return a PERMISSIN DENIED error if he is not allowed to. This way we limit widening the number of mutations just for specific attributes like tags and keep a clean and simple CRUD schema.

backend/hexa/workspaces/models.py
```diff
@@ -360,6 +375,29 @@ def update_if_has_perm(self, *, principal: User, **kwargs):
...
+        if not principal.has_perm("workspaces.manage_tags", self):
+            raise PermissionDenied
+
+        slugs = []
+        for name in tag_names:
+            slug = slugify(name)
+            if not slug:
+                raise InvalidTag(f"'{name}' cannot be converted to a valid tag")
+            if slug not in slugs:
+                slugs.append(slug)
```
yolanfery: [Minor] this kind of logic can probably live in the Tag class

We already an abstraction very similar (https://github.com/BLSQ/openhexa-app/blob/150cf641bce01aa1d772f4fdbcb87a76dcf78d35/backend/hexa/tags/models.py#L62-L63) might be worth seeing if we can reuse/extend to reduce logic duplication

backend/hexa/workspaces/schema/mutations.py
```diff
@@ -91,11 +92,21 @@ def resolve_update_workspace(_, info, **kwargs):
...
-        workspace.update_if_has_perm(principal=request.user, **args)
+        has_tags = "tags" in input
+
+        with transaction.atomic():
+            if args or not has_tags:
+                workspace.update_if_has_perm(principal=request.user, **args)
+            if has_tags:
+                workspace.set_tags_if_has_perm(
+                    principal=request.user, tag_names=input["tags"]
+                )
```
yolanfery, line 99: I have a small problem with `set_tags_if_has_perm` and this flow of logic and I think we can restructure it a bit to have the model match the actual intent

What about (still in a transaction) : 
- If there are tags, create them like `Tag.create_if_has_perm(principal=request.user, input["tags"])` and include them in the args. Method and its name to be defined correctly and reuse/adapt existing logic if possible
- In all cases, after that, just do like before `workspace.update_if_has_perm(principal=request.user, **args)`

So the flow is like the mental model : create new tags if you can and assign them to the workspace. And no need for a specific tag setter on the workspace model

yolanfery (dismissed):
UI now looks great

Just proposing a small refactor of the model/abstraction logic

backend/hexa/workspaces/schema/mutations.py
yolanfery: Since Dimitri is OOO, I did the refactoring

PR #1998, opened 2026-08-11
feat: Graphql endpoint that allows execution ofsaved queries(HEXA-1763)

DimitriKwihangana:
Web apps can run a saved query against the workspace database without seeing or editing the SQL. A member saves the query in the Data Studio; the app names it by slug and gets rows back. Private web apps only for now.

NOTE: the proxy was validating top-level fields only, so `USER_READ` reached `workspace { database { executeSQL } }` and let any private web app run arbitrary SQL. 

## Changes

- `slug` on `SavedQuery`  unique per workspace, generated at creation, never changes on rename. Existing rows backfilled.
- `executeSavedQuery` and `savedQueryBySlug`. Reuses `ExecuteSQLResult`;`ExecuteSQLError` gains  `SAVED_QUERY_NOT_FOUND`.
- `DATABASE_READ` web app scope  grants `executeSavedQuery` only, not the lookups, which would expose the SQL body.
- Raw `executeSQL` refused for web app requests, and logged as `DENIED`.
- `QueryLog` records which saved query ran, plus a server-set `WEBAPP` origin.
- Data Studio URLs use slugs; old UUID links redirect.

## How/what to test

`docker compose run app migrate` first.

- Private web app with **Read database** on: `executeSavedQuery` returns rows;
   `executeSQL` is refused even with `USER_READ`.


backend/hexa/data_studio/migrations/0006_savedquery_slug.py
```diff
@@ -0,0 +1,78 @@
...
+    for workspace_id, slug in SavedQuery.objects.exclude(slug="").values_list(
+        "workspace_id", "slug"
+    ):
+        taken[workspace_id].add(slug)
+
+    updated = []
+    for saved_query in SavedQuery.objects.filter(slug=""):
+        used = taken[saved_query.workspace_id]
+        suffix = ""
+        while True:
```
yolanfery: Can we have a max retry attempt here ?

To avoid as much as we can the migration script (which is critical) to be in an infinite loop even if it is unlikely

backend/hexa/data_studio/tests/test_models.py
```diff
@@ -111,6 +111,32 @@ def test_content_is_stored_runnable(self):
...
+        query = self._create(name="My query")
+        query.update_if_has_perm(principal=self.USER_EDITOR, name="Something else")
+        query.refresh_from_db()
+        self.assertEqual("my-query", query.slug)
+
+    def test_slug_falls_back_when_name_has_nothing_to_slugify(self):
+        self.assertEqual("query", self._create(name="!@#$%").slug)
+
+    def test_slug_reusable_across_workspaces(self):
+        # Uniqueness is per workspace, so the same name elsewhere keeps its slug.
```
yolanfery: I know that this is the pattern that we have been using so far but what do you think about having a unique slug across workspaces ? This would simplify querying (no need of the workspace slug anymore) and the mental model (the slug is uniquely identifying a saved query)

Besides historical reason I can't remember the benefit of having it scoped workspace

backend/hexa/data_studio/tests/test_schema.py
```diff
@@ -453,3 +512,145 @@ def test_create_saved_query_unauthenticated(self):
...
+            "workspaceSlug": str((workspace or self.WORKSPACE).slug),
+            "slug": slug,
+        }
+        if max_rows is not None:
+            payload["maxRows"] = max_rows
+        return self.run_query(self.EXECUTE_QUERY, {"input": payload})["data"][
+            "executeSavedQuery"
+        ]
+
+    def test_execute_saved_query(self):
```
yolanfery, line 671: Cool test 👌

frontend/src/pages/workspaces/[workspaceSlug]/data-studio/queries/[querySlug].tsx
```diff
@@ -65,26 +72,50 @@ export const getServerSideProps = createGetServerSideProps({
...
-        notFound: true,
-      };
+    if (!data.workspace) {
+      return { notFound: true };
     }
+
+    if (!data.savedQueryBySlug) {
+      // Saved queries used to be addressed by id. A link made before the move
+      // still resolves, once: redirect it to the slug URL rather than 404.
+      if (UUID_RE.test(querySlug)) {
```
Comment (yolanfery) [inline frontend/src/pages/workspaces/[workspaceSlug]/data-studio/queries/[querySlug].tsx]: We can probably use something like this function https://github.com/BLSQ/openhexa-app/blob/382ba0fdfc899fdb20adfc88416ff7fc3cba9f9a/frontend/src/core/helpers/index.ts#L5-L7 instead of a regex

backend/hexa/webapps/tests/test_graphql_proxy.py
```diff
@@ -363,6 +364,122 @@ def test_disallowed_operation_returns_403(self):
         data = json.loads(response.content)
         self.assertIn("runPipeline", data["errors"][0]["message"])
 
+    def test_user_read_cannot_reach_execute_sql(self):
```
yolanfery, line 367: Good catch and test

Hope we will not find additional holes like this in this part of the system

yolanfery (changes requested):
Great work ! Looking good 

Just 2 minor recommandations and a broader question

backend/hexa/data_studio/migrations/0006_savedquery_slug.py
DimitriKwihangana: Sure thing!

backend/hexa/data_studio/tests/test_models.py
DimitriKwihangana: I agree with this! Do you mean changing it just for saved queries? My instinct is to keep it per-workspace, since going global here means refactoring for consistency everywhere else. Let me know what you think.

yolanfery: Yes, just for saved queries. This brings inconsistency but it simplifies the model/logic and I think it might be worth it. 

But I am not blocking for this

backend/hexa/data_studio/migrations/0006_savedquery_slug.py
```diff
@@ -0,0 +1,82 @@
...
+        suffix = ""
+        for _attempt in range(SLUG_COLLISION_ATTEMPTS):
+            slug = (
+                slugify(saved_query.name[: SLUG_MAX_LENGTH - len(suffix)] + suffix)
+                or "query"
+            )
+            if slug not in taken:
+                break
+            suffix = "-" + secrets.token_hex(3)
+        else:
```
yolanfery, line 43: Didn't know about `for .... else ... ` 🤯

yolanfery (approved):
Great work Dimitri

PR #2000, opened 2026-08-12
feat: account settings - add workspace tokens (HEXA-1754)

mrivar:
Add workspace tokens into the account settings for better visibility for users

## Changes

- New permission for generating workspace tokens
- New graphql endpoint to get the workspace membership of a user (needed to show workspace role) -> this is an extra, we can decide if it's nice for users or if we should drop the role view from the workspace tokens
- Frontend:
  - Workspace token list in account settings
  - Added link in create pipeline from CLI page to account settings, directing users to where they can find the tokens for all workspaces

## Screenshots / screencast

image


backend/hexa/workspaces/schema/queries.py
```diff
@@ -16,6 +18,19 @@
...
+        queryset = queryset.prefetch_related(
+            Prefetch(
+                "workspacemembership_set",
+                # access_token is deferred: it is a secret and is only ever read
+                # through the generateWorkspaceToken mutation, never from here.
+                queryset=WorkspaceMembership.objects.filter(user=request.user).defer(
+                    "access_token"
+                ),
+                to_attr="current_user_memberships",
+            )
```
yolanfery: Not sure this complexity of prefetching (and assuming in the resolver that it can be prefetched) is worth it. Maybe this is a smell that we just need smaller paging when showing workspaces with memberships to reduce the number of queries 

Also it seems, still with this prefetch logic, we have an issue because there is a query for the permission ` permissions { generateToken }` also

frontend/src/workspaces/features/WorkspaceAccessToken/WorkspaceAccessToken.tsx
```diff
@@ -0,0 +1,75 @@
...
+      toast.error(t("Failed to retrieve the access token"));
+      return;
+    }
+    setToken(result.token);
+  };
+
+  if (!canGenerate) {
+    return (
+      <span className="text-sm italic text-gray-500">
+        {t("Not available for viewers")}
```
yolanfery: Maybe we can guide users to their workspace admins to get one ?

frontend/src/workspaces/features/WorkspaceAccessToken/WorkspaceAccessToken.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ) : (
+        <Button variant="secondary" onClick={onShowClick} disabled={loading}>
+          {t("Show")}
+        </Button>
+      )}
+      {temporary && (
+        <span
+          className="shrink-0 text-xs italic text-gray-500"
+          title={t(
+            "You are not a member of this workspace, so its token expires and has to be generated again.",
```
yolanfery: A "direct" member is probably a better wording ? 

And "so its token expires" sounds odds I prefer something like "so this token expires" but pretty sure we can improve this wording overall

backend/hexa/workspaces/graphql/schema.graphql
```diff
@@ -20,6 +20,9 @@ type Workspace {
     # The permissions of the workspace.
```
yolanfery, line 20: For the UI : 
- The temporary flag could look better ? Maybe with a clock icon or a warning sign
- I think we need some naming/type for when the user is a member through non-direct membership to also fill the role column
- After using the show button I feel like we should be able to hide
- The token is not fully visible, this is also the case for the CLI dialog, it was previously and think its better when it's fully visible (to compare with a local token for instance)

Image

yolanfery (changes requested):
Great stuff ! 

I think we can still polish it a bit and revisit the prefetch and query optimisations

backend/hexa/workspaces/schema/queries.py
mrivar: You're right, I tihnk we can keep the improvement without having to change the pagination, by using a better strategy to keep the queries short (without N+1 queries) but without that much complexity

frontend/src/workspaces/features/WorkspaceAccessToken/WorkspaceAccessToken.tsx
mrivar: Sure, not a bad idea. It will just increase the scope of the PR but it can absolutely be done

mrivar: I think I'll skip it for now because it adds too much complexity, but we could open a ticket to tackle it in the soon future

yolanfery: Maybe we can just change the wording 😛 that's cheap and avoid users having to face a wall

mrivar: fair enough! that sounds like a good balance for now 😄 thx!

backend/hexa/workspaces/graphql/schema.graphql
mrivar: Thanks for the extensive feedback! After some iterations, this is what I came up with:
image

frontend/src/workspaces/features/WorkspaceAccessToken/WorkspaceAccessToken.tsx
mrivar: Improved, thx

backend/hexa/workspaces/graphql/schema.graphql
yolanfery: Looking great

backend/hexa/workspaces/models.py
```diff
@@ -462,6 +462,54 @@ def delete_if_has_perm(self, *, principal: User):
...
+    memberships = {
+        membership.workspace_id: membership
+        for membership in WorkspaceMembership.objects.filter(
+            user=user, workspace__in=workspaces
+        ).defer("access_token")
+    }
+    for workspace in workspaces:
+        workspace._current_membership = (user.pk, memberships.get(workspace.id))
+
+
```
yolanfery: [Minor] Why are those not part of the user model ? It seems they always require a user as the main parameter and could be part of the abstraction

frontend/src/workspaces/features/WorkspaceAccessToken/WorkspaceAccessTokenField.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+          canGenerate={false}
+        />
+      </TestApp>,
+    );
+
+    expect(screen.queryByRole("button")).not.toBeInTheDocument();
+    // No role is shown near the field, so the message says why as well
+    expect(
+      screen.getByText(
+        "Viewers cannot generate tokens. Ask a workspace admin for an editor role.",
```
yolanfery, line 77: 👌

backend/hexa/workspaces/models.py
yolanfery: [Minor] its maybe just me but `prime_workspace_memberships` is difficult to understand and it's not clear that it's related to `get_workspace_membership`  , maybe we can have `preload_workspace_memberships` and `get_preloaded_workspace_membership`

Or something different, or keeping it that way if you like it, just raising it was not clear directly for me

yolanfery (approved):

backend/hexa/workspaces/models.py
mrivar: About the user model, the user it is using here may not be a User as we know it as it comes from the request. It could be an AnonymousUser or a ServicePrincipal. We could have it in the workspace tho, which is the one responsible for saving the cache data 🤔
About the `prime_workspace_memberships` you're totally right. It comes from priming (loading) the cache, but I also feel it's not really used that much and it's confusing. I'll change it

yolanfery: The branch `if not isinstance(user, User) or not user.is_authenticated:` can be done at the resolver level so the the method are in the model

yolanfery: But at the workspace model could work also with `prime_workspace_memberships` being static like `preload_workspace_memberships_for_user`

mrivar: Still, I would argue that logic is better inside the function (otherwise you always have to remember to put that logic outside it everytime)
I moved everything to the Workspace model and it looks much better now :)

yolanfery: Ok, sure, no strong opinion against this and it's looking way better already, thanks

yolanfery (approved):

PR #2003, opened 2026-08-17
feat: saved query on delete policy (HEXA-1749)

mrivar:
Follow-up to https://github.com/BLSQ/openhexa-app/pull/1986.

After adding visibility to saved queries (specifically private visibility), deleting a user left private queries orphaned: rows no workspace or organization role grants access to, reachable only from the admin console. This PR decides what happens to them:

- If the user is deleted, their private queries go with them — nobody could reach one anyway, so keeping it means keeping a row for nobody
- Shared queries are kept (the workspace, and the webapps and pipelines built on them, still rely on queries they did not write)
- If a user is only removed from a workspace, the queries are kept: added back to the workspace, they find their queries again. When the account is finally deleted, those hidden queries go too

### Scope: this is not an erasure policy

Worth being explicit, since "delete the user's private queries" reads like a privacy feature and isn't one. The rule implemented here is *leave no row nobody can reach* — it removes queries that became unreachable, not everything the account left behind. `QueryLog` still keeps the verbatim SQL of everything that user ran, with `user` set to NULL: a log entry is worth keeping once nobody is named on it, because what it answers ("how much was this workspace queried, what failed") is about the workspace rather than the person. If we ever need to erase that text too, that is a retention decision about the audit trail as a whole and should be taken as one — not settled quietly by the saved-query deletion policy. Recorded as a comment on `QueryLog.user`.

## Changes

- Added the policy on how saved queries are deleted based on what's written above, as a per-visibility `on_delete` handler on `SavedQuery.created_by`. It dispatches to Django's own `CASCADE` / `SET_NULL`, so the admin's delete-confirmation page nests the affected queries under the user being deleted
- Enforced at the model layer on purpose: user deletion happens through the Django admin or a shell, so a policy living in a service or a mutation would simply be bypassed
- Added a `CheckConstraint` so a private query can never lose its author by any other code path
- Added a migration to clean up orphaned queries from users deleted before this change
- Tests


backend/hexa/data_studio/models.py
```diff
@@ -184,6 +265,12 @@ class Origin(models.TextChoices):
     workspace = models.ForeignKey(
         Workspace, on_delete=models.CASCADE, related_name="query_logs"
     )
+    # SET_NULL, unlike SavedQuery.created_by: a log entry is worth keeping even once
+    # nobody is named on it, because what it answers ("how much was this workspace
+    # queried, what failed") is about the workspace rather than the person. The SQL
+    # text stays with it; dropping that is a retention decision about the audit trail
+    # as a whole, not something the saved-query deletion policy should settle on its
+    # own (see `saved_queries_on_author_deleted`).
```
yolanfery, line 273: Do you think it is worth the complexity to somehow store a shortened version of the user info somehow, so that we can still understand where the queries come from even if the author user is deleted ?

backend/hexa/data_studio/tests/test_models.py
```diff
@@ -216,6 +223,202 @@ def test_content_keeps_literals_verbatim(self):
...
+        query = self.create_saved_query(
+            user=self.USER_VIEWER, visibility=SavedQueryVisibility.PRIVATE
+        )
+
+        self._revoke_membership(self.USER_VIEWER)
+        WorkspaceMembership.objects.create(
+            workspace=self.WORKSPACE,
+            user=self.USER_VIEWER,
+            role=WorkspaceMembershipRole.VIEWER,
+        )
```
yolanfery, line 407: Lets check in between that it's not accessible 😜

yolanfery (approved):
I like the approach 👌

Just a minor comment and a question around audit preservation

backend/hexa/data_studio/models.py
mrivar: I thought about this but my main concern is whether it violates some privacy concern.
If it doesn't, we could keep something like the email/name.
If it does, the only useful thing I can see adding in the admin UI, a label indicating that the original owner's account is deleted if the user is null.
wdyt?

backend/hexa/data_studio/tests/test_models.py
mrivar: Good one! Adding it, thx

backend/hexa/data_studio/models.py
yolanfery: Ok we can keep things simple and private, this should be a low probability issue 👍

backend/hexa/data_studio/tests/test_models.py
yolanfery: Thanks, looking good to merge

PR #2006, opened 2026-08-18
feat: Map widget in data studio (HEXA-1779)

DimitriKwihangana:
Data Studio results can now be drawn on a map, and the editor's panels can be resized. The map is opt-in the same way charts are: alias your columns to a convention and the result is drawn instead of listed.


bramj (commented):
@DimitriKwihangana I fully agree with this feature 😄 really cool!!

Not a comprehensive review of this PR, but my main question is about the dependency we're adding: since we're on a clean slate here (we don't have mapping libs in OH yet), I wondered if `leaflet` is still the best tool for the job in 2026:
- The repo appears to be not very maintained https://github.com/PaulLeCam/react-leaflet/releases (any reason why you didn't pick the latest version?)
- Feedback from colleagues with lot's of geo experience points to preferring [maplibre](https://github.com/maplibre/maplibre-gl-js) and [deck.gl](http://deck.gl/) (see Slack conversation on internal support channel). Can you check what this feature would look like with these libraries instead?

Minor point: I feel like we need to refactor the Data Studio frontend a bit, check out the amount of files in  in the `frontend/src/workspaces/features/DataStudioEditor` path compared to the other feature dirs. Maybe we can start by moving the widget-related components to a subfolder?

DimitriKwihangana:
> @DimitriKwihangana I fully agree with this feature 😄 really cool!!
> 
> Not a comprehensive review of this PR, but my main question is about the dependency we're adding: since we're on a clean slate here (we don't have mapping libs in OH yet), I wondered if `leaflet` is still the best tool for the job in 2026:
> 
> * The repo appears to be not very maintained https://github.com/PaulLeCam/react-leaflet/releases (any reason why you didn't pick the latest version?)
> * Feedback from colleagues with lot's of geo experience points to preferring [maplibre](https://github.com/maplibre/maplibre-gl-js) and [deck.gl](http://deck.gl/) (see Slack conversation on internal support channel). Can you check what this feature would look like with these libraries instead?
> 
> Minor point: I feel like we need to refactor the Data Studio frontend a bit, check out the amount of files in in the `frontend/src/workspaces/features/DataStudioEditor` path compared to the other feature dirs. Maybe we can start by moving the widget-related components to a subfolder?

Thanks @bramj, I just saw the conversation from slack in the daily. I'm currently refactoring to MapLibre with [deck.gl](http://deck.gl/).

bramj:
This is great work @DimitriKwihangana ! The rebased PR (was a bit tricky because the panel resizing being a part of it as well) is here: https://github.com/BLSQ/openhexa-app/pull/2049. Added a (very) small improvement and going to merge now.

:clap: :rocket:

PR #2007, opened 2026-08-18
fix: improve webapps for big files (HEXA-1745)

yolanfery:
Ensure we support HTTP range by using streaming for all file types except HTML. 

## Changes

- Detect if a file needs to be compressed or not (binaries don't need to be compressed), to enable using HTTP ranges
- Raise an explicit exception on file too big
- Support HTTP range by forwarding headers between OH and the Git server
- Improve UI in the code editor to correctly show the file type/size and a message when the file is too large to be viewed


backend/hexa/git/forgejo.py
```diff
@@ -191,9 +193,40 @@ def get_file(
             if e.status_code == 404:
                 raise GitFileNotFound(path) from e
             raise
-        content = response.json().get("content", "")
+        payload = response.json()
+        content = payload.get("content", "")
+        size = payload.get("size", 0)
+        if not content and size > 0:
+            raise GitFileTooLarge(path, size)
```
mrivar, line 213: Nice one!
I found 2 instances in which we don't catch this exception, in webapps/models.py:353 and :366, not sure what the behaviour should be but it's worth a check

backend/hexa/webapps/views.py
```diff
@@ -92,32 +97,86 @@ def auth_token(request, webapp_id):
...
+
+def _guess_content_type(path):
+    content_type, _ = mimetypes.guess_type(path)
+    return content_type or "application/octet-stream"
+
+
+def _drain_upstream(upstream):
+    """Yield the upstream body, releasing the connection once it is exhausted."""
+    try:
+        yield from upstream.iter_content(chunk_size=STREAM_CHUNK_SIZE)
```
mrivar: There's a possible hidden silent bug here: `iter_content` silently uncompresses gzip files, but we keep the header the same. Thus, for example, if the file we're trying to get is a 100kb gzip and it is decompressed into a 500kb file, we return a truncated broken file.
I think it wouldn't be an issue atm, but if we activate gzip compression in Forgejo for example and this hasn't been delt with it might be an issue one day. Worth a look/test.

backend/hexa/git/forgejo.py
```diff
@@ -191,9 +193,40 @@ def get_file(
...
+            url,
+            params={"ref": ref},
+            headers=headers or {},
+            stream=True,
+            allow_redirects=False,
+        )
+        if response.status_code == 404:
+            response.close()
+            raise GitFileNotFound(path)
+        if not response.ok and response.status_code != 416:
```
mrivar: allow_redirects=False + response.ok is true for 3xx. Might be worth adding a catcher for 3xx responses, but not really a blocker (it was already existent in get_file), your call

backend/hexa/git/forgejo.py
```diff
@@ -191,9 +193,40 @@ def get_file(
...
+        self,
+        repo_name: str,
+        path: str,
+        ref: str = "main",
+        *,
+        org_slug: str | None = None,
+        headers: dict[str, str] | None = None,
+    ) -> requests.Response:
+        org_slug = org_slug or self._username
+        url = f"{self._url}/api/v1/repos/{org_slug}/{repo_name}/raw/{path}"
```
mrivar: Right now if a user slips a file path with ? into it, it could override our own params. E.g. if there is a request `GET https://myapp.webapps.openhexa.org/config.js%3Fref=main`, we would ask for `http://forgejo:3000/api/v1/repos/org/repo/raw/config.js?ref=main`, that would substitute any value we add of ref for main. A quick quote can fix it, so I'd say it's worth it:

url = f"{self._url}/api/v1/repos/{org_slug}/{repo_name}/raw/{quote(path, safe='/')}"

Also found the same behaviour (pre-existing) in backend/hexa/git/forgejo.py:189

frontend/src/workspaces/features/FilesEditor/FileEditorPane.tsx
```diff
@@ -46,6 +48,25 @@ const FileEditorPane = ({
...
+
+  const metaParts = [
+    fileTypeLabel,
+    lineCount != null
+      ? `${lineCount} ${lineCount > 1 ? t("lines") : t("line")}`
+      : size != null
+        ? filesize(size)
+        : null,
+    currentFileIsModified ? t("Modified") : null,
+  ].filter(Boolean);
```
mrivar, line 68: would be nice to have both size and lines instead of only one of them

frontend/src/workspaces/features/FilesEditor/FileEditorPane.tsx
```diff
@@ -148,19 +165,33 @@ const FileEditorPane = ({
...
-          <CodeMirrorClient
-            key={selectedFile.id + (isDiffMode ? "-diff" : "")}
-            value={currentFileContent}
-            readOnly={!isEditable}
-            onChange={onContentChange}
-            extensions={extensions}
-            height="100%"
-            style={{ width: "100%", height: "100%" }}
-          />
+      {isTooLarge ? (
```
mrivar, line 168: would we want to add a warning like this one in the FileTree?

backend/hexa/webapps/views.py
```diff
@@ -92,32 +97,86 @@ def auth_token(request, webapp_id):
...
-
-    return HttpResponse(content, content_type=content_type)
+    except GitFileTooLarge as e:
+        logger.error(
+            "Webapp %s cannot serve %s: %s bytes exceeds the git API blob limit",
+            git_webapp.pk,
+            path,
+            e.size,
+        )
+        return HttpResponseServerError("File too large to be served")
```
mrivar, line 179: WDYT of switching it to status 413 Content Too Large?

backend/hexa/webapps/views.py
```diff
@@ -92,32 +97,86 @@ def auth_token(request, webapp_id):
...
+
 def serve_webapp(request, git_webapp, path="index.html"):
     if not is_safe_path(path):
         return HttpResponseBadRequest("Invalid path")
 
     if not git_webapp.published_commit:
         return HttpResponseNotFound("No published version")
 
+    content_type = _guess_content_type(path)
     client = get_forgejo_client()
```
mrivar, line 148: Worth adding an ETag + If-None-Match short-circuit before we call Forgejo:

```python
etag = f'"{git_webapp.published_commit}:{path}"'
  if request.headers.get("If-None-Match") == etag:
    return HttpResponseNotModified()
```

In order to avoid doing an extra round-trip to forgejo if we already have the file cached. But up to discussion; no strong opinion here.
Also, if this is added, we have to check on the upstream.close() in _serve_stream, as it would leak a connection on a 304 path

mrivar (changes requested):
Great feature! Just some small improvements and suggestions 😄

backend/hexa/webapps/views.py
yolanfery: 413 is for a request that is too large, here the response is the issue, so I think a server error is fine here

backend/hexa/git/forgejo.py
yolanfery: Improved, thanks

backend/hexa/webapps/views.py
yolanfery: Improved, thanks

backend/hexa/git/forgejo.py
yolanfery: Improved, thanks

backend/hexa/git/forgejo.py
yolanfery: Yes, better to fix it, good catch

frontend/src/workspaces/features/FilesEditor/FileEditorPane.tsx
yolanfery: Not aligned with this, I think it makes sense to keep the size of the file for binaries and number of lines for code

frontend/src/workspaces/features/FilesEditor/FileEditorPane.tsx
yolanfery: I personally think it brings a bit of unneeded clutter so I left it like it was

backend/hexa/webapps/views.py
yolanfery: There is already the Etag mechanism at a higher level but I improved it a bit, thanks

frontend/src/workspaces/features/FilesEditor/FileEditorPane.tsx
mrivar: Sure, no prob. It is a visual thing so in case we feel the need in the future it's a quick thing anyway.

mrivar (approved):
Looks good to me, thanks for the improvements!! 🚀

PR #2008, opened 2026-08-19
fix: git webapp clone sometimes directing users to Github (HEXA-1745)

yolanfery:
Just add a command ensuring GCM picks our origin domain and not Github's one. Also ensure that wrong scope token are returning 401, giving a chance for the Git client to recover from the issue


bramj (approved):
Good catch! 👍

PR #2010, opened 2026-08-21
fix: Workspace copier small improvements (HEXA-1756)

bramj:
Several small improvements to the workspace copier:

- Notebook pipeline copying: First copy the `.ipynb` file to avoid notebook pipeline copies from breaking when done _without_ doing the file copying first. (HEXA-1756)
- Pipeline copier: don't copy the schedule metadata: This can cause pipelines to immediately start running after a copy.
Usually you want a human to validate, change some connection settings etc, and then manually enable pipelines one by one.
- Output rendering:
  - Add timestamps
  - Human-readable file sizes for dataset files
  - fix: Also render output on failed run (for Django admin)

**Notes:**
- Intentionally light on testing, we'll redo the whole test approach soon so didn't want to spend time on this.
- Battle-tested on POLIO WS copy

### DIFF
--- a/backend/hexa/workspace_copier/resources/pipelines.py
+++ b/backend/hexa/workspace_copier/resources/pipelines.py
@@ -10,6 +10,7 @@ rather than relying on the files copier having run.
 """
 
 import base64
+import tempfile
 from typing import Any
 
 import httpx
@@ -22,7 +23,7 @@ from hexa.workspace_copier.options import CopyOptions
 from hexa.workspace_copier.progress import ProgressReporter
 from hexa.workspace_copier.resources.base import ResourceCopier
 from hexa.workspace_copier.resources.files import download, upload
-from hexa.workspace_copier.results import CopyResult, PipelinesResult
+from hexa.workspace_copier.results import CopyResult, PipelinesResult, format_bytes
 from hexa.workspace_copier.transport import GraphQLError, gql
 
 PIPELINES_PAGE_SIZE = 50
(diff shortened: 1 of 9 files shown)


backend/hexa/workspace_copier/resources/pipelines.py
```diff
@@ -277,6 +296,19 @@ def _fetch_source_detail(source: Client, pipeline_id: str) -> dict[str, Any]:
...
 
+def _copy_notebook_file(
+    source: Endpoint,
+    target: Endpoint,
+    notebook_path: str,
+    http_client: httpx.Client,
+    reporter: ProgressReporter,
+) -> None:
+    """Transfer a notebook pipeline's ``.ipynb`` to the target bucket."""
+    content = download(source.client, source.slug, notebook_path, http_client)
```
mrivar: The `download` function has 5 parameters, it seems we are missing `destination` here. In the tests it's unreachable as it mocks `_copy_notebook_file`, although if it worked for POLIO WS, then it leaves me confused. Wouldn't it break when running?
Maybe I'm missing something?

mrivar (changes requested):
Looking good! I just have a small question that left me baffled 😵‍💫 but maybe I'm missing something?
In case it's a problem it's a quick change luckily

backend/hexa/workspace_copier/resources/pipelines.py
bramj: Oops! Good catch, I actually ran into this issue when copying for the POLIO WS, did the commit and forgot to push it :see_no_evil: 

Just added it, should be good now :+1:

mrivar: Perfect! Looks good to me then

mrivar (approved):

PR #2011, opened 2026-08-21
fix: UI of navbar with limited width (HEXA-1750)

yolanfery:
Ellipsis on breadcrumbs and tooltip on hover + small search box on smaller screen 

Screenshot 2026-08-21 at 10 36 12


DimitriKwihangana (approved):
It looks much better!

PR #2012, opened 2026-08-21
feat: allow left and bottom panels to be adjustable(HEXA-1778)

DimitriKwihangana:
Allow resizing o fthe panels in Data studio
image


yolanfery:
Since Dimitri is OOO, I did some changes to : 
- Define a reusable component instead of a hook
- Some UI improvements
- Use cookies instead of local storage, to paint the panel at the right size directly to avoid flickering when the page is hydrated

mrivar (approved):
Good work! It looks and works very nicely.
Good job on the refactoring to have a reusable component as well.

It would be nice if when running the data studio, and the results block is closed, it would open again? Otherwise the page feels stale when pressing run.
Not sure tho, and it can be left for a future improvement anyway.

yolanfery:
> It would be nice if when running the data studio, and the results block is closed, it would open again? Otherwise the page feels stale when pressing run.

Great point ! Who wants to run a query and not see the result ? 😅

Added, thanks

PR #2014, opened 2026-08-22
fix: unit tests leak database tables (HEXA-1785)

yolanfery:
Unit tests were not cleaning up created database tables causing bloat to docker disk. Add a method to provision the database on demand while automatically registering a cleanup method.


backend/hexa/workspaces/tests/test_schema/test_workspace.py
```diff
@@ -125,24 +126,20 @@ def setUpTestData(cls):
             role=OrganizationMembershipRole.ADMIN,
         )
 
-        with (
```
yolanfery, line 128: We get a lot of simplification in this file 😁

backend/hexa/workspaces/tests/test_schema/test_workspace.py
```diff
@@ -28,6 +28,7 @@
     WorkspaceMembership,
     WorkspaceMembershipRole,
 )
+from hexa.workspaces.tests.testutils import create_workspace
```
mrivar, line 31: I think for the tests in this file, we have reversed the logic. They used to patch creating the db and now they actually create it and leak it, as they use the mutation `createWorkspace` instead of the new `provision_db_on` improvement. Worth checking

mrivar (dismissed):
Amazing improvements!! Our computers will be very thankful 🙂

Just one little catch, we might have reversed the logic in the tests in `test_workspace.py‎`

backend/hexa/workspaces/tests/test_schema/test_workspace.py
yolanfery: Good catch, the mutation was not patched anymore. Fixed it with a test class level patcher

PR #2015, opened 2026-08-24
fix: Set tool limit to 30, 10 was too low (HEXA-1786)

bramj:
Agents would burn their tool limit while looking for files for example.

https://logfire-eu.pydantic.dev/openhexa/openhexa-app/issues/19


mcrimi (approved):
lgtm

PR #2027, opened 2026-08-26
fix: make it possible to delete webapp/pipeline files from UI and webassistant (HEXA-1774)

yolanfery:
Enable deleting files from the UI and enable the assistant to propose deletions

## Screenshots / screencast
Screenshot 2026-08-26 at 16 07 20


backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -28,10 +32,13 @@ def propose_pipeline_version(
...
     Always call this tool with the files you modified or created in modified_files.
     Each entry must have a 'name' (e.g. 'pipeline.py') and 'content' (the full file text).
-    List any files to remove in deleted_files.
+    List any files to remove in deleted_files. A directory path removes everything under it.
+    Deletions cover binary files too, even though their content is never inlined here.
     Unchanged files are preserved automatically.
     """
     current_files: dict[str, str] = {}
+    deleted_paths: set[str] = set()
+    archived_paths: set[str] = set()
```
mrivar: I wonder if a naming such as `known_paths` would be more readable than `archived_paths` (Took me a bit to understand what it was initially). Also I wonder if `all_paths` should match the name for symmetry

backend/hexa/assistant/agents/edit_pipeline_agent.py
```diff
@@ -149,6 +167,16 @@ def _extra_instructions(self) -> str:
 
         if pending and pending.tool_output:
             proposed_files = pending.tool_output.get("files", [])
+            pending_deletions = pending.tool_output.get("deleted_paths", [])
+            if pending_deletions:
+                lines.append("")
+                lines.append(
+                    "### Files Already Marked For Deletion (Pending)\n"
+                    "Do not list these again in `deleted_files`; they are already removed "
+                    "by the pending proposal."
```
mrivar: Not sure if it's an issue and also the solution to it might be too overcomplicated, but this assertion might not be true, as the user might have "rejected" the delete by signaling it from the UI.
Again, taking that into account might be too complicated and I'm not sure how much on an issue it actually is, but just dropping the idea.

mrivar (commented):
Looks super cool! I like seeing the coding UI being more advanced, even if it's not supposed to be the default IDE, it's nice to have it with a better more friendly vibe.

I added 2 things for discussion plus I have an extra suggestion:
As pipelines break if there's no `pipeline.py` or if there's no files at all, I wonder if we should add a warning in the UI when the user tries to do that, so at least he realizes he might be breaking the pipeline by deleting those files.

I'll leave it as a comment rather than approving or requesting changes, as I feel these observations aren't that much of an issue but are worth thinking about them 😄

backend/hexa/assistant/agents/edit_pipeline_agent.py
yolanfery: Makes sense, thanks

backend/hexa/assistant/agents/edit_pipeline_agent.py
yolanfery: Good catch, improved it a bit

yolanfery:
> I have an extra suggestion: As pipelines break if there's no `pipeline.py` or if there's no files at all, I wonder if we should add a warning in the UI when the user tries to do that, so at least he realizes he might be breaking the pipeline by deleting those files.

Not totally convinced, and this would prevent users "starting fresh" by deleting everything. I will leave like it is now and we notice users facing this we can add this later, its a small change

PR #2029, opened 2026-08-26
chore: Add skill to help draft the release message

bramj:
This was already something I used on my local machine, but I think it makes sense to make it a shared resource.

You prompt it with:
- the version about to be released
- the current version on prod

And it drafts the release message in our usual format, for a human to review.

This is a v1, the output is quite decent already and saves some time:

Screenshot 2026-08-26 at 14 59 27


yolanfery (approved):
Great idea to share it, thanks !! I will try it next time I need to release

A few optional ideas to improve it 
- Give it a complete and actual example from the past
- Ask it to start with an informal summary line catching attention and containing the old version number and the new one (with a link to the release page in github)
- I like it when we include the Jira ticket number (HEXA-XXXX) linking to the actual PR per line or to the commit in github if missing. It is easier to get more context 
-  One day maybe : Using playwright MCP to generate screenshots of the features

bramj:
> One day maybe : Using playwright MCP to generate screenshots of the features

Now that's some AI-maximalism right there :grin: 

Great ideas for improving it :+1: I'll just merge it now, and we all try to improve on it every time we do a release.

PR #2030, opened 2026-08-26
fix: Flicker on workspace tags modal + small mistake on documentation

bramj:
- fix: Correct small mistake in web apps documentation: On how to use the `executeSavedQuery` endpoint.
- fix: Flicker on the workspace tags modal

### DIFF
--- a/docs/en/static-webapps.md
+++ b/docs/en/static-webapps.md
@@ -989,7 +989,6 @@ type Query {
 }
 
 input ExecuteSavedQueryInput {
-  workspaceSlug: String!
   slug: String!
   maxRows: Int
 }
@@ -1020,8 +1019,8 @@ enum ExecuteSQLError {
 </details>
 
 Find the slug in the Data Studio: it is the last segment of the saved query's
-URL. `SAVED_QUERY_NOT_FOUND` covers both an unknown slug and a workspace you
-cannot see, so it does not reveal what exists.
+URL. Note that the query must be shared with the workspace, a private one fails
+with `SAVED_QUERY_NOT_FOUND`.
 
 ```html
 <!DOCTYPE html>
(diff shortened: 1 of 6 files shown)


docs/en/static-webapps.md
```diff
@@ -1020,8 +1019,8 @@ enum ExecuteSQLError {
 </details>
 
 Find the slug in the Data Studio: it is the last segment of the saved query's
-URL. `SAVED_QUERY_NOT_FOUND` covers both an unknown slug and a workspace you
-cannot see, so it does not reveal what exists.
+URL. Note that the query must be shared with the workspace, a private one fails
+with `SAVED_QUERY_NOT_FOUND`.
```
yolanfery, line 1023: French documentation need updating also then

yolanfery (approved):

docs/en/static-webapps.md
bramj: 🤦

bramj:
@yolanfery To see if it had any impact, I repeated my same initial prompt, but starting with just the last commit ([change to CLAUDE.md](https://github.com/BLSQ/openhexa-app/pull/2030/commits/62dae0e7dcb1a24f895696fb6f2e9111b464928d)) on main. 
This time, it actually proposed me to fix both EN and FR versions, that's great!

I think it's worth it to try and add bits of info to it when we see it fail like this.

yolanfery:
> @yolanfery To see if it had any impact, I repeated my same initial prompt, but starting with just the last commit ([change to CLAUDE.md](https://github.com/BLSQ/openhexa-app/pull/2030/commits/62dae0e7dcb1a24f895696fb6f2e9111b464928d)) on main. This time, it actually proposed me to fix both EN and FR versions, that's great!
> 
> I think it's worth it to try and add bits of info to it when we see it fail like this.

Indeed thanks !

PR #2033, opened 2026-08-27
fix: too much data for declared Content-Length

yolanfery:
Fix for https://bluesquareorg.sentry.io/issues/7695207720

Ensure that 304 responses are not requiring compression since they have empty body. That would result in a body greater than the original one.


bramj (approved):
Good catch! 👍

PR #2034, opened 2026-08-27
fix: ensure all buckets have the same cors rules (HEXA-1794)

yolanfery:
Default CORS values where changed [2 years ago](https://github.com/BLSQ/openhexa-app/commit/1362f0dcf78cf69a7d49f244912c7a86e495e40f)... but older buckets were not migrated


backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
```diff
@@ -0,0 +1,56 @@
...
+    """Re-apply the storage backend CORS policy to every existing workspace bucket.
+
+    The policy was only ever written by create_bucket, so buckets created before
+    the wildcard policy landed (commit 1362f0dc, July 2024) kept the original
+    one, restricted to app.openhexa.org / accessmod.openhexa.org and to
+    GET + PUT. Browser uploads and downloads from any other origin fail their
+    preflight on those buckets.
+    """
+    if not isinstance(storage, SupportsBucketCors):
+        return
```
bramj, line 21: Can this give issues for S3? Maybe it is safer to skip this migration for all local hostings?

backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
```diff
@@ -0,0 +1,56 @@
...
+        .exclude(bucket_name="")
+        .values_list("bucket_name", flat=True)
+    )
+    if not bucket_names:
+        return
+
+    failed = []
+    for bucket_name in bucket_names:
+        try:
+            storage.set_bucket_cors(bucket_name)
```
bramj, line 35: For S3, should we pass via `_apply_bucket_config`? 

https://github.com/BLSQ/openhexa-app/blob/main/backend/hexa/files/backends/s3.py#L147

backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
```diff
@@ -0,0 +1,56 @@
...
+        except Exception as exc:
+            failed.append(bucket_name)
+            print(f"[0064] Skipping {bucket_name}: {exc}")
+
+    if len(failed) == len(bucket_names):
+        raise RuntimeError(
+            f"[0064] CORS backfill failed for all {len(bucket_names)} buckets. "
+            "Check the storage credentials and that the role is allowed to "
+            "update bucket configuration."
+        )
```
bramj, line 51: What do you think of failing with a simple log here? I think this particular problem is not an issue on localhostings, but I suppose the migration could fail when the storage backend is S3.

bramj (approved):
Nice one, good refactoring. Just a few questions about S3, I'd prefer to never have the migration break a deployment.

backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
yolanfery: There is a try catch for individual calls so that if there is failure, it's not blocking

backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
yolanfery: There is already a try catch, no need for wrapping it with one more

backend/hexa/workspaces/migrations/0064_backfill_bucket_cors.py
yolanfery: Totally makes sense yes, I changed it, thanks

PR #2035, opened 2026-08-28
fix: attribute issue for unauthenticated error

yolanfery:
Sentry is reporting some issue for unauthenticated user


mrivar (approved):
Good one, didn't think of this, I wonder if we have any similar issues anywhere else in the code

yolanfery:
> Good one, didn't think of this, I wonder if we have any similar issues anywhere else in the code

A quick pass with Claude revealed nothing relevant for now

PR #2036, opened 2026-08-28
feat: Add the workspace description to the agent's base context (HEXA-1795)

bramj:
A quick v1 to help our colleagues in the Ebola response.

A proper workspace-specific way of adding instructions can come later, but for now this can already be a powerful tool.

Note: it will be very powerful if the agent can also _edit_ the WS-specific instructions, but let's keep that for when we have a better solution.

## Changes

- Inject `workspace.description` into every assistant agent's system prompt as an extra block.
- Blank and untouched default boilerplate descriptions are skipped. 
- Descriptions with > 10k characters are truncated with a note pointing at the `get_workspace` tool
- `get_workspace` tool is now available to all agents
- Refactor to allow setting a list of tools that are common on all agents

## How/what to test

Add instructions to the WS description and play with the agent.

## Screenshots / screencast

image
image


backend/hexa/assistant/agents/base.py
```diff
@@ -204,6 +208,55 @@ def strip_proposals(messages: list[ModelMessage]) -> list[ModelMessage]:
...
+            "<workspace_description>\n"
+            f"{description}\n"
+            "</workspace_description>"
+            f"{truncation_note}"
+        )
+
+    @staticmethod
+    def _is_default_description(workspace, description: str) -> bool:
+        if DEFAULT_WORKSPACE_DESCRIPTION_SENTINEL in description:
+            return True
```
yolanfery, line 252: Not sure we want this, the user could have modified part fo the description and left this there. I would just keep comparing the default value with the current and drop this sentinel check

yolanfery: Or at least the SENTINEL shouldn't be a magic string and actually be injected in the default description to ensure maintenance

backend/hexa/assistant/agents/base.py
```diff
@@ -160,6 +167,8 @@ def _strip_proposal_outputs(
 
 class BaseAgent:
     instruction_set = InstructionSet.GENERAL
+    # Tools every agent gets, regardless of instruction set.
+    common_tools: list = [get_workspace]
```
yolanfery, line 171: Slowly getting where all agents will have all get tools 😜

yolanfery (approved):
Sounds like a great start and to unblock the team

backend/hexa/assistant/agents/base.py
bramj: Indeed!

backend/hexa/assistant/agents/base.py
bramj: I agree, but just about the maintenance point: this logic should be reverted soon enough. I think we don't want to advertise this new feature, but rather do a proper implementation first, where we have dedicated "workspace instructions" that are versioned.

PR #2039, opened 2026-09-01
chore: Truncate the preview_dataset_file tool (internal agent + MCP) (HEXA-1796)

bramj:
While looking into traces on Logfire, the `preview_dataset_file` tool appears as a frequent cause of context overload.

## Changes

Changes are only to the tool, the GraphQL endpoint used by the frontend is not affected.

- Remove separate "columns" key with SHAs (useless for an LLM), keep ordering intact in the sample's keys
- Drop sample from 50 to 3 rows

## Impact on a real test dataset

**Before: `62,676 characters, ~15,669 tokens`**
```json
{
  "data": {
    "datasetVersionFile": {
      "id": "4360b152-e949-4990-b5b9-fe07a0c6d0a0",
      "filename": "something.csv",
      "contentType": "text/csv",
      "size": 20588,
      "rows": 70,
      "properties": {
        "columns": {
          "08ac28ead0d7f1d59eabb25471945121": "ITN_ACCESS_CI_UPPER_BOUND",
          "1e85baf2ade4293360c0b6665b431188": "MORBIDITY_RISK1",
          "20603301afda409105d1f3e8757d60e0": "ADM2_ID",
          "24bd86e0ac5ff5bc896dedfb70b83990": "ADM2",
          ...
        },
        "column_order": [
          "9d9e86a6e63f728c8354f129f0e7dd4e",
          "772dba222917155b6b905f825e04be33",
          "24bd86e0ac5ff5bc896dedfb70b83990",
          ...
        ]
      },
      "fileSample": {
        "sample": [
          ... 50 samples
        ],
        "status": "FINISHED",
        "statusReason": null
      }
    }
  }
}
```

**After: `3,525 characters, ~881 tokens`**

```json
{
  "id": "4360b152-e949-4990-b5b9-fe07a0c6d0a0",
  "filename": "BFA_WebApp.csv",
  "contentType": "text/csv",
  "size": 20588,
  "rows": 70,
  "fileSample": {
    "sample": [
      {
        "ADM1": "CENTRE OUEST",
        "ADM1_ID": "TMFuUYJyGPX",
        "ADM2": "DS NANORO",
        ...
      },
      ... 2 more samples
    ],
    "status": "FINISHED",
    "statusReason": null,
    "sampleRowsAvailable": 50
  }
}
```

### DIFF
--- a/backend/hexa/mcp/tools/datasets.py
+++ b/backend/hexa/mcp/tools/datasets.py
@@ -92,7 +92,7 @@ def _ordered_row(row: dict, columns: list) -> dict:
 
 @tool
 def preview_dataset_file(user, file_id: str) -> dict:
-    """Preview the content of a dataset file by its ID (from get_dataset's file list). Returns file metadata and, for tabular files (CSV, Parquet, etc.), the first few rows of the stored sample, with the columns of each row in the order they appear in the file. When no sample row is available, the ordered 'columns' of the file are returned instead. This is a preview only: 'rows' is the row count of the whole file and 'fileSample.sampleRowsAvailable' the size of the stored sample, both usually larger than the number of rows returned here. The sample status can be PROCESSING (still generating), FINISHED (sample ready), or FAILED."""
+    """Preview the content of a dataset file by its ID (from get_dataset's file list). Returns file metadata and, for tabular files (CSV, Parquet, etc.), the first few rows of the stored sample, with the columns of each row in the order they appear in the file. When no sample row is available, the ordered 'columns' of the file are returned instead. This is a preview only: 'rows' is the row count of the whole file, usually larger than the number of rows returned here. The sample status can be PROCESSING (still generating), FINISHED (sample ready), or FAILED."""
     data = execute_graphql(user, "PreviewDatasetFile", {"id": file_id})
     if "errors" in data:
         return data
@@ -107,7 +107,6 @@ def preview_dataset_file(user, file_id: str) -> dict:
         file_sample["sample"] = [
             _ordered_row(row, columns) for row in rows[:PREVIEW_SAMPLE_ROWS]
         ]
-        file_sample["sampleRowsAvailable"] = len(rows)
     # The sample rows already name every column, so `columns` would only repeat them.
     if not rows and columns:
         file_data["columns"] = columns
--- a/backend/hexa/mcp/tools/datasets.py
+++ b/backend/hexa/mcp/tools/datasets.py
@@ -4,6 +4,10 @@ from hexa.mcp.protocol import tool
 
 from ._graphql import execute_graphql
 
+# LLM callers only need enough rows to infer the schema and value shapes; the stored
+# sample (WORKSPACE_DATASETS_FILE_SNAPSHOT_SIZE rows) is sized for the UI table.
+PREVIEW_SAMPLE_ROWS = 5
+
 
 @tool
 def list_datasets(user, workspace_slug: str, page: int = 1, per_page: int = 10) -> dict:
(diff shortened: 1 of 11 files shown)


backend/hexa/mcp/tools/datasets.py
```diff
@@ -55,15 +59,58 @@ def get_dataset(
...
         return {"error": "Dataset file not found"}
+
+    columns = _column_names(file_data)
+    rows = _sample_rows(file_data)
+    file_sample = file_data.get("fileSample")
+    if file_sample:
+        file_sample["sample"] = [
+            _ordered_row(row, columns) for row in rows[:PREVIEW_SAMPLE_ROWS]
+        ]
+        file_sample["sampleRowsAvailable"] = len(rows)
```
yolanfery: What's the goal of having `sampleRowsAvailable` ?

backend/hexa/mcp/tools/datasets.py
```diff
@@ -55,15 +59,58 @@ def get_dataset(
...
     if file_data is None:
         return {"error": "Dataset file not found"}
+
+    columns = _column_names(file_data)
+    rows = _sample_rows(file_data)
+    file_sample = file_data.get("fileSample")
+    if file_sample:
+        file_sample["sample"] = [
+            _ordered_row(row, columns) for row in rows[:PREVIEW_SAMPLE_ROWS]
+        ]
```
yolanfery: Don't you think it would be valuable to keep the MCP layer as thin as possible and add this kind of logic to the graphql resolver ? 

This reduces the risk of duplicating logic for the FE (or SDK client) and the MCP and other clients get improvements for free

yolanfery (commented):
Sounds like a fair direction 👍

Just not sure about the need for `sampleRowsAvailable ` and the implementation approach (versus doing more in the GQL layer)

backend/hexa/mcp/tools/datasets.py
bramj: Absolutely right, that's not useful

backend/hexa/mcp/tools/datasets.py
bramj: Fair point. I didn't want to touch the frontend for this, but it would be nice of the row part can be moved there, and the MCP layer just reduces the sample size.
Let me check if that's doable, not sure if there's a reason for the FE to require the format as it is right now.

yolanfery: To be fair, I think at some point we will not escape having to add custom logic in the MCP layer but I would rather delay this as much as possible if we can (?), I feel like it's the safest bet for consistency and future maintenance

yolanfery: For the sample size, this can be a parameter of the resolver also 

Doing `min(param_sample_size, env_sample_size)` until we support a more dynamic (on-demand) sample generation with a given size

bramj: Agreed on that. I think the verbosity of GraphQL in general doesn't play that well with LLM contexts, but let's delay big changes on this as long as we can :+1:

bramj: @yolanfery Okay so it does add quite a bit of complexity.

This is adding the `columns` field to the GraphQL endpoint and changing the frontend to use it: https://github.com/BLSQ/openhexa-app/pull/2039/changes/d11eff2350a58b8103e48df47883870d673d768f..a203eccf378f5544d6a0b1fe753aea5f1376e45a
Unfortunately, that's just the sample table tab, the "columns" tab (that shows the basic statistics on each column) still needs the `properties` key with the hashes. Changing that is a lot more involved.

I'm not really convinced this is worth it and I propose to revert and keep the changes in the MCP layer. WDYT?

yolanfery: Actually I like the end result, complexity is removed from part of the frontend (in https://github.com/BLSQ/openhexa-app/pull/2039/changes#diff-524fd395b783e0636b5468f600b05da277b047709b5c92b4e71c9c45b50023e2) and from the MCP layer, I think it's a great sign ! We can leave the rest of the frontend as it is.

I would also go one step further an parametrize the sample size (like commented in https://github.com/BLSQ/openhexa-app/pull/2039#discussion_r3911864278) but I am not blocking for this since it's a partial parametrisation since it can't exceed the env variable.

yolanfery (approved):

backend/hexa/mcp/tools/datasets.py
bramj: Alright, let's go like this then!

> I would also go one step further an parametrize the sample size (like commented in https://github.com/BLSQ/openhexa-app/pull/2039#discussion_r3911864278) but I am not blocking for this since it's a partial parametrisation since it can't exceed the env variable.

Woops I missed that comment earlier, my bad. I've thought about parametrizing it, but I don't really see the value right now. Like you say it's partial, and if we ever want to make the samples somehow dynamic in size, an API like that would make sense. But I doubt we'll ever want that, so I vote to not add it now.

yolanfery: Sure no worries

PR #2049, opened 2026-09-02
feat: map widget for Data Studio (HEXA-1779)

bramj:
**Note: This PR is a re-implementation of https://github.com/BLSQ/openhexa-app/pull/2006. The history became to complicated because of parts of the panel resizing in the code. Rebasing was too complicated, so I started a new branch, keeping the old one just in case.**

Queries that alias columns to `map_latitude/map_longitude` or `map_geometry` render as a MapLibre map in a first tab, with the table one tab away, the same way charts already work. Raw PostGIS WKB is detected and the tab explains the `ST_AsGeoJSON` fix instead of silently falling back to a table.

Chart and table widgets move into a widgets/ folder alongside the map. MapLibre's tiling worker is copied into `public/maplibre/` before dev and build, since Next cannot resolve it from `import.meta.url`.

**Shapes:**
Screenshot 2026-09-02 at 11 16 23

**Points (10K no sweat):**
Screenshot 2026-09-02 at 11 18 30


frontend/src/workspaces/features/DataStudioEditor/widgets/chart.ts
bramj, line 1: Nice, lot cleaner with the `widgets` folder 👍

docs/en/sql-widgets.md
```diff
@@ -13,19 +13,40 @@ The chart appears in a first tab and the usual tabular output stays in a second
 | Bar | `bar_label`, `bar_quantity` | Comparing a measure across categories (districts, facilities, age groups) |
 | Line | `line_x`, `line_y` | A trend over an ordered axis (months, weeks, epidemiological periods) |
 | Pie | `pie_label`, `pie_quantity` | Part-to-whole, when there are only a few categories |
+| Map | `map_latitude`, `map_longitude` | Places with known coordinates (facilities, households, survey points) |
+| Map | `map_geometry` | Areas rather than points (districts, catchment areas, regions) |
```
bramj, line 17: Perfect, so cool 👍

PR #2050, opened 2026-09-02
fix: Make DB speed improvements CI/test-only

bramj:
My local dev DB got corrupted because of these settings (a known risk). Small improvement is to just do this config for CI, and not for local development.

Martin edit:
Fix for DB corruption:
```shell
docker compose down && docker volume rm openhexa-app_pgdata
docker compose run app fixtures
```

### DIFF
--- a/docker-compose.ci.yaml
+++ b/docker-compose.ci.yaml
@@ -1,0 +1,9 @@
+# CI-only overrides, layered on top of docker-compose.yaml via COMPOSE_FILE.
+services:
+  db:
+    # Trade crash-durability for speed. Only safe on a throwaway database: with
+    # fsync off, a killed container can leave data pages ahead of the WAL and
+    # corrupt the cluster, so never use this on a local dev database. To get the
+    # same speedup for a local test run:
+    #   COMPOSE_FILE=docker-compose.yaml:docker-compose.ci.yaml docker compose run app test ...
+    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
--- a/docker-compose.ci.yaml
+++ b/docker-compose.ci.yaml
@@ -1,9 +1,0 @@
-# CI-only overrides, layered on top of docker-compose.yaml via COMPOSE_FILE.
-services:
-  db:
-    # Trade crash-durability for speed. Only safe on a throwaway database: with
-    # fsync off, a killed container can leave data pages ahead of the WAL and
-    # corrupt the cluster, so never use this on a local dev database. To get the
-    # same speedup for a local test run:
-    #   COMPOSE_FILE=docker-compose.yaml:docker-compose.ci.yaml docker compose run app test ...
-    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
--- a/docker-compose.test.yaml
+++ b/docker-compose.test.yaml
@@ -1,0 +1,15 @@
+# Throwaway test database, layered on top of docker-compose.yaml. The data
+# directory is a tmpfs, so it lives in RAM and dies with the container — that is
+# what makes the durability settings below safe to use anywhere, including
+# locally. Measured ~30% faster than the durable configuration on the full suite.
+services:
+  db:
+    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
+    volumes:
+      # Replaces the pgdata volume from the base file (compose merges mounts by
+      # target). Peak usage is ~600MB with 8 parallel worker databases; tmpfs
+      # allocates lazily, so this is a ceiling, not a reservation.
+      - type: tmpfs
+        target: /var/lib/postgresql/data
+        tmpfs:
+          size: 2147483648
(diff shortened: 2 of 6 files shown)


docker-compose.ci.yaml
```diff
@@ -0,0 +1,9 @@
+# CI-only overrides, layered on top of docker-compose.yaml via COMPOSE_FILE.
+services:
+  db:
+    # Trade crash-durability for speed. Only safe on a throwaway database: with
+    # fsync off, a killed container can leave data pages ahead of the WAL and
+    # corrupt the cluster, so never use this on a local dev database. To get the
+    # same speedup for a local test run:
+    #   COMPOSE_FILE=docker-compose.yaml:docker-compose.ci.yaml docker compose run app test ...
+    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
```
mrivar: I'm looking into whether we can make it safe for a local DB tests, to have the advantage also locally. Apparently it can help adding a throwaway test DB (tmpfs type volume). wdyt?
If we want to keep it CI-only and not for local, then we can skip this, but it could be handy

docker-compose.yaml
```diff
@@ -33,9 +33,6 @@ services:
   db:
     platform: linux/amd64
     image: postgis/postgis:16-3.5
-    # Trade crash-durability for speed: fine for a dev/CI database, do not
-    # reuse for production deployments.
-    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
```
mrivar, line 38: should we keep `synchronous_commit=off` for local dev deployments?
Apparently mac users can feel a difference in speed (slower with sync commit on), but for linux it would be similar

mrivar (commented):
Cool! Just some minor comments if we want to maximise, given that we are already changing these, but none mandatorily

docker-compose.yaml
bramj: For running tests locally there's an argument to be made, but for running the DB during development it's fine (this was the setting for many years until I changed it a couple of years ago).

docker-compose.ci.yaml
bramj: Feel free to change this PR if you want. I agree it's useful to keep the speed improvement when running the tests locally, but it's a very small gain so if it makes the setup more complex I'd vote against it.

docker-compose.yaml
mrivar: oh yeah, i meant for tests sorry (specially given that sometimes LLMs like to run the tests in the bg)

docker-compose.ci.yaml
mrivar: I'll look into it , I'll see if it's not too overcomplicated

mrivar: some benchmark results to see if it's worth it for local tests:

| Config                                         | Run 1 | Run 2 | Mean   | vs baseline     |
|------------------------------------------------|-------|-------|--------|-----------------|
| durable (baseline)                             | 177s  | 183s  | 180s   | —               |
| synchronous_commit=off only                    | 188s  | 174s  | 181s   | +0.6% (nothing) |
| fsync/sync_commit/full_page_writes=off on disk | 116s  | 130s  | 123s   | -31.7%          |
| same flags + tmpfs                             | 112s  | 113s  | 112.5s | -37.5%          |

Run it a second time with the computer more overload, only with the baseline and the flags + tmpfs and got similar results:

| Config                                         | Run 1 | Run 2 | Mean   | vs baseline     |
|------------------------------------------------|-------|-------|--------|-----------------|
| durable (baseline)                             | 230s  | 189s  | 209.5   | —               |
| same flags + tmpfs                             | 164s  | 136s  | 150s | -28.4%          |

I'd argue a ~30% improvement is worth it if it's not overcomplicated

docker-compose.test.yaml
```diff
@@ -0,0 +1,15 @@
...
+  db:
+    command: postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off
+    volumes:
+      # Replaces the pgdata volume from the base file (compose merges mounts by
+      # target). Peak usage is ~600MB with 8 parallel worker databases; tmpfs
+      # allocates lazily, so this is a ceiling, not a reservation.
+      - type: tmpfs
+        target: /var/lib/postgresql/data
+        tmpfs:
+          size: 2147483648
```
mrivar, line 15: Puts the testing data directory on a tmpfs, so it lives in RAM and dies with the container, removing the corruption risk

mrivar (approved):

bramj:
@mrivar That's really nice! I'm seeing an improvement from ~2 mins to 1min 10sec on my local machine, that's even better than 30% :grin: 

I had a port clash issue that I fixed in the last commit + I did some changes (more opinionated) to the Makefile:
  - Running `make` defaults to a "help" output
  - Add consistency to the b/f (backend/frontend) namings
  - Update `make tb` to run the local tests in the advised way

image

Let me know what you think about those last changes.

PR #2052, opened 2026-09-02
feat: naming agent simpler model (HEXA-1806)

mrivar:
Utility model (simpler and cheaper) for simple stuff such as naming conversations.
It should save around 80% on naming costs.

Question: do we have haiku prepared in the vertex env?


backend/hexa/assistant/model_builder.py
```diff
@@ -129,6 +134,37 @@ def build(self) -> BuiltModel:
...
+        if not utility_model:
+            return self.build()
+        if utility_model not in AiSettings.Model.values:
+            # A typo would otherwise be indistinguishable from a model that is
+            # legitimately unmapped for this provider.
+            logger.warning(
+                "ASSISTANT_UTILITY_MODEL=%r is not a known model; "
+                "utility agents will run on the main model",
+                utility_model,
+            )
```
yolanfery: Shouldn't this check be part of the more generic _build ?

backend/hexa/assistant/model_builder.py
```diff
@@ -129,6 +134,37 @@ def build(self) -> BuiltModel:
...
+        """Cheap model for small utility agents such as conversation naming.
+
+        Returns the organization's main model when the optimization is disabled
+        or when the provider exposes no id for the utility model, so enabling it
+        for a provider later is a one-line addition to that provider's id map.
+
+        `model` overrides ASSISTANT_UTILITY_MODEL for a single build, which lets
+        callers compare models without changing the deployed setting.
+        """
+        utility_model = model or settings.ASSISTANT_UTILITY_MODEL
```
yolanfery: Shouldn't this come from the AISettings instead of the env directly ?

backend/hexa/assistant/model_builder.py
```diff
@@ -129,6 +134,37 @@ def build(self) -> BuiltModel:
...
             model=model,
-            api_name=self._model_api_name,
+            api_name=model_api_name,
             provider_id=provider_id,
         )
+
+    def build(self) -> BuiltModel:
+        return self._build(self._model_api_name)
+
+    def build_utility(self, model: str | None = None) -> BuiltModel:
```
yolanfery: I don't understand why building a model with a different LLM model should be so different. 

I am expecting only the name to change
I am proposing to just keep `build(self, model_api_name: str)`

backend/hexa/assistant/model_builder.py
yolanfery: I don't think this is used somewhere and it feels wrong to use a builder for building different model names and then exposing only one of them

The comment applies to `self._model_api_name = get_api_name(ai_settings.provider, model)` also

yolanfery (changes requested):
Not sure if I am missing anything but I feel like we need to simplify a bit the builder

mrivar:
@yolanfery Thanks for the review! I agree that the builder was doing too much and was deviating from the original design pattern. I've tried to slim it down so it only contains the logic to build new models, while keeping business logic out of it. Let me know what you think.

Also, for the utility model I've added an env kill switch (off by default), in case we encounter an error on managed domains, so that we can revert back without breaking anything.

yolanfery (approved):
Thanks for reworking it

I still think exposing a `build_utility` method in the generic builder is a smell of loose interface design  but we can improve it later, don't want to block for this

backend/config/settings/base.py
```diff
@@ -624,6 +624,12 @@ def _load_oidc_providers() -> list[dict]:
 # europe-west1 keeps data in the EU for GDPR.
 VERTEX_PROJECT_ID = os.environ.get("VERTEX_PROJECT_ID")
 VERTEX_REGION = os.environ.get("VERTEX_REGION", "europe-west1")
+# Kill switch for the cheaper model small "utility agents" (conversation naming)
+# run on. Set to false to put them back on the organization's main model without
+# a deploy — e.g. if the provider turns out not to serve the utility model.
+ASSISTANT_UTILITY_MODEL_ENABLED = (
+    os.environ.get("ASSISTANT_UTILITY_MODEL_ENABLED", "true") == "true"
+)
```
bramj: No need to make this optional, we control the infra

bramj (changes requested):
Sorry to block on this, but I think the approach is not the right one here.

What I would expect is that we can configure a model on each of our agents. So we can say now the `GenerateSqlAgent` will use model X, without impacting any of the other ones. This should be something we can easily change, I think an ENV var with a mapping can work for a v1 for example.

What if the "naming agent" become a real `naming_agent.py`? It would not inherit from `BaseAgent`, but asides from that it makes total sense to consider it a regular agent with an instruction set, `[]` tools, and some output validation (like the SQL agent).

Then we can have a simple mapping that specifies the model on each of the agents, allowing us to also easily test out an open model on just one of the agents. WDYT?

backend/config/settings/base.py
mrivar: Answering here to the env mapping. Right now we have the UI where we can configure which models we choose and what not. I'd say this env mapping shouldn't only be for special cases, like this naming agent, but anything else should go with the org/user specific config, right?
Or let me know if we want to have both env and UI config and how do they overlap... but I feel it's an overcomplication of the logic tbh

.env.dist
```diff
@@ -301,6 +301,13 @@ VERTEX_PROJECT_ID=
 VERTEX_REGION=europe-west1
 GOOGLE_APPLICATION_CREDENTIALS=
 
+# Pins individual assistant agents to a model, overriding the agent's default.
+# JSON object of agent key -> model (haiku, sonnet, opus). Only "naming" and
+# "generate_sql" can be pinned; every other agent runs on the model the
+# organization configured. Leave empty for the defaults (naming on haiku).
+# Example: ASSISTANT_AGENT_MODELS={"naming":"haiku","generate_sql":"sonnet"}
+ASSISTANT_AGENT_MODELS=
```
bramj, line 309: I like this, this will help greatly with experimentation later on 👍

backend/hexa/assistant/agents/generate_sql_agent.py
```diff
@@ -60,6 +60,10 @@ class GenerateSqlAgent(BaseAgent):
     """
 
     instruction_set = InstructionSet.GENERATE_SQL
+    # Opted into ASSISTANT_AGENT_MODELS: writing SQL is a self-contained task
+    # whose model we may want to tune independently. Left unpinned by default,
+    # so it runs on the organization's model unless the setting says otherwise.
+    agent_key = "generate_sql"
```
bramj, line 66: Why not have this on all agents? Also, the comment is superfluous.

backend/hexa/assistant/model_builder.py
```diff
@@ -11,24 +15,20 @@
...
 _DIRECT_ANTHROPIC_MODEL_IDS: dict[str, str] = {
     AiSettings.Model.HAIKU.value: "claude-haiku-4-5-20251001",
     AiSettings.Model.OPUS.value: "claude-opus-4-6",
     AiSettings.Model.SONNET.value: "claude-sonnet-4-6",
 }
 
 _VERTEX_ANTHROPIC_MODEL_IDS: dict[str, str] = {
+    AiSettings.Model.HAIKU.value: "claude-haiku-4-5",
     AiSettings.Model.OPUS.value: "claude-opus-4-6",
 }
```
bramj, line 33: Outside of this PR, but I don't really know why we need key-value pairs here, there's no translation on these anywhere.
Not for now of course, but I feel this construct (aligning on the BYOK case for no good reason) adds needless complexity.

backend/hexa/assistant/agents/base.py
```diff
@@ -198,21 +159,25 @@ class BaseAgent:
     max_requests: int = 30
     output_retries: int | None = None
     history_strip_tools: set[str] = set()
+    # Agents that leave `agent_key` unset always run on the model the
+    # organization chose; only those that opt in with a key can be pinned to a
+    # model through ASSISTANT_AGENT_MODELS. `default_model` is the model the
+    # agent asks for when the environment says nothing, None meaning the
+    # organization's own.
+    agent_key: str | None = None
```
bramj, line 167: I would drop the comment and simply add the agent_key on all agents. There's no reason why we should limit the ones we can override, WDYT?

backend/hexa/assistant/agents/naming_agent.py
```diff
@@ -0,0 +1,114 @@
...
+    return title
+
+
+class NamingResult(NamedTuple):
+    title: str
+    usage: RunUsage
+    cost: Decimal | None
+
+
+class NamingAgent:
```
bramj, line 63: I like this!

backend/hexa/assistant/model_selection.py
```diff
@@ -0,0 +1,99 @@
...
+    if not raw:
+        return {}
+    try:
+        parsed = json.loads(raw)
+    except ValueError:
+        logger.error("ASSISTANT_AGENT_MODELS is not valid JSON; ignoring it")
+        return {}
+    if not isinstance(parsed, dict):
+        logger.error("ASSISTANT_AGENT_MODELS must be a JSON object; ignoring it")
+        return {}
```
bramj, line 31: There are still ways this can break (e.g. by incorrectly setting the ENV var such as `ASSISTANT_AGENT_MODELS='{"naming": ["haiku"]}'`). 

Instead of various checks maybe a more general `try`/`catch`?

backend/hexa/assistant/apps.py
```diff
@@ -11,6 +11,14 @@ class AssistantConfig(CoreAppConfig):
...
     def ready(self):
         super().ready()
+        # Imported here rather than at module level: pulling the agents in at
+        # import time would load the instruction docs before the app registry
+        # is ready.
+        from hexa.assistant.agents import pinnable_agent_keys
+        from hexa.assistant.model_selection import warn_on_unknown_overrides
+
+        warn_on_unknown_overrides(pinnable_agent_keys())
+
```
bramj, line 21: Doing this in `ready` feels like a code smell. I can't put my finger on it exactly, but do we really need this?

bramj (changes requested):
Tested and working well, this is much better! 👍 

I first expected that for BYOK, the chosen model would override anything, but it chooses Haiku as well. That's nice actually and it makes me think we could at some point simplify and ask only an API key from the user and let us choose the model.

I also realize that we'll need more changes to this setup when in the future we want to:
- Add support for more providers in BYOK (OpenAI, OpenRouter, ...)
- Be able to swap to a non-Anthropic model in "managed" mode (here we'll need changes to Pydantic as well)

So just some minor comments and good to go!

backend/hexa/assistant/model_selection.py
mrivar: agreed, I'm making the code simple but I'm going to leave it so that having a typo/error in one of the agents doesn't drop hte ones without typos/errors

backend/hexa/assistant/model_builder.py
mrivar: good point, I'll investigate better ways to do this in the next iteration

bramj (approved):
Nice work @mrivar , let's go! 🚀

PR #2053, opened 2026-09-02
feat: saved queries versioning (HEXA-1768)

mrivar:
Add git forgejo versioning for saved quereis, mimicking the current webapp versioning.

About the deploying plan, as per written in the README:
1. **Deploy.** Every pre-existing query has `repository` null; new ones get a repository at creation.
2. **`manage.py backfill_saved_query_repositories`** creates the missing ones, committing the SQL each query already has as its first version. `--dry-run` first reports what is missing, and what has drifted from the version on record. It is idempotent.
3. **Whatever step 2 misses heals itself.** The first edit that changes the SQL calls `initialize_repository` *before* applying the change, so the history still starts at the query's pre-edit SQL rather than at that edit.

Basically, the "clean" way is to do step 2. In on-prem environments (or localhosted instances) where we cannot be 100% sure if the command  `backfill_saved_query_repositories` was run, it is safe anyway as the repos are created on the first query save, as long as their forgejo instance is up and running.


backend/hexa/data_studio/graphql/schema.graphql
```diff
@@ -234,6 +240,12 @@ input UpdateSavedQueryInput {
   description: String
   "Only the author may change this; PERMISSION_DENIED otherwise."
   visibility: SavedQueryVisibility
+  """
+  The version this edit was made against. When given, the edit is refused with
+  VERSION_CONFLICT if the SQL has changed since. Omit it to overwrite whatever is
+  there.
+  """
+  expectedVersion: String
```
yolanfery: This pattern seems weird (just new compared to what exists). Can we just add a commit on top of what is existing ? 

I think the pattern across the platform is this to ignore those kind of conflicts and just write on top

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
```diff
@@ -0,0 +1,95 @@
...
+
+from hexa.data_studio.models import SavedQuery
+from hexa.git.exceptions import GitError
+
+# For queries whose author's account was deleted. `create_repo` reads only
+# `display_name` and `email`, and naming the instance beats crediting whoever ran this.
+SYSTEM_AUTHOR = SimpleNamespace(display_name="OpenHEXA", email="noreply@openhexa.org")
+
+
+class Command(BaseCommand):
```
yolanfery: Why not a migration instead of a command ?

What happens if local instance forget to run this command ?

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
```diff
@@ -0,0 +1,95 @@
...
+        style = (
+            self.style.SUCCESS
+            if not (missing or drifted or unreadable)
+            else self.style.WARNING
+        )
+        self.stdout.write(
+            style(
+                f"Checked. missing={missing} drifted={drifted} unreadable={unreadable}"
+            )
+        )
```
yolanfery: I feel like those can be merged together and the `--check` param probably renamed to `--dry-run`

backend/hexa/data_studio/models.py
```diff
@@ -73,14 +95,21 @@ def create_if_has_perm(
...
+        with transaction.atomic():
+            saved_query = self.create(
+                workspace=workspace,
+                created_by=principal,
+                name=name,
+                content=content,
+                description=description,
+                visibility=visibility or SavedQueryVisibility.PRIVATE,
+            )
+            saved_query.record_version(principal)
```
yolanfery: What does `record_version` means ? I feel like we can do something more descriptive of what's the intent here

yolanfery: For example : https://github.com/BLSQ/openhexa-app/blob/72681179c930f0a14e49847ec2c90ae8f685efd8/backend/hexa/webapps/models.py#L434-L449

This pattern is way more readable, we see that the repo is created and the reason why we save the newly created instance

backend/hexa/data_studio/models.py
```diff
@@ -180,6 +214,9 @@ class SavedQuery(Base):
         choices=SavedQueryVisibility.choices,
         default=SavedQueryVisibility.PRIVATE,
     )
+    # Which commit `content` matches, so drift can be found. Not a published pointer.
+    # Null until the repository exists (see hexa/git README).
+    last_commit = models.CharField(max_length=64, blank=True, null=True)
```
yolanfery: I don't think we need this. I feel like this can end up in a nightmare to keep it synced. Can we just rely on the last one on the git server ?

backend/hexa/data_studio/models.py
```diff
@@ -236,38 +273,151 @@ def save(self, *args, **kwargs):
...
-                "data_studio.update_saved_query_visibility", self
-            ):
-                raise PermissionDenied
-            self.visibility = visibility
+    def default_repository_name(self) -> str:
+        """Keyed on the pk, not the slug: a deleted query releases its slug, and reusing
+        one lands on the archived repository it left behind. Migration 0010 derives the
+        same name.
+        """
+        return f"{self.workspace.slug}-query-{self.id}"
```
yolanfery: The repo name will be ugly. We need to use slugs. We can have soft delete on the queries, so that slugs are kept clean. Let's try to be consistent with webapps, we can improve what's exisitng but I think we should keep the approaches the same

frontend/src/workspaces/features/SavedQueries/savedQueryErrors.ts
```diff
@@ -32,6 +41,14 @@ export const updateSavedQueryErrorMessage = (
...
   }
+  if (errors?.includes(UpdateSavedQueryError.VersioningUnavailable)) {
+    return VERSIONING_UNAVAILABLE_MESSAGE(t);
+  }
+  if (errors?.includes(UpdateSavedQueryError.VersionConflict)) {
+    return t(
+      "Someone else saved this query while you were editing. Reload it and reapply your changes.",
+    );
+  }
   console.error("Unhandled updateSavedQuery error", errors);
```
yolanfery, line 47: Again I think we can just follow a "commit-on-top" pattern

backend/hexa/data_studio/schema.py
```diff
@@ -269,6 +273,11 @@ def resolve_create_saved_query(_, info, **kwargs):
         return {"success": False, "errors": ["WORKSPACE_NOT_FOUND"]}
     except PermissionDenied:
         return {"success": False, "errors": ["PERMISSION_DENIED"]}
+    # The transaction rolled back, so nothing was created: reported as a failure
+    # rather than as a saved query with no history.
+    except GitError:
+        logger.exception("Could not record the first version of a saved query")
```
yolanfery: This seems like a very temporary logging

backend/hexa/git/forgejo.py
```diff
@@ -426,16 +446,7 @@ def get_commits(
...
-            {
-                "id": c["sha"],
-                "message": c["commit"]["message"],
-                "author_name": c["commit"]["author"]["name"],
-                "author_email": c["commit"]["author"]["email"],
-                "date": c["commit"]["author"]["date"],
-            }
-            for c in response.json()
-        ]
+        return [_commit_summary(commit) for commit in response.json()]
```
yolanfery, line 449: 👍 seems like a good idea to have consistency indeed

backend/hexa/git/mixins.py
```diff
@@ -61,3 +95,17 @@ def create_repo(self, *, files: list[dict] | None = None, user: User) -> str:
...
+
+    class Meta:
+        abstract = True
+
+    @property
+    def git_org(self) -> GitOrg:
+        return GitOrg(
+            slug=self.workspace.organization.slug,
+            display_name=self.workspace.organization.name,
+        )
```
yolanfery: Not sure about this abstraction. Line 109 and 110 assume out-of-nowhere that workspace and organization exist on self

backend/hexa/git/mixins.py
```diff
@@ -27,6 +34,33 @@ def git_org(self) -> GitOrg:
...
+            "page": page,
+        }
+
+    def get_commit_diff(self, sha: str) -> dict:
+        return {
+            **self.client.get_commit(self.git_org.slug, self.repository, sha),
+            "raw_diff": self.client.get_commit_diff(
+                self.git_org.slug, self.repository, sha
+            ),
+        }
```
yolanfery, line 64: 👍 Nice initiative to extract common logic from the webapps

yolanfery (changes requested):
Good start 🚀

Few comments but more generally I think we need to rework the model and migration part. 

- Let's aim to keep things consistent between webapps and saved queries. This will help us for pipelines later
- Can you please clarify what's the migration plan ? It's not fully clear to me yet at this point and I think it would be valuable to put it in writing. Until the management command is run (and fully successful), saved queries are not creating versions, just keeping the last version in DB ? So we will always live with this branching in the codebase since we can't enforce running this command on localhosted instance ? Maybe we should just try to fetch from the repo and catch exception and fallback on the database (acting as a cache) stored value ? This could be a more generic pattern available and applied (optionally) to pipelines (maybe webapps) helping with instances that did not run the command but also instances having a git server downtime (or wanting a more performant fetch)
 
 [Sorry for the unstructured text 🤪 but I think it's important to clarify now and define a good direction before moving forward]

backend/hexa/data_studio/graphql/schema.graphql
mrivar: Yeah you're right, I went a lil bit overboard with this. Was just thinking on how to properly act when conflicting changes between 2 different users, but it is not that big of a deal. Removing it to keep it cleaner

backend/hexa/data_studio/models.py
mrivar: Yeah it doesn't hold up much value without the conflict lock and I agree it could have broguht more problems than solutions.
The second "task" it held was making sure if a repo was already created or not for the query, but that can be checked in easier ways.

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
mrivar: good catch!

frontend/src/workspaces/features/SavedQueries/savedQueryErrors.ts
mrivar: yep, removing the conflict logic altogether

backend/hexa/git/mixins.py
mrivar: fair enough, it doesn't save that much atm anyway

backend/hexa/data_studio/models.py
mrivar: fair enough, going for that approach and a better rename for commiting/saving versions into git, thanks!

backend/hexa/data_studio/schema.py
mrivar: Sure, but I'd say it'd be handy to know if something went wrong, and it shouldn't add much noise anyway. I'd add more information to it tho.

backend/hexa/data_studio/models.py
mrivar: Done, using the slug now. I'm not sure about the soft-deleting for all queries tho, when deleting a user it deletes any private queries on cascade. I added a way to make sure they're safe anyways tho, see (the unique tail does it):

```python
return build_repo_name(
      f"{self.workspace.slug}-query-{self.slug}", unique=self.id.hex[:8]
  )
```

Btw, I saw an issue with the forgejo repo and using this slug, which is that it only allows for repos of maximum 100 chars long. I added a cap for the saved queries, but the static webapps are at risk of failing without properly creating the repo -> needs a long workspace and webapp name. Rare but could happen

yolanfery: Good catch, I will improve it https://bluesquare.atlassian.net/browse/HEXA-1808

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
```diff
@@ -0,0 +1,87 @@
...
+from types import SimpleNamespace
+
+from django.core.management.base import BaseCommand
+
+from hexa.data_studio.models import SavedQuery
+from hexa.git.exceptions import GitError
+
+# For queries whose author's account was deleted. `create_repo` reads only
+# `display_name` and `email`, and naming the instance beats crediting whoever ran this.
+SYSTEM_AUTHOR = SimpleNamespace(display_name="OpenHEXA", email="noreply@openhexa.org")
```
yolanfery, line 11: This does not sound ideal for localhosting right ?

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
```diff
@@ -0,0 +1,87 @@
...
+        tally = Counter()
+
+        queryset = SavedQuery.objects.select_related(
+            "workspace__organization", "created_by"
+        )
+        for saved_query in queryset.iterator():
+            # Only queries with no repository are touched, so a re-run is free.
+            if not saved_query.has_history:
+                self._create(saved_query, tally, dry_run)
+            elif dry_run:
```
yolanfery, line 40: We are not checking for dry run before the creation ?

yolanfery: I see now that `_create` checks for it in its logic, I think we can make this cleaner by making things more readable and avoiding to check for dry run everywhere

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
```diff
@@ -0,0 +1,87 @@
...
+                # A round trip per query, only worth paying for when nothing is written
+                # anyway: the audit of what the run above would have skipped.
+                self._report_drift(saved_query, tally)
+            else:
+                tally["already_created"] += 1
```
yolanfery, line 45: Not fan of this pattern of using a magic string to count stuff. We can probably use a different pattern or at the very minimum a typed structure

backend/hexa/data_studio/models.py
```diff
@@ -73,14 +84,22 @@ def create_if_has_perm(
...
+                workspace=workspace,
+                created_by=principal,
+                name=name,
+                content=content,
+                description=description,
+                visibility=visibility or SavedQueryVisibility.PRIVATE,
+            )
+            saved_query.initialize_repository(principal)
+            # Saved on its own: `initialize_repository` is what names the repository.
+            saved_query.save(update_fields=["repository"])
```
yolanfery, line 100: 👍

backend/hexa/git/mixins.py
```diff
@@ -14,6 +14,15 @@
...
 class GitRepoMixin(models.Model):
+    """A model whose history lives in a git repository of its own.
+
+    A subclass answers `git_org` and `has_history` and inherits the reading side.
+    Writing is deliberately not declared: committing one file and committing a tree
+    have too little in common for one signature. See `hexa/git` README.
+    """
+
+    # Not null, for a model created with its repository. One that can exist before its
+    # repository does overrides this nullable, and `has_history` with it.
```
yolanfery, line 25: Not sure we need those comments

backend/hexa/git/naming.py
```diff
@@ -0,0 +1,13 @@
...
+
+def build_repo_name(readable: str, *, unique: str) -> str:
+    """Compose a repository name that fits Forgejo's limit and stays unique whatever it costs.
+
+    `readable` is for whoever browses the repository list and is cut to make room;
+    `unique` is kept whole, so two names cannot collide by having been cut at the same
+    point — which also makes the name safe to derive from an identifier a deleted row
+    releases, since the repository it left behind is named after a different one.
+    """
+    return f"{readable[: REPO_NAME_MAX_LENGTH - len(unique) - 1]}-{unique}"
```
yolanfery, line 13: Again, I doubt it's the right approach because it makes the repo names (used in git) a bit less humand readable/guessable 

I think we can soft delete queries (we can make the user field nullable if thats the part blocking), to ensure slugs are unique

backend/hexa/data_studio/README.md
```diff
@@ -26,6 +26,49 @@ calls it separately so it can refuse a request before reserving a concurrency sl
...
+   creation.
+2. **`manage.py backfill_saved_query_repositories`** creates the missing ones, committing
+   the SQL each query already has as its first version. `--dry-run` first reports what is
+   missing, and what has drifted from the version on record. It is idempotent.
+3. **Whatever step 2 misses heals itself.** The first edit that changes the SQL calls
+   `initialize_repository` *before* applying the change, so the history still starts at the
+   query's pre-edit SQL rather than at that edit.
+
+Running the command is therefore an accelerator, not a prerequisite — which is what makes
+this safe on self-hosted instances nobody can make run it.
```
yolanfery, line 59: Fair enough 👍

backend/hexa/data_studio/models.py
yolanfery: > I'm not sure about the soft-deleting for all queries tho, when deleting a user it deletes any private queries on cascade.

Maybe we can nullify the relation so that we can soft delete ? I am worried to start creating repos with those kind of names (they are less human-friendly, they can't be guessed from the saved query slug (think about an AI or human that wants to infer the git url quickly)) and that's a decision we gonna need to live with (migration is difficult once the repo are created...)

yolanfery (changes requested):

backend/hexa/git/mixins.py
mrivar: Yeah sorry about that, I need to go through the comments. I closed this PR before the rewriting comments policy.

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
mrivar: oh right, any idea what should we have instead?

yolanfery: Nullify it 😛

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
mrivar: Agree, improving the whole backfill command

backend/hexa/data_studio/management/commands/backfill_saved_query_repositories.py
mrivar: Yeah, let's simplify it

backend/hexa/git/naming.py
mrivar: Yeah I guess we'll have to go for that. What do you think we should do about the 100 chars limit? is that something we can change via config or should we still keep it into account when creating new repos?

yolanfery: Yeah it's a difficult question

Maybe we should cap the new workspace/webapp slugs ? So that they stay short and readable ? 
And keep your approach of truncation as a fallback for existing long webapps, workspace and pipelines [since we want to migrate to git for those as well]

WDYT ? @mrivar @bramj

yolanfery: (So that we limit occurrences of such unguessable repo names)

bramj: @yolanfery Just seeing this message now, sounds like a good idea.

And I think Yolan is doing a fix for the 100 char limit @mrivar , right?

mrivar: That's a good option.
I was thinking we could also make MCP tools aware of how we construct the repo names (new function `build_repo_name`) so it can easily get the name of the repos too. We also have `self.repository` where they can get the repo names.
Btw, I added the fallback to webapps as well in this PR (it was a one line addition)

yolanfery: > I was thinking we could also make MCP tools aware of how we construct the repo names (new function build_repo_name) so it can easily get the name of the repos too. We also have self.repository where they can get the repo names.

Good idea, I will just add the repo url to the MCP get webapp query in my next PR

> Btw, I added the fallback to webapps as well in this PR (it was a one line addition)

Thanks !!

yolanfery (approved):
Well done !

PR #2055, opened 2026-09-02
chore: PR digest

yolanfery:
An attempt at facilitating reviews by digesting the content of the PR for the reviewer

It's a starting point and we gonna need to tweak as we go, I feel like the result is still a bit verbose

### DIFF
--- a/.claude/skills/pr-digest/SKILL.md
+++ b/.claude/skills/pr-digest/SKILL.md
@@ -1,0 +1,77 @@
+---
+name: pr-digest
+description: Digest a pull request so a reviewer can go straight to what matters — ranked pointers into the diff, the dependency/model/migration/tooling changes, the judgment calls behind them, and a map of which files are critical. Use when asked to summarise, digest, or write a review guide for a PR.
+---
+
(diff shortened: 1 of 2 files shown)


yolanfery:
@bramj can you add BLSQBOT_TOKEN and LOGFIRE_WRITE_TOKEN to this repo env vars ? It seems I don't have access/permissions to get those (might be a good idea to add BLSQBOT credentials to 1password)

bramj:
@yolanfery done! I've added `LOGFIRE_WRITE_TOKEN` to OpenHexa infra vault. BLSQ bot is already in 1password but in a general infra vault.

bramj (approved):
Very cool, curious to see how this plays out in the wild.

I agree that it's a bit verbose, but using it will make it clear where we can improve it, so let's do just that 🙂

yolanfery:
> @yolanfery done! I've added `LOGFIRE_WRITE_TOKEN` to OpenHexa infra vault. BLSQ bot is already in 1password but in a general infra vault.

Thanks a lot ! (and my bad 😅)

PR #2062, opened 2026-09-11
feat visualise pipelines with dag

DimitriKwihangana:

## Changes

- Change 1
- Change 2

## How/what to test

## Screenshots / screencast
image


backend/hexa/pipelines/dag.py
```diff
@@ -0,0 +1,172 @@
...
+    Task ids are function names; edge sources are either a task id or a parameter code, and
+    the caller discriminates by looking the source up in the task list.
+
+    Unlike ``_parse_parameters_from_zipfile``, which raises so a bad upload is rejected, this
+    is deliberately silent: it feeds a read-only view and must never break the pipeline page.
+    """
+    try:
+        tree = ast.parse(source)
+    except SyntaxError:
+        return _empty()
```
mrivar: Very minor, but we could also catch other errors such as `RecursionError`, `ValueError`, or `MemoryError` here; or wrap it in a generic exception. They're more rare but it would be nice to catch as it breaks the pipelines page.

backend/hexa/pipelines/dag.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Only ``pipeline.py`` at the archive root is read, mirroring the SDK's ``get_pipeline()``.
+    Scanning the archive for a ``@pipeline`` decorator would risk describing a module that
+    never runs.
+    """
+    try:
+        with ZipFile(io.BytesIO(zipfile_data)) as zip_file:
+            source = zip_file.read(ENTRYPOINT).decode()
+    except Exception:
+        return _empty()
```
mrivar: I wonder if we want a logger warning here and below in the SyntaxError. To be analysed, to be honest it might be only noise

backend/hexa/pipelines/schema/types.py
```diff
@@ -385,6 +386,20 @@ def resolve_pipeline_version_zipfile(version: PipelineVersion, info, **kwargs):
...
+    """Derive the task graph from the version's code.
+
+    Computed on the fly rather than stored: it needs no migration and works on versions
+    uploaded long before this field existed. The archive bytes are already loaded with the
+    version row, so this costs CPU but no extra query.
+    """
+    if not version.zipfile:
+        return {"tasks": [], "edges": []}
+
+    return extract_dag_from_zipfile(version.zipfile)
```
mrivar: We re-parse the dag on every page load. Would be worth to consider if we want to save it as a JSON in the DB instead to avoid re-parsing every time (it can be done lazily, so the first time the pipeline page is loaded for old pipelines, or on upload versions for new ones).
But it's worth a compute analysis to make a decision on how worthwhile it is.

backend/hexa/pipelines/dag.py
```diff
@@ -0,0 +1,172 @@
...
+        if isinstance(statement, ast.Assign):
+            call = statement.value
+            if len(statement.targets) == 1 and isinstance(
+                statement.targets[0], ast.Name
+            ):
+                target = statement.targets[0].id
+        elif isinstance(statement, ast.Expr):
+            call = statement.value
+        else:
+            continue
```
mrivar: Claude found some edge-cases on ast.walk that might be worth a look:
- `ast.walk` visits code breadth-first, not in source order. All top-level statements are handled before anything nested in an `if`, `for`, `with` or `try`, so the map of which name holds which task's result gets out of date.
- `a = t1()` / `for …: t2(a)` / `a = t3()` gives `t3 → t2`. At runtime the edge is `t1 → t2`.
- `with …: a = t1()` then `t2(a)` gives no edge at all.
- When a name is reassigned to something other than a task, the old entry is kept (dag.py:165). `a = t1(); a = foo(); t2(a)` gives `t1 → t2`.
- The walk also enters helper functions defined inside the pipeline. A helper argument named like a pipeline parameter creates a false edge: `def h(x): t2(x)` gives `x → t2`.

mrivar (changes requested):
Looks really cool!! I have some comments about some efficiency improvements and some edge-cases, but nothing super big. This looks quite impressive for a first version!
I notices that when outputs of functions don't go into inputs of the enxt one, they're not "connected", which I guess is intentional in keeping the DAG analysis conservative. But just wanted to point it out as some nodes might look disconnected/out of order in certain pipelines.
Also, if disconnected, nodes appear in order of appearance in the code rather than order of execution. Again, just pointing it out. Still very good for a first iteration and we can ignore it for now.

PR #2065, opened 2026-09-16
feat: agents with more flexible providers (HEXA-1809)

mrivar:
For our assistant agents: Have a more flexible way to change in between providers in case we have errors/problems

## Changes

- New env variables to manage the providers/models more flexibly
  - `ASSISTANT_MANAGED_PROVIDERS`: providers in our managed vertex backend
  - `ASSISTANT_MODELS`: map of supported simple names to models available
  - `ASSISTANT_AGENT_MODELS`: mapping of agents to models (can be models defined in ASSISTANT_MODELS or api names directly such as `google-cloud:gemini-3-pro`)
- Changes the way models for agents are built, so it depends on those env variables instead of hardcoded model builders only


backend/config/settings/base.py
```diff
@@ -624,8 +624,18 @@ def _load_oidc_providers() -> list[dict]:
...
+# The providers enabled in Vertex (see ai_models/vertex.py PROVIDERS), comma separated.
+ASSISTANT_MANAGED_PROVIDERS = os.environ.get("ASSISTANT_MANAGED_PROVIDERS", "")
+# Maps the logical models managed organizations run on (see AiSettings.Model) to a
+# pydantic-ai model id "<provider>:<model>", as JSON.
+# Example: {"opus": "anthropic:claude-opus-5"}.
+# Overrides the code defaults; anything left out keeps its default.
+ASSISTANT_MANAGED_MODELS = os.environ.get("ASSISTANT_MANAGED_MODELS", "")
+# Pins assistant agents to models, as a JSON map of agents (see AgentKey enum)
+# to either a logical model or a model id.
+# Example: {"naming": "haiku", "generate_sql": "google-cloud:gemini-3-pro-preview"}
```
yolanfery: [Minor] Lot to read here, we might want to simplify

.env.dist
```diff
@@ -299,16 +299,38 @@ ASSISTANT_MANAGED=false
...
+# Example: ASSISTANT_MANAGED_MODELS={"opus":"anthropic:claude-opus-5"}
+ASSISTANT_MANAGED_MODELS=
+
+# Pins assistant agents to a model, overriding the agent's default.
+# JSON object of agent key -> model.
+# Agent keys are defined in each agent in the code; the model is either a logical
+# model from ASSISTANT_MANAGED_MODELS or a "<provider>:<model>" id of its own.
+# An entry naming an unknown agent or model, or one the organization holds no credentials for, is
+# logged and skipped, leaving the other entries applied; invalid JSON drops them all.
+# Example: ASSISTANT_AGENT_MODELS={"naming":"haiku","generate_sql":"google-cloud:gemini-3-pro-preview"}
```
yolanfery: I think its way too verbose

yolanfery (approved):
Well done, I only had 2 comments but there were already fixed by your last commits

I have the feeling that it's a bit complex (lot of wrappers, config class) for what we want to do but I can't propose an alternative for now

PR #2068, opened 2026-09-18
fix: type error on Parquet arrays datasets

mrivar:
Little side quest, saw this Sentry error and thought it could be a quick fix: https://bluesquareorg.sentry.io/issues/7529757472/?alert_rule_id=7767430&alert_type=issue&notification_uuid=cd451ca3-590f-4556-a10b-2894d7efba1a&project=5893042&referrer=slack

## Changes

- On Parquet datasets, when encoding, sometimes it uses `np.ndarray`  instead of python lists. Added a way to encode them as well and avoid this errors


bramj (approved):
Nice one @mrivar !

